PR #27, opened 2018-10-19
Changed the behaviour of the type-action when the allAtOnce option is

weyert:
Changed the behaviour of the type-action when the allAtOnce option is set to false now it does the following:

 * If `allAtOnce` is set to false the typing will trigger the following events, for each character `keyDown`, and `keyUp`, if `keyDown` did not cancel the event it will also trigger the `change`-event 
 * If `delay` is set higher then zero it will wait n ms before typing the next character when `allAtOnce` is not set


Gpx:
Thanks for your PR @weyert. I made some changes and opened a new one at #29

PR #34, opened 2018-10-22
Fix the issue with Babel 6 causing `regeneratorRuntime is undefined

weyert:
I updated the `user-event`-package in my work project and I was getting `regeneratorRuntime is undefined` errors while running unit tests that import this package.

I think the PR should fix the problem it appears to do for me.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,9 @@
 {
-  "presets": ["env", "react"]
+  "presets": ["env", "react"],
+  "plugins": [
+    ["transform-runtime", {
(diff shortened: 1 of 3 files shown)


Gpx:
I had the same issue. I took the opportunity to upgrade to babel 7. I'm waiting for the GitHub outage to finish before I merge

weyert:
Oh cool!

Gpx:
It should be fixed in v1.4.1. Thanks @weyert!

weyert:
Thank you!

PR #70, opened 2019-03-17
Disabled checkboxes should ignore clicks.

robcaldecott:
I noticed that a `disabled` checkbox would still allow `click` to work which does not feel right (it is not how `button` or other elements behave.) Here is a small PR to address this.

### DIFF
--- a/__tests__/click.js
+++ b/__tests__/click.js
@@ -191,6 +191,15 @@ describe("userEvent.click", () => {
     expect(getByTestId("input")).toHaveProperty("checked", true);
   });
 
+  it('should not check a disabled <input type="checkbox">', () => {
+    const { getByTestId } = render(
+      <input id="input" data-testid="input" type="checkbox" disabled />
+    );
+    expect(getByTestId("input")).toHaveProperty("checked", false);
+    userEvent.click(getByTestId("input"));
+    expect(getByTestId("input")).toHaveProperty("checked", false);
+  });
+
   it("should submit a form when clicking on a <button>", () => {
     const onSubmit = jest.fn();
     const { getByText } = render(
--- a/src/index.js
+++ b/src/index.js
@@ -96,7 +96,9 @@ const userEvent = {
         break;
       case "INPUT":
         if (element.type === "checkbox") {
-          clickCheckbox(element);
+          if (!element.disabled) {
+            clickCheckbox(element);
+          }
           break;
         }
       default:


Gpx:
Hi! Thanks for your PR.

I'm not sure about this use-case. In theory, you can click a checkbox that is disabled, the only thing is nothing should happen.

What is happening right now if you execute `userEvent.click` on a disabled checkbox?

robcaldecottvelo:
It will toggle the checked state currently which doesn't feel correct. Happy to discuss this further but this caught me out recently writing tests using `fireEvent.click` which lead me to this repo.

Gpx:
Interesting. I will try to investigate a bit more this afternoon after work

Gpx:
Ok, I've investigated it a bit further.

You are right. In the browser, the click event wouldn't fire on a disabled input: https://jsbin.com/hitofu/edit?html,js,output

In fact [`jsdom`](https://github.com/jsdom/jsdom/) doesn't fire click events for such components. You can find the code that blocks it [here](https://github.com/jsdom/jsdom/blob/49bf806f30aae3eff20567d032a74124e12f1284/lib/jsdom/living/nodes/HTMLInputElement-impl.js#L76-L78).

I tried to write these tests:

```jsx
 it("userEvent", () => {
  const onClick = jest.fn();
  const { getByTestId } = render(
    <input data-testid="input" disabled onClick={onClick} />
  );
  const input = getByTestId("input");
  expect(input).toHaveProperty("checked", false);
  userEvent.click(input);
  expect(input).toHaveProperty("checked", false);
  expect(onClick).not.toHaveBeenCalled();
});

it("fireEvent", () => {
  const onClick = jest.fn();
  const { getByTestId } = render(
    <input data-testid="input" disabled onClick={onClick} />
  );
  const input = getByTestId("input");
  expect(input).toHaveProperty("checked", false);
  fireEvent.click(input);
  expect(input).toHaveProperty("checked", false);
  expect(onClick).not.toHaveBeenCalled();
});
```

They both pass.

I think this is the correct behavior. Is there something I'm missing? Maybe you're running an old version of `jsdom`?

I'm going to close the PR for now but feel free to comment and we can open it again if needed.

robcaldecottvelo:
Thanks for looking. We're using the version of `jsdom` that comes with `create-react-app 2` which is `11.12.0` which should have the fix. Odd, but we can manage. BTW don't your tests needs `type="checkbox"`?

Gpx:
Ah! Of course, you're right. If I put the `type="checkbox"` the test fails.
I think `jsdom` is doing it. I'm trying to reproduce it

Gpx:
It looks like [this code](https://github.com/jsdom/jsdom/blob/49bf806f30aae3eff20567d032a74124e12f1284/lib/jsdom/living/nodes/HTMLInputElement-impl.js#L57) is run even if the input is disabled. I'm trying to figure out who's calling it

Gpx:
Ok, this seems to be the offender: https://github.com/jsdom/jsdom/blob/11.12.0/lib/jsdom/living/nodes/HTMLElement-impl.js#L27

Which exists in version 11.12.0.

The latest version is 14.0.0 and it does not have the bug.

It looks like jest is the one requiring that version:

```
npm list jsdom
user-event@0.0.0-development
└─┬ jest@24.5.0
  └─┬ jest-cli@24.5.0
    └─┬ jest-config@24.5.0
      └─┬ jest-environment-jsdom@24.5.0
        └── jsdom@11.12.0
```

https://github.com/facebook/jest/blob/master/packages/jest-environment-jsdom/package.json#L18

---

I would wait until jest updates their dependency

robcaldecottvelo:
Thanks again! Jest is getting a major update with CRA3 which should land soon.

robcaldecottvelo:
Although I fear Jest will be languishing on an older jsdom because of Node minimum version requirements.

PR #118, opened 2019-05-09
Add pretty-quick instead of lint-staged + format code

MichaelDeBoey:
I wanted to set the following Prettier config too, but that would have lead to too much code changes

```js
semi: false,
singleQuote:  true,
tabWidth: 4,
```


Gpx:
Thanks for the PR but I prefer to keep the default settings for prettier to avoid bikeshedding over it. As for `lint-staged`, it's useful because it would allow adding additional steps in the future (lint for example.

PR #126, opened 2019-06-06
Added selectOptions

michaellasky:
Simulates selecting options from a <select> or <select multiple> tag.

### DIFF
--- a/__tests__/selectoptions.js
+++ b/__tests__/selectoptions.js
@@ -162,15 +162,11 @@ describe("userEvent.selectOptions", () => {
             3
           </option>
         </select>
-        <input type="submit" data-testid="submit" value="Submit" />
       </form>
     );
 
     userEvent.selectOptions(getByTestId("element"), ["1", "3"]);
 
-    fireEvent.click(getByTestId("submit"));
-
-    expect(onSubmit.mock.calls.length).toBe(1);
     expect(getByTestId("val1").selected).toBe(true);
     expect(getByTestId("val2").selected).toBe(false);
     expect(getByTestId("val3").selected).toBe(true);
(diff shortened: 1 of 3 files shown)


michaellasky:
Some thoughts:

- Should the selector function passed to selectOptions receive the option element as a parameter (as it currently does)?  Or should we strictly pass only the value (or child text node?) of the option, basically simulating the criteria a user would select an option with?

- Currently it just simulates the relevant events on the dom nodes and then sets the selected attribute. It assumes the select element is in a virgin state with nothing selected.  Is this a valid assumption? Should it explicitly set an option element's "selected" attribute to false if it isn't selected?

__tests__/selectoptions.js
```diff
@@ -0,0 +1,190 @@
...
+            3
+          </option>
+        </select>
+        <input type="submit" data-testid="submit" value="Submit" />
+      </form>
+    );
+
+    userEvent.selectOptions(
+      getByTestId("element"),
+      option => option.value === "1" || option.value === "3"
```
Gpx: I like the function approach, but I think it would be handy to be able to pass an array of values or a single value as well

`userEvent.selectOptions(element, ['1', '3'])`
`userEvent.selectOptions(element, '3')`

michaellasky: Yeah, I think that's a good idea.  The function seemed best while I was trying different ideas, but I think a simple array or scalar value is more readable, and I don't foresee ever needing more than that.  And it also conforms more to the idea of writing tests from the perspective of a user.  

I can change that ASAP (most likely tonight).  Any thoughts as to whether I keep the ability to pass the function in there?  Can you think of any situation where you'd want to pass a function rather than just an array of values?

michaellasky: Also will update the README

Gpx: I can't think of any but it shouldn't be a big deal to keep it

Gpx: I see now you've removed it. Let's keep it like that for now then

Gpx:
Merged in #130

Issue #119, opened 2019-05-13
.type doesn't work with vue-testing-library


afontcu:
Hi! First of all, thank you for your work on this repo!

I wanted to highlight that the current implementation of `.type()` isn't working when used with `vue-testing-library`.

Not quite sure why yet thou, I believe it has something to do with using `fireEvent.change` instead of `fireEvent.input` [here](https://github.com/testing-library/user-event/blob/master/src/index.js#L166). I have passing tests using `fireEvent.input`, and the minute I change them to use `.change`, they start failing. Is there any reason to prefer change event over input?

Thanks!

Gpx:
Hi, `user-event` is only tested with React at the moment for the simple reason that it's what I use at work. I would be keen to add tests that cover other frameworks.

I think the difference between `input` and `change` is that in React you have two separate handlers for them: `onInput` and `onChange`. I also never seen anyone using `onInput`.

I don't really know how it works in Vue-land. Is it possible that you have this distinction too?

afontcu:
Hi! Thanks for answering.

Vue also maps native events such as Input and Change. However, `input` is the most widely used in Vue world. It is even used in the built-in [v-model](https://vuejs.org/v2/guide/forms.html) feature, which is just syntax sugar to create a two-way binding of an input value.

Main difference (AFAIK) is that change event triggers when the element has finished changing (when it loses focus), while input event triggers every time a value is modified. You can see it in [this Codesandbox](https://codesandbox.io/s/o1j015y5p5).

Gpx:
Since we're trying to simulate a browser I think we should fire both events instead of replacing it.
Right now `type` is firing `keydown -> change -> keyup`. We should change it to be `keydown -> input -> change -> keyup`

afontcu:
AFAIK (and correct me if I'm wrong), `change` should only be emitted _after_ moving away from the input.


> Note: The input event is fired every time the value of the element changes. This is unlike the change event, which only fires when the value is committed, such as by pressing the enter key, selecting a value from a list of options, and the like. ([source](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event))


If that's the case, would it make sense send an `input` event for every char, and then send a `change` event with the whole string?

`keydown`
`input "h"`
`keyup`
`keydown`
`input "i"`
`keyup`
`change "hi"`

Gpx:
> AFAIK (and correct me if I'm wrong), change should only be emitted after moving away from the input.

[...]

PR #129, opened 2019-06-06
Add Vue tests

afontcu:
Hi! 👋 As discussed in #119, I'm submitting a draft PR with a proof of concept for Vue tests. I just added tests for `userEvent.click`, as firstly I'd like to discuss the result and see if it's worth it :)

Tests do the exact same thing as React's. The main difference is the callback prop pattern (which I changed in favor of emitted events, way more idiomatic).

There are several things I'd like to highlight (and debate):

~1. List of emitted events doesn't include `mouse*` events, because AFAIK `wasAnotherElementFocused` is always false (Added code comments). Should I find a way to mount components differently, so I can check that mouse events are being sent?~

2. Also, `change` events are triggered twice. (e.g. [here](https://github.com/testing-library/user-event/pull/129/files#diff-d6a3965b2177d19c49ab5daaf670314cR78)).

3. In VTL, all `fireEvent.*` are converted into async functions. Here, they are (obviously) not. Not sure about the consequences, but I'm a bit worried. Yet, tests work 🤷‍♂.

I wasn't able to configure JSX for Vue (as it clashed with React's). So I first wrote tests using render functions, but I believe tagged templates are easily read. I wish I could use JSX, but it's OK I guess!

All in all, I feel it looks quite good! 😄 

So, (2) needs fixing, but I feel those are the kind of differences we were discussing in the issue linked above. (3) worries me the most. I guess @dfcook has the answer 😇 .

Any thoughts?

---

oh, and btw: there's an ongoing error with js-dom and `Error: Not implemented: HTMLFormElement.prototype.submit` ([related](https://github.com/jsdom/jsdom/issues/1937)). It's just a console warning, and tests are still passing. I think my changes had nothing to do with it 🤔

### DIFF
--- a/__tests__/vue/click.js
+++ b/__tests__/vue/click.js
@@ -1,0 +1,342 @@
+import { render, cleanup } from "@testing-library/vue";
+import "jest-dom/extend-expect";
+import userEvent from "../../src";
+
+afterEach(cleanup);
+
+describe("userEvent.click", () => {
+  it.each(["input", "textarea"])(
+    "should fire the correct events for <%s>",
+    type => {
+      const events = [];
+      const eventsHandler = jest.fn(evt => events.push(evt.type));
+      const { getByTestId } = render({
+        render: function(h) {
+          return h(type, {
+            attrs: {
+              "data-testid": "element"
+            },
+            on: {
+              mouseOver: eventsHandler,
+              mouseMove: eventsHandler,
+              mouseDown: eventsHandler,
+              focus: eventsHandler,
+              mouseUp: eventsHandler,
+              click: eventsHandler
+            }
+          });
+        }
+      });
+
+      userEvent.click(getByTestId("element"));
+
+      // baseElement is always <body>, si wasAnotherElementFocused is false
+      expect(events).toEqual([
+        // "mouseover",
+        // "mousemove",
+        // "mousedown",
+        "focus",
+        // "mouseup",
+        "click"
+      ]);
+    }
+  );
+
+  it('should fire the correct events for <input type="checkbox">', () => {
+    const events = [];
+    const eventsHandler = jest.fn(evt => events.push(evt.type));
+
+    const { getByTestId } = render({
+      render: function(h) {
+        return h("input", {
+          attrs: {
+            type: "checkbox",
+            "data-testid": "element"
+          },
+          on: {
+            mouseOver: eventsHandler,
+            mouseMove: eventsHandler,
+            mouseDown: eventsHandler,
+            focus: eventsHandler,
+            mouseUp: eventsHandler,
+            click: eventsHandler,
+            change: eventsHandler
+          }
+        });
+      }
+    });
+
+    userEvent.click(getByTestId("element"));
+
+    expect(events).toEqual([
+      // "mouseover",
+      // "mousemove",
+      // "mousedown",
+      // "mouseup",
+      "click",
+      "change",
+      "change" // Right now it's receiving two `change` events.
+    ]);
+
+    expect(getByTestId("element")).toHaveProperty("checked", true);
+  });
+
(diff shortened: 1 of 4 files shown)


__tests__/vue/click.js
```diff
@@ -0,0 +1,342 @@
...
+              mouseUp: eventsHandler,
+              click: eventsHandler
+            }
+          });
+        }
+      });
+
+      userEvent.click(getByTestId("element"));
+
+      // baseElement is always <body>, si wasAnotherElementFocused is false
```
afontcu: Missing events here:

__tests__/vue/click.js
```diff
@@ -0,0 +1,342 @@
...
+            change: eventsHandler
+          }
+        });
+      }
+    });
+
+    userEvent.click(getByTestId("element"));
+
+    expect(events).toEqual([
+      // "mouseover",
```
afontcu: here:

__tests__/vue/click.js
```diff
@@ -0,0 +1,342 @@
...
```
afontcu: here:

__tests__/vue/click.js
```diff
@@ -0,0 +1,342 @@
...
+            mouseUp: eventsHandler,
+            click: eventsHandler
+          }
+        });
+      }
+    });
+
+    userEvent.click(getByTestId("div"));
+    expect(events).toEqual([
+      // "mouseover",
```
afontcu: ...and here:

calebeby:
@afontcu For (4), it looks like the functions are all being converted to async so that they don't resolve till the next tick of the event loop (using `await wait()`), probably so that vue can run the event handlers and rerender before the next assertion runs

Gpx:
Hey @afontcu, thanks for writing this. It seems Vue and React think very differently about events. What do you say if we export two different `userEvent` objects? One for Vue and one for React.

Something like: `import @testing-library/user-event/react`.

Eventually, if this library ends up being included in the main packages it should make sense

__tests__/vue/click.js
```diff
@@ -0,0 +1,342 @@
...
+      const events = [];
+      const eventsHandler = jest.fn(evt => events.push(evt.type));
+      const { getByTestId } = render({
+        render: function(h) {
+          return h(type, {
+            attrs: {
+              "data-testid": "element"
+            },
+            on: {
+              mouseOver: eventsHandler,
```
calebeby:
```suggestion
              mouseover: eventsHandler,
```

@afontcu If you lowercase the handler properties, then they fire correctly (applies to all the events in this file)

calebeby: @afontcu I created a PR to your branch that fixes this: https://github.com/afontcu/user-event/pull/1

afontcu: Oh wow, I didn't even think about that. Really nice catch! 🙌

afontcu:
Hi everybody!

Thanks to @calebeby PR, now tests are passing while tracking for the right events. This alleviates the differences between Vue and React implementations, I think.

Do you folks think that exporting two different objects is the best approach here? I guess it'd work, but it'd be amazing if we could find the way for a single object to work.

What's missing from my point of view is that double `change` events being emitted. Apart from that, are we on the right track to provide a single object for both Vue and React?



Thanks all for your inputs (no pun intended)! 😃

calebeby:
I don't think that it is necessary for `user-event` to have separate versions for different libraries. If `user-event` fires the same events that real users do, then it should be compatible with _all_ libraries.

I created a couple codesandbox's that we could use to check our work against the real events that browsers fire:

https://codesandbox.io/s/input-typecheckbox-events-9m8ts
https://codesandbox.io/s/input-typetext-events-tkftr

afontcu:
I'm closing this since a new, rebased PR has been merged. I'll open up a new one with additional tests. Thanks all for the feedback!

PR #131, opened 2019-06-07
selectOptions should fire onChange on a SELECT element

bogdanbodnar:
For first, thanks for this library :)

I use `onChange` event handler on my select elements and I realized that selectOptions doesn't fire it. I've added a simple test to prove it.

I'm would like to make a proposal which fixed it but I'm not sure about implementation. The following change doesn't work properly and changes the value in case of multiple select.
```js
function selectOption(select, option) {
  fireEvent.change(select, { target: { value: option.value } });
//...
}
```
Thx

### DIFF
--- a/__tests__/react/selectoptions.js
+++ b/__tests__/react/selectoptions.js
@@ -199,6 +199,27 @@ describe("userEvent.selectOptions", () => {
     expect(getByTestId("val3").selected).toBe(false);
   });
 
+  it("should fire onChange event on a SELECT element", () => {
+    const onChangeHandler = jest.fn();
+
+    const { getByTestId } = render(
+      <select data-testid="element" onChange={onChangeHandler}>
+        <option data-testid="val1" value="1">
+          1
+        </option>
+        <option data-testid="val2" value="2">
+          2
+        </option>
+        <option data-testid="val3" value="3">
+          3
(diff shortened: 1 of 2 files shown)


Gpx:
Hi, thanks for the report. I'll take a look at it

DanweDE:
:wave: Any updates on this? Is there a workaround?

happycollision:
Implementation wise, I think you can do `fireEvent.change(element)` without additional values. This should ensure that the event uses the current values on the element. So if you fire it after each selection, that should do it. Someone can correct me if I am wrong here.

So at https://github.com/testing-library/user-event/blob/master/src/index.js#L159 you'd add the step of `fireEvent.change(element)` in the `forEach` block after `selectOption`, and a couple lines down, you'd add the same thing after the single instance of `selectOption`

bogdanbodnar:
@happycollision, thanks, It looks to be working. It was a long time ago and I'm a little bit out of context now, thus it would be nice if somebody checks it one more time, then it can be merged.

pietmichal:
@bogdanbodnar I've tested your changes within my test suite and onChange events were fired as they should.

Gpx:
Merged in #158. Thank you very much to everyone!

PR #134, opened 2019-06-25
chore: added TypeScript type definitions

weyert:
Added TypeScript definitions for `@testing-library/user-event`-package


Gpx:
Thank you!

PR #136, opened 2019-07-05
Switch @testing-library/dom to be a peerDep + devDep

### LINKED ISSUE #135: Allow for non-fixed version of @testing-library/dom
Hi, thanks for this package, it's pretty awesome :eyes: :tada: 

I was wondering if the `package.json` could be updated to have a semver _range_ for the `@testing-library/dom` dependency, rather than a fixed version: https://github.com/testing-library/user-event/blob/master/package.json#L56

Suggestion:

```diff
-    "@testing-library/dom": "5.1.1"
+    "@testing-library/dom": "^5.1.1"
```

This way I don't have to have multiple versions of `@testing-library/dom` installed, and you don't have to constantly update the version in your `package.json` :smile: 

Happy to submit a PR if you want

calebeby:
Closes #135

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "travis-deploy-once": "5.0.11"
   },
   "peerDependencies": {
-    "@testing-library/dom": "^5.1.1"
+    "@testing-library/dom": ">=5"
   },
   "husky": {
     "hooks": {
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "@babel/core": "7.4.5",
     "@babel/preset-env": "7.4.5",
     "@babel/preset-react": "7.0.0",
+    "@testing-library/dom": "^5.1.1",
     "@testing-library/react": "8.0.1",
     "all-contributors-cli": "6.6.1",
     "babel-core": "7.0.0-bridge.0",
@@ -52,8 +53,8 @@
     "semantic-release": "15.13.12",
     "travis-deploy-once": "5.0.11"
   },
-  "dependencies": {
-    "@testing-library/dom": "5.1.1"
+  "peerDependencies": {
+    "@testing-library/dom": "^5.1.1"
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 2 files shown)


NMinhNguyen:
Strictly speaking, this could be considered a breaking change - because what was being installed before, now has to be installed separately. You might need to update the docs to reflect the peer dependency requirement as well, similar to how React Testing Library mentions `react` and `react-dom` as `peerDependencies`: https://github.com/testing-library/react-testing-library#installation

calebeby:
Maybe we could keep it as a dependency, but switch it to a version range, like I originally suggested, so it's not a breaking change 😄

NMinhNguyen:
@kentcdodds @Gpx do you guys have a preference on which approach to take here? :slightly_smiling_face:

package.json
```diff
@@ -52,8 +53,8 @@
     "semantic-release": "15.13.12",
     "travis-deploy-once": "5.0.11"
   },
-  "dependencies": {
-    "@testing-library/dom": "5.1.1"
+  "peerDependencies": {
+    "@testing-library/dom": "^5.1.1"
```
kentcdodds:
```suggestion
    "@testing-library/dom": ">=5"
```

kentcdodds: Peer Dependencies should have a version range that's as permissible as possible. That way people can install their own version without conflicting with the one this package says it needs.

calebeby: @kentcdodds what if `@testing-library/dom` releases a v6 that has breaking changes that affect this package? That's why I went with a caret range. Maybe the caret range should be `^5.0.0` though?

kentcdodds: Chances are that a breaking change in `@testing-library/dom` wont impact this package. If it does, then we can fix this. The last thing I want is to add to the already ridiculous amount of warnings people get when they install deps in their project.

Gpx:
Thanks to everyone 🎉

PR #137, opened 2019-07-05
Remove duplicated change event for boolean inputs

calebeby:
For `<input type="checkbox">` and `<input type="radio">`, the `change` event is automatically fired when the value changes. So for these types of inputs, we don't need to manually fire the change event. This was the cause of the duplicated `change` events seen in #129. I assume that the reason why we weren't seeing the duplicated `change` events is because React checks if the value is actually different from the last value before passing the event to the user: https://reactjs.org/docs/dom-elements.html#onchange

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -38,7 +38,6 @@ function clickBooleanElement(element) {
   fireEvent.mouseDown(element);
   fireEvent.mouseUp(element);
   fireEvent.click(element);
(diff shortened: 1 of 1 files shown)


calebeby:
@Gpx Not sure what to do about the code coverage change :thinking:

Gpx:
Merged in #144 
Thank you 👍

PR #141, opened 2019-07-19
Fix onchange text input

calebeby:
The change event should only be fired when the value modification gets "committed". For text inputs, this means once the input gets blurred.

I used https://codesandbox.io/s/inputtypetext-keyboard-events-m5o2u to test real browser behavior

### DIFF
--- a/__tests__/type.js
+++ b/__tests__/type.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { cleanup, render, wait } from "@testing-library/react";
+import { cleanup, render, wait, fireEvent } from "@testing-library/react";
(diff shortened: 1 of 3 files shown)


Gpx:
Merged in #147

dilraj-vidyard:
Does this allow me to do:

```
const wrapper = mount(TaggableTextInput, {...}

const randomInput = wrapper.find('.random-input').element;
await user.type(randomInput, 'Random text here {{company}}.');
```

Because I am STILL not seeing changes to the input value... I even added like `@input=handleInput`` and `@keypress={handleKeypress}` and it doesn't work.

PR #145, opened 2019-07-20
test: 💍 add Vue tests for type event

Gpx:
Rebase of #129


Gpx:
Merging these tests. @afontcu it would be great if we can migrate also the remaining tests to Vue

afontcu:
I sure will! Thank you for this 🙌

Gpx:
Thank you

PR #147, opened 2019-07-20
Fix/onchange text input

Gpx:
Rebase of #141 

It looks like React handles the `change` event differently from what a browser does.

In a browser:

```
keydown
keypress
input
keyup
change
```

And `change` fires only after the input loses focus.

In React this is the sequence:

```
keydown
keypress
input
change
keyup
```

`change` fires without need to blur the input.

You can try both sequences [here](https://codesandbox.io/s/inputtypetext-keyboard-events-m5o2u) and [here](https://codesandbox.io/s/laughing-wildflower-2qvmh).

@calebeby what do you think? Does it make sense to you?

### DIFF
--- a/__tests__/react/type.js
+++ b/__tests__/react/type.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { cleanup, render, wait } from "@testing-library/react";
+import { cleanup, render, wait, fireEvent } from "@testing-library/react";
 import "@testing-library/jest-dom/extend-expect";
 import userEvent from "../../src";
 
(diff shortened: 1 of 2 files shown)


calebeby:
@Gpx In react for text inputs, `onChange` and `onInput` are exactly the same. React never actually listens for the DOM `change` event for text inputs. If you do `onChange` in React on a text input it is actually listening for the `input` event. So if `user-event` fires a DOM `change` event the same way a real user does, it won't affect React

PR #148, opened 2019-07-21
Fix label.querySelector for select

skywickenden:
When testing SELECT elements, If a label uses `htmlFor` then everything works fine, but if instead the label contains the form element inside it then an error is thrown.

    const { getByTestId } = render(
      <form onSubmit={onSubmit}>
        <label>
          Example Select
          <select data-testid="element">
            <option data-testid="val1" value="1">
              1
            </option>
            <option data-testid="val2" value="2">
              2
            </option>
            <option data-testid="val3" value="3">
              3
            </option>
          </select>
        </label>
      </form>
    );

    userEvent.selectOptions(getByTestId("element"), "2");

    expect(getByTestId("val1").selected).toBe(false);
    expect(getByTestId("val2").selected).toBe(true);
    expect(getByTestId("val3").selected).toBe(false);

Error is : `TypeError: Cannot read property 'focus' of null` 

    at clickLabel (node_modules/@testing-library/user-event/dist/index.js:38:11)
    at clickElement (node_modules/@testing-library/user-event/dist/index.js:73:20)
    at Object.selectOptions (node_modules/@testing-library/user-event/dist/index.js:198:5)

This already seems to be handled for INPUT and TEXTAREA. This pull request adds SELECT plus a couple of tests to check both usages work.

### DIFF
--- a/__tests__/react/selectoptions.js
+++ b/__tests__/react/selectoptions.js
@@ -171,4 +171,60 @@ describe("userEvent.selectOptions", () => {
     expect(getByTestId("val2").selected).toBe(false);
     expect(getByTestId("val3").selected).toBe(true);
   });
+
+  it("sets the selected prop on the selected OPTION using htmlFor", () => {
+    const onSubmit = jest.fn();
+
+    const { getByTestId } = render(
+      <form onSubmit={onSubmit}>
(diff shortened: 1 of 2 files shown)


Gpx:
Merged in #150 
Thank you 🎉

PR #151, opened 2019-07-23
Add additional tests for Vue

afontcu:
Hi! 👋 this is a draft PR to add additional tests for Vue.

- [x] tests for dblclick
- [x] tests for selectoptions
- [ ] tests for type

I can't seem to find the way to make the `preventDefault` test pass ([link](https://github.com/testing-library/user-event/compare/master...afontcu:additional-vue-tests?expand=1#diff-c15ea404ef18eb0acc40c894ece19235R41)). Any help would be greatly appreciated!

Apart from that, the only difference I noticed from React tests is that `focus` event isn't being emitted twice on selectoptions ([link](https://github.com/testing-library/user-event/compare/master...afontcu:additional-vue-tests?expand=1#diff-4e8bbadf93b5bc4474efd2532de7c58eR55)).

### DIFF
--- a/__tests__/react/selectoptions.js
+++ b/__tests__/react/selectoptions.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { render, cleanup, fireEvent } from "@testing-library/react";
+import { render, cleanup } from "@testing-library/react";
 import "@testing-library/jest-dom/extend-expect";
 import userEvent from "../../src";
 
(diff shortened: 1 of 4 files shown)


Gpx:
@afontcu you should be able to create new branches directly in this repo, let me know if that's not the case. Please use `npm run cm` to generate commit messages so that the changelog is created automatically

afontcu:
@Gpx okay, will do! I'm gonna create a proper PR once I've found a way to fix the missing bits :)

thanks!

PR #153, opened 2019-07-31
test: 💍 add Vue tests for type, dblclick, selectoptions

afontcu:
Hi! 👋 this is a PR to add additional tests for Vue.

- [x] Add tests for `dblclick`
- [x] Add tests for `selectoptions`
- [x] Add tests for `type`
- [x] (upstream) fix VTL so it doesn't mutate DOM Testing Library' `fireEvent` methods.
- [x] Fix `preventDefault` test for `type`

~I can't seem to find the way to make the `preventDefault` test pass ([link](https://github.com/testing-library/user-event/compare/add-vue-tests?expand=1#diff-c15ea404ef18eb0acc40c894ece19235R47-R48)). Any help would be greatly appreciated!~

Apart from that, I've only noticed small differences from React tests:
* `focus` event isn't being emitted twice on selectoptions ([link](https://github.com/testing-library/user-event/compare/add-vue-tests?expand=1#diff-4e8bbadf93b5bc4474efd2532de7c58eR55)).
* input/change events ([link](https://github.com/testing-library/user-event/compare/add-vue-tests?expand=1#diff-c15ea404ef18eb0acc40c894ece19235R77-R80)).

I guess @calebeby might want to take a look at this, too :)

### DIFF
--- a/__tests__/vue/dblclick.js
+++ b/__tests__/vue/dblclick.js
@@ -1,0 +1,125 @@
+import { render, cleanup } from "@testing-library/vue";
+import "@testing-library/jest-dom/extend-expect";
+import userEvent from "../../src";
+
+afterEach(cleanup);
+
+describe("userEvent.dblClick", () => {
+  it.each(["input", "textarea"])(
+    "should fire the correct events for <%s>",
+    type => {
+      const events = [];
+      const eventsHandler = jest.fn(evt => events.push(evt.type));
+      const { getByTestId } = render({
+        render: function(h) {
+          return h(type, {
+            attrs: {
+              "data-testid": "element"
+            },
+            on: {
+              mouseover: eventsHandler,
+              mousemove: eventsHandler,
+              mousedown: eventsHandler,
+              focus: eventsHandler,
+              mouseup: eventsHandler,
+              click: eventsHandler,
+              dblclick: eventsHandler
+            }
+          });
+        }
+      });
+
+      userEvent.dblClick(getByTestId("element"));
+
+      expect(events).toEqual([
+        "mouseover",
+        "mousemove",
+        "mousedown",
+        "focus",
+        "mouseup",
+        "click",
+        "mousedown",
+        "mouseup",
+        "click",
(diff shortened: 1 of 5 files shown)


calebeby:
Yeah I can take a look this afternoon (PST)

calebeby:
Sorry, I totally forgot to look at this, I finally remembered today!

> I can't seem to find the way to make the preventDefault test pass
As far as I can tell this is because `@testing-library/vue` overrides `@testing-library/dom`'s `fireEvent.*` methods, making them return `Promise<void>` rather than `boolean`. So whenever we import `@testing-library/vue`, `user-event` has no way of knowing whether the fired events were canceled via `preventDefault()`

Relevant `@testing-library/vue` code: https://github.com/testing-library/vue-testing-library/blob/master/src/vue-testing-library.js#L96

calebeby:
It's frustrating that `@testing-library/vue` mutates the `fireEvent` provided by `@testing-library/dom` rather than leaving the original object alone and re-exporting its own modified copy. Maybe that change could be made to `@testing-library/vue`?

afontcu:
Thank you for your insights. It makes sense. In order to test it out, I've awaited all `fireEvent` calls in user-event, and then awaited `userEvent.type`. It made the failing test to pass.

I'm gonna open up a PR in VTL to propose the change you suggested. I just noticed that [RTL is exporting its own version](https://github.com/testing-library/react-testing-library/blob/master/src/pure.js#L106) of `fireEvent`, instead of polluting the one from DTL.

I'd love to know, too, whether if making `fireEvent` return a promise is _totally_ necessary due to Vue internals. There might be some alternative that keeps VTL public API as similar as possible to RTL's and DTL's.

Thanks! 🙌

afontcu:
Update: VTL 2.0 has been released, and after the updated, `preventDefault` test started passing 🙌 huge props to @calebeby for pointing out the design flaw in VTL.

Gpx:
Thank you both for this 🎉

PR #155, opened 2019-08-06
Add ops.target to selectOptions

michalchmura:
In some cases, the `option.value` and `option.text` are not tightly coupled.

Some developers want to hide the internal `option.value` and show nicely-formatted text.

This PR will provide a way to specify how the options should be extracted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ one character at the time. `false` is the default value.
 are typed. By default it's 0. You can use this option if your component has a
 different behavior for fast or slow users.
 
-### `selectOptions(element, values)`
+### `selectOptions(element, values, [options])`
 
 Selects the specified option(s) of a `<select>` or a `<select multiple>`
 element.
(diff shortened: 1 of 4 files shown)


michalchmura:
I tried recreating the PR with `npm run cm` from a new branch from within this repo but I don't have the write permissions 👎

dougbacelar:
I think this makes sense because:
* You might have some option.text that you want to assert on e.g. translations, dynamically generated text
* Users don't interact with select values

I think its important to at least display a warning when trying to select an option that doesn't exist--as its very likely a bug.

--- 
An alternative to that, would be `selectOptions` also accepting the option nodes themselves--like the single `selectOption` function:

```js
userEvent.selectOptions(getByLabelText('Select'), [getByText('first option')])
```

That way you also get an error from `getByText('first option')` if `first option` doesn't exist. 

---

Issue https://github.com/testing-library/user-event/issues/203 has nice comments and is related to this PR.

kentcdodds:
This is managed best in https://github.com/testing-library/user-event/pull/297 so I'm going to close this. Thanks!

PR #164, opened 2019-09-09
Fix example in README for `type` method

zperrault:
Was reading the README and noticed that this example destructures `getByText` but uses `getByTestId`.

Wasn't able to find a `CONTRIBUTORS` file so I'm not sure if you're accepting PRs but I hope this is helpful! Happy to make any changes to commit format etc.

And finally a huge thanks to this project and all of the `@testing-library` ecosystem!


Gpx:
Thank you!

Issue #167, opened 2019-09-13
userEvent.tab


benmonro:
Add the ability to simulate a user tabbing through a UI for RTL: `userEvent.tab(direction)`

more of a question to see if you guys would be open to me opening a PR for `userEvent.tab` 
from what I've read in threads on RTL and JSDOM, the ability to tab is not really supported by JSDOM.  However, I came up with a working solution that simulates tabbing.  I'd love to open a PR and contribute here, but before I do wanted to check to make sure you guys would be open to it.

alexkrolick:
Would this be like click where it takes a target element? So it would be like "press tab as many times as needed to get from the current focus to the target"?

benmonro:
my thought is that this would just be a generic tab key event.  My first pass (not accounting for direction) will be based on this code (though I will add more to it for the purposes of `user-event`)

```
export function fireTabKeyPress() {
    const focusableElements = document.querySelectorAll('input, button, select, textarea, a, [tabindex]');
    const list = Array.prototype.filter.call(focusableElements, function (item) { return item.getAttribute('tabindex') !== '-1'; });
    const index = list.indexOf(document.activeElement);
    const next = list[index + 1] || list[0];

    if (next.getAttribute('tabindex') === null) {
        next.setAttribute('tabindex', '0'); // jsdom requires tabIndex=0 for an item to become 'document.activeElement' (the browser does not)
    }
    next.focus();
}```

benmonro:
@alexkrolick would love to get some feedback on my PR.  So you think it can be merged?

benmonro:
@gpx I didn't see that, thanks for getting back tome.

I believe it's important to support a11y for keyboard only users. I live by the guiding principle of testing library:
>We try to only expose methods and utilities that encourage you to write tests that closely resemble how your web pages are used.

since KB only users will tab through elements in your pages/components, I'm of the opinion that writing tests in this way is very important.

The implementation I put together in #169 is working for us at Walmart Labs.  Would love to have you poke holes in that and let me know if there are any shortcomings, etc.

alexkrolick:
@benmunro I don't have too much context on this codebase. The implementation seems good to me.

To @Gpx's point, I am not sure on the scope of user-event. The closest thing is probably PhantomJS (discontinued 2018), which did have the goal of emulating browser behavior for user interactions rather than plain event model of jsdom: https://phantomjs.org/api/webpage/method/send-event.html

[...]

PR #169, opened 2019-09-13
feat: userEvent.tab

benmonro:
implements https://github.com/testing-library/user-event/issues/167

### DIFF
--- a/__tests__/react/tab.js
+++ b/__tests__/react/tab.js
@@ -33,7 +33,8 @@ describe("userEvent.tab", () => {
 
         userEvent.tab();
 
-        expect(document.activeElement).toBe(checkbox);
+         // cycle goes back to first element
+         expect(document.activeElement).toBe(checkbox); 
     });
 
     it("should go backwards when shift = true", () => {
--- a/__tests__/react/tab.js
+++ b/__tests__/react/tab.js
@@ -6,7 +6,7 @@ import userEvent from "../../src";
 afterEach(cleanup);
 
 describe("userEvent.tab", () => {
-    it("should go from one element to the next, then back to the 1st", () => {
+     it("should cycle elements in document tab order", () => {
         const { getAllByTestId } = render(
             <div>
                 <input data-testid="element" type="checkbox" />
--- a/__tests__/react/tab.js
+++ b/__tests__/react/tab.js
@@ -1,9 +1,7 @@
 import React from "react";
-import { render, cleanup } from "@testing-library/react";
+import { render } from "@testing-library/react";
 import "@testing-library/jest-dom/extend-expect";
 import userEvent from "../../src";
-
-afterEach(cleanup);
 
 describe("userEvent.tab", () => {
      it("should cycle elements in document tab order", () => {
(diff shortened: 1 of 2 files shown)


benmonro:
@alexkrolick this should be good to go.  welcome your feedback.  I was on the fence about adding a warning for tabIndex w/ a positive value, but decided that there are lint rules for that, i could be convinced either way though.  let me know.

__tests__/react/tab.js
```diff
@@ -0,0 +1,160 @@
+import React from "react";
+import { render, cleanup } from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
+import userEvent from "../../src";
+
+afterEach(cleanup);
+
+describe("userEvent.tab", () => {
+    it("should go from one element to the next, then back to the 1st", () => {
```
alexkrolick:
```suggestion
     it("should cycle elements in document tab order", () => {
```

__tests__/react/tab.js
```diff
@@ -0,0 +1,160 @@
...
+
+        expect(document.activeElement).toBe(radio);
+
+        userEvent.tab();
+
+        expect(document.activeElement).toBe(number);
+
+        userEvent.tab();
+
+        expect(document.activeElement).toBe(checkbox);
```
alexkrolick:
```suggestion
         // cycle goes back to first element
         expect(document.activeElement).toBe(checkbox); 
```

src/index.js
```diff
@@ -224,6 +224,30 @@ const userEvent = {
...
+    });
+    const index = list.indexOf(document.activeElement);
+
+    let nextIndex = shift ? index - 1 : index + 1;
+    let defaultIndex = shift ? list.length - 1 : 0;
+
+    const next = list[nextIndex] || list[defaultIndex];
+
+    if (next.getAttribute("tabindex") === null) {
+      next.setAttribute("tabindex", "0"); // jsdom requires tabIndex=0 for an item to become 'document.activeElement' (the browser does not)
```
kentcdodds: Should we unset this after `next.focus()`? It would be good if this function could be as pure as possible.

benmonro: @kentcdodds good call, i'd thought about this but was probably driving and forgot to add it. lol

done

__tests__/react/tab.js
```diff
@@ -0,0 +1,161 @@
+import React from "react";
+import { render, cleanup } from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
+import userEvent from "../../src";
+
+afterEach(cleanup);
```
Gpx: This should not be needed anymore with the latest version of RTL

Gpx:
Can you please rebase and use `npm commit` so that we can release a new version?

benmonro:
@Gpx not sure what `npm commit` is, but afaict this should be good to merge.

kentcdodds:
@gpx, if you use GitHub's rebase and merge feature then you can write the commit message yourself and not worry about the commit messages that people write themselves. Saves everyone a lot of time and you don't need commitizen 👍

PR #184, opened 2019-11-11
Adds tab function to docs and typings

GentlemanHal:
Just a simple patch to add the tab function to docs and typings.


typings/index.d.ts
```diff
@@ -15,6 +15,7 @@ declare const userEvent: {
         text: string,
         userOpts?: IUserOptions
     ) => Promise<void>;
+    tab: (userOpts?: { shift: boolean }) => void;
```
YuanchengWu: This type is no longer up to date. Could you add `focusTrap` to it?
better yet, define an interface and add defaults to the options

```ts
interface TabUserOptions {
  shift: boolean = false;
  focusTrap: Document = document;
}

declare const userEvent: {
  ...
  tab: (userOpts?: TabUserOptions) => void;
};

GentlemanHal: Should the type of `focusTrap` be `TargetElement` which is defined as `type TargetElement = Element | Window;`?

GentlemanHal: Or maybe it needs to be `Document | Element`?

GentlemanHal: Looking at `lib.dom.d.ts` `querySelectorAll` is defined in `ParentNode` which `Document` and `Element` both extend, so `Document | Element` seems like the correct choice.

Gpx: Are you happy with the current state?

GentlemanHal: I am 👍

Gpx: Cool, then I'm going to merge it

Issue #161, opened 2019-08-27
should type be constrained by maxLength?


afontcu:
Is `maxLength` an attribute `user-event` should take into account?

If so, the following test should pass:

```js
// __tests__/vue/type.js

test("maxlength", () => {
  const input = jest.fn();

  const maxlength = 10;

  const { getByTestId } = renderComponent('input',
    { input },
    { maxlength }
  );

  const text = "Hello, world!";
  userEvent.type(getByTestId("input"), text);

  expect(input).toHaveBeenCalledTimes(maxlength);
  expect(getByTestId("input")).toHaveProperty("value", text.slice(0, maxlength));
});
```

I'll be happy to implement it if you think it makes sense.

afontcu:
As far as I can tell, both are failing ("hello world!" is 13 characters long, so event is being fired 13 times).

weyert:
Yes, the type-helper method doesn't check the `maxLength`-attribute this time. Personally, I have more issues with the lack of diacritics support or special characters like {tab} or {shift} etc. I hope I need this requirement soon at work so I can work on it. Only it is a bit problematic because of all the different keyboard layouts available throughout the world.

weyert:
Sounds good to me. I can work on something tomorrow at work

weyert:
Woops, got sidetracked by a hackaton at work. I will pick this up on Monday :)

carlos-cne:
Here still not working.
```javascript
it('Should input have a 60 max length even try to add more', async () => {
    render(<CreateBanner />);

    const inputValue = screen.getByPlaceholderText(/Título do banner/);

    await userEvent.type(
      inputValue,
      'A text that has more than sixty characters to test input limit'
    );

    expect(inputValue.value).toHaveLength(60);
  });
});
``` 
If I'm doing wrong, please let me know the right way.

gautamkrishnar:
Working fine for me. I am using the version 8.1.0 of the user-event library.

kentcdodds:
This should be working properly now.

PR #185, opened 2019-11-19
Handle maxLength for userEvent.type

klujanrosas:
This is related to #161 .
I took the liberty to implement it. I look forward to your comments or suggestions.

### DIFF
--- a/__tests__/vue/type.js
+++ b/__tests__/vue/type.js
@@ -145,4 +145,31 @@ describe("userEvent.type", () => {
       expect(input).toHaveBeenCalledTimes(1);
     }
   );
+
+  it.each(["input", "textarea"])(
+    "should type text in <%s> up to maxLength if provided",
+    type => {
+      const input = jest.fn();
+      const maxLength = 10;
+
+      const { getByTestId } = renderComponent(type, { input }, { maxLength });
+
+      const text = "superlongtext";
+      const slicedText = text.slice(0, maxLength);
+
+      const inputEl = getByTestId("input");
+
+      userEvent.type(inputEl, text, {
+        allAtOnce: true
+      });
+      expect(inputEl).toHaveProperty("value", slicedText);
+
+      inputEl.value = "";
+      input.mockClear();
+
+      userEvent.type(inputEl, text);
+      expect(inputEl).toHaveProperty("value", slicedText);
+      expect(input).toHaveBeenCalledTimes(slicedText.length);
+    }
+  );
 });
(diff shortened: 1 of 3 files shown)


klujanrosas:
@afontcu I also can't think of a reason why they would fail, but one can never be too cautious!

I will find some time to add those tests as well, thank you

klujanrosas:
@afontcu sorry to bother you, could this be reviewed again?

kentcdodds:
Are we sure the browser doesn't continue to fire input events even though the maxlength is reached? We don't have to actually update the value, but we should make sure we fire all the same events the browser does.

calebeby:
@kentcdodds I just tested an input with a maxlength, it looks like it only fires `input` events till the max length is reached.

kentcdodds:
Cool. Are we still firing keyup/keydown events?

calebeby:
I'm pretty sure that we should be, but I'm not sure whether it does in this PR

klujanrosas:
@kentcdodds @calebeby  that makes sense, I will update this PR so it still fires keyup/keydown events, thank you for your comments

carlos-cne:
Hi! I arrived here because I have an input with a max-length attribute.
When I call userEvent.type(...) with a string more than max-length, it does not work and I got a string with more than max-length.

I'm testing if input.value is equal to max-length
`expect(inputValue.value).toHaveLength(60);`

kentcdodds:
Hi @carlos-cne,

Yes, once this PR is updated and then merged that issue should go away.

klujanrosas:
@Gpx @kentcdodds I've updated this to reflect what was suggested in the comments above, let me know if there's anything else that I should cover or change :hatching_chick:

__tests__/vue/type.js
```diff
@@ -145,4 +145,48 @@ describe("userEvent.type", () => {
...
+
+      userEvent.type(inputEl, text, {
+        allAtOnce: true
+      });
+
+      expect(inputEl).toHaveProperty("value", slicedText);
+      expect(input).toHaveBeenCalledTimes(1);
+      expect(keydown).not.toHaveBeenCalled();
+      expect(keypress).not.toHaveBeenCalled();
+      expect(keyup).not.toHaveBeenCalled();
```
kentcdodds: Wait, shouldn't these all be called several times?

klujanrosas: not for `allAtOnce` afaik

kentcdodds: Oh

Oh! lol, my bad.

kentcdodds:
@all-contributors please add @klujanrosas for code and tests

PR #193, opened 2019-12-12
feat: focus trap on tab

benmonro:
This feature allows you to easily test tabbing in focus traps. 

example:

```userEvent.tab({focusTrap: containerElement})```

also adds tab to readme and converts expects to use jest dom for tabs

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ const { getByTestId } = render(
 );
 
 userEvent.selectOptions(getByTestId("select-multiple"), ["1", "3"]);
-
+j
 expect(getByTestId("val1").selected).toBe(true);
 expect(getByTestId("val2").selected).toBe(false);
 expect(getByTestId("val3").selected).toBe(true);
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -167,6 +167,55 @@ expect(getByTestId("val3").selected).toBe(true);
...
 
+### `tab({shift, focusTrap})`
+
+Fires a tab event changing the document.activeElement in the same way the
+browser does.
+
+Options:
+
+- `shift` (default `false`) can be true or false to invert tab direction.
+- `focusTrap` (default `document`) a container element to restrict the tabbing
```
kentcdodds: What's the motivation here? Doesn't this create a [test user](https://kcd.im/test-user) scenario?

benmonro: @kentcdodds so let me start with the motivation for a focus trap option.  We are using focus-trap-react which keeps tabbing confined within an element.  Because jsdom doesn't have any way to deal with tabbing, these types of components won't work in jsdom.

So since jsdom doesn't seem to support tabs yet, focus trap basically won't work with userEvent.tab().  So does this introduce a test user?  Maybe.  But my hope is that jsDom can add support for tab and this entire feature can basically just adopt that.  In the meantime this seems like the least bad option.  That said if you have another way of writing a test that can let the user tab through a focus trap, I'd love to get some better ideas.

Here's the code for focus-trap which ftR uses under the hood.  
https://github.com/davidtheclark/focus-trap/blob/master/index.js

kentcdodds: That makes sense. I would add a note about the original intent for the option and why it should be used sparingly.

Then you've got a 👍 from me :) Thanks!

benmonro: Will do!  Thx Kent!

benmonro:
@Gpx any chance you can merge this?  we could really use it. :)

afontcu:
Vue tests are missing 🙈 no problem though, I guess I can work on them even after merging! 😄

kentcdodds:
I'm kinda thinking we should just ditch the framework specific stuff and use raw DOM stuff like in DOM Testing Library.

calebeby:
> I'm kinda thinking we should just ditch the framework specific stuff and use raw DOM stuff like in DOM Testing Library.

Fully agree 👍

afontcu:
> I'm kinda thinking we should just ditch the framework specific stuff and use raw DOM stuff like in DOM Testing Library.

Yeah, makes sense.

PR #196, opened 2019-12-17
Fix/allow selecting options in optgroups

jordyvandomselaar:
I'm opening this PR to solve an issue where selectOptions does not work if the options are nested inside optgroup elements. I think querySelectorAll would solve this and possibly other cases where nesting could be an issue.


kentcdodds:
@all-contributors please add @jordyvandomselaar for code and tests

PR #200, opened 2019-12-19
fix: updated the documentation for the `type` method

### LINKED ISSUE #198: README needs to be updated for user-event
The [`type`](https://github.com/testing-library/user-event#typeelement-text-options) method is [async](https://github.com/testing-library/user-event/blob/master/src/index.js#L176), but the documentation doesn't reflect it.

maheshjag:
Fixes #198

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ test("double click", () => {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


Gpx:
Thanks 🙇

PR #205, opened 2020-01-03
fix: account for sorting bug in node 10

benmonro:
prior to node 11, Array.sort was unstable for arrays w/ length > 10.
https://twitter.com/mathias/status/1036626116654637057
for this reason, the tab() function needs to account for this in it's sorting.
for example under node 10 in this test:
> 'abcdefghijklmnopqrstuvwxyz'.split('').sort(() => 0).join('')
will give you 'nacdefghijklmbopqrstuvwxyz'

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,7 @@
 import { fireEvent } from "@testing-library/dom";
 
 function wait(time) {
-  return new Promise(function(resolve) {
+  return new Promise(function (resolve) {
     setTimeout(() => resolve(), time);
   });
 }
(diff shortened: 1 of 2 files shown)


benmonro:
@kentcdodds @Gpx found a pretty little stank nasty in node 10 and prior.  

code cov is off by a branch but it's 100% if you count the fact that it's tested in each env on CI. ;)

src/index.js
```diff
@@ -12,6 +12,23 @@ function findTagInParents(element, tagName) {
   return findTagInParents(element.parentNode, tagName);
 }
 
+let sortListByTabIndex;
+let nodeVersion = +process.version.slice(1).split('.')[0];
+if (nodeVersion < 11) {
```
kentcdodds: Is there any way we can make this consistent regardless of node version? I'm not super thrilled about the added complexity here 😬

benmonro: Yeah I thought about that but it would mean an additional map over the entire list.  Felt like it would be punishing node 12 for the sins of node 10.  Also didn't think it warranted adding a separate sort just because of this. As always open to other options I might be missing

kentcdodds: I don't think we should be worried about perf at this stage... We're paying the price of complexity for an unnoticeable perf gain 🤷‍♂️

benmonro: alrighty.  pushed a version that just has a single sort @kentcdodds

PR #208, opened 2020-01-09
fix(click): click no longer blurs when focus changes during onClick event

jmcriffey:
I came across this issue while testing form components. It is common for a form to focus inputs that are invalid, and I was seeing strange behavior where I couldn't invalidate and focus the same input element twice. I did a bit of digging and see that events call blur after the click event is fired, so I believe we can fix this by checking to see if focus has already changed before calling blur.

### DIFF
--- a/__tests__/react/click.js
+++ b/__tests__/react/click.js
@@ -198,6 +198,32 @@ describe("userEvent.click", () => {
     expect(b).toHaveFocus();
   });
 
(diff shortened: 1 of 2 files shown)


kentcdodds:
@all-contributors please add @jmcriffey for code and tests

benmonro:
hi @jmcriffey  thanks for this. hope all is well

PR #216, opened 2020-02-09
fix: export typings

timdeschryver:
I was trying to export the `tab` function, but received the following while trying to build:

```
An unhandled exception occurred: projects/testing-library/src/lib/user-events/tab.ts:8:7 - error TS4023: Exported variable 'tab' has or is using name 'ITabUserOptions' from external module "node_modules/@testing-library/user-event/typings/index" but cannot be named.
```

After adding the exports in my local node_modules, I was possible to build my project.


Gpx:
Thanks

PR #218, opened 2020-02-13
fix(190): Add blur events in correct order

jsmapr1:
Blur occurs after mouse down, but before click.
I can't find any official documentation, but here's a nice code pen (not using library):
https://codepen.io/mudassir0909/pen/eIHqB

This is important because an `onBlur` event in React happens before `onClick` and if it rerenders the element, the `onClick` event never happens.

This covered a case from a previous PR #208, so I remove the related code.

Let me know if you need anything else. 🎉

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -41,11 +41,12 @@ function clickBooleanElement(element) {
   fireEvent.click(element);
 }
 
-function clickElement(element) {
+function clickElement(element, previousElement) {
   fireEvent.mouseOver(element);
   fireEvent.mouseMove(element);
   const continueDefaultHandling = fireEvent.mouseDown(element);
   if (continueDefaultHandling) {
+    previousElement && previousElement.blur();
     element.focus();
   }
   fireEvent.mouseUp(element);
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -55,11 +56,14 @@ function clickElement(element) {
...
 }
 
-function dblClickElement(element) {
+function dblClickElement(element, previousElement) {
   fireEvent.mouseOver(element);
   fireEvent.mouseMove(element);
-  fireEvent.mouseDown(element);
-  element.focus();
+  const continueDefaultHandling = fireEvent.mouseDown(element);
+  if (continueDefaultHandling) {
```
jsmapr1: Added to be consistent with single click

src/index.js
```diff
@@ -101,15 +105,6 @@ function fireChangeEvent(event) {
   event.target.removeEventListener("blur", fireChangeEvent);
 }
 
-function blurFocusedElement(element, focusedElement, wasAnotherElementFocused) {
```
jsmapr1: Now that we focus before click, we don't need to blur elements at the end of the actions.

Gpx:
Merged in #218, thanks!

PR #227, opened 2020-02-21
Inline fire-event for zero dependencies

Raynos:
The other library is very very large and this library is very small.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,9 +56,6 @@
     "vue": "^2.6.11",
     "vue-template-compiler": "^2.6.11"
   },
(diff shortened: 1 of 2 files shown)


Raynos:
Ugh there are fields on the function so it's not as easy as I thought -.-

Raynos:
I am still :+1: for inlining this entire file https://github.com/testing-library/dom-testing-library/blob/master/src/events.js

Gpx:
The reason we're not doing this is that user-event might move to the main library at some point

PR #228, opened 2020-02-27
src: Add support for ctrl click.

Raynos:
We pass the `init` argument through so that the underlying
click event can be created with ctrlKey set to true if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,15 @@ test("click", () => {
 });
 ```
 
+You can also ctrlClick / shiftClick etc with
+
+```js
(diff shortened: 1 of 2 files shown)


Raynos:
This is a small change but a pretty great improvement for managing click events.

Is there any appetite for reviewing / merging this PR ?

cc @Gpx

Raynos:
Is there anything I can do to get this reviewed or merged ? cc @alexkrolick

Raynos:
I am going to close out this PR as it has gotten no response.

Anyone who wants to use this feature can check out http://npmjs.org/raynos-user-event

Issue #178, opened 2019-10-03
Typing in an input or textarea element should make it the element the activeElement. 


lothisoft:
An input or textarea element becomes the activeElement of the document (gets the input focus) when a user types into the field.  

Currently, after the type function executes on an active element, document.activeElement still points to document.body and the onFocus handler is not called.

riotrah:
Agreed, mid-type focus should be testable, or at least configurable so that it doesn't blur on finish

Gpx:
This looks like an easy fix, does anyone want to take it?

eps1lon:
> An input or textarea element becomes the activeElement of the document (gets the input focus) when a user types into the field.

It's the other way around. The active element is the target of keyboard interactions which is why I don't think type() should focus but rather it should check that the target is `document.activeElement || document.body || document.documentElement`

> The event target of a key event is the currently focused element which is processing the keyboard activity. [...] If no suitable element is in focus, the event target will be the HTML body element if available, otherwise the root element.

-- https://w3c.github.io/uievents/#events-keyboard-event-order

Otherwise this tool might hide bugs in your implementation if you expect that a previous event shifted focus.

kentcdodds:
Note: implementing this will be a breaking change.

NMinhNguyen:
Spoke to @eps1lon a bit about this on Twitter and I'm a little bit conflicted myself. On the one hand, currently user events like `click` and `tab` do focus the target element for you, and seems like Cypress' `type` [dispatches a click](https://docs.cypress.io/api/commands/type.html#When-element-is-not-in-focus) for you and additionally [supports `body`](https://docs.cypress.io/api/commands/type.html#Global-Shortcuts) for things like keyboard shortcuts.

On the other hand, typically a user has no way to focus an input element apart from clicking or tabbing into it first and only then typing (unless something else programmatically brought it into focus beforehand).

That being said, maybe there is value in mimicking Cypress here and performing a user click (or just focus)? As well as [throwing if the target is not focusable, unless it's `body`](https://github.com/cypress-io/cypress/blob/ee494d04ee532edec41f227ced0c639b3db78f5c/packages/driver/src/cy/keyboard.ts#L367-L386) as suggested by @kentcdodds on the PR.

kentcdodds:
> That being said, maybe there is value in mimicking Cypress here and performing a user click (or just focus)? As well as throwing if the target is not focusable, unless it's body as suggested by @kentcdodds on the PR.

[...]

PR #231, opened 2020-03-03
fix: Focus on type

doniyor2109:
PR for #178

### DIFF
--- a/__tests__/react/type.js
+++ b/__tests__/react/type.js
@@ -214,7 +214,7 @@ describe("userEvent.type", () => {
     const { getByTestId } = render(
       React.createElement(type, {
         "data-testid": "input",
-        onFocus: onFocus
+        onFocus
       })
     );
     userEvent.type(getByTestId("input"), "Text");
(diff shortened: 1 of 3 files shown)


NMinhNguyen:
@Gpx could you please review this? 🙂

__tests__/react/type.js
```diff
@@ -208,4 +208,17 @@ describe("userEvent.type", () => {
       expect(onKeyUp).not.toHaveBeenCalled();
     }
   );
+
+  it.each(["input", "textarea"])("should focus on type", type => {
+    const onFocus = jest.fn();
+    const { getByTestId } = render(
+      React.createElement(type, {
+        "data-testid": "input",
+        onFocus: onFocus
```
NMinhNguyen, line 217: nit: could use a shorthand for consistency with the rest of the test file

```suggestion
        onFocus
```

src/index.js
```diff
@@ -187,6 +187,8 @@ const userEvent = {
 
     const previousText = element.value;
 
+    element.focus();
```
kentcdodds, line 190: Rather than focusing I think we should verify that the `element === document.activeElement` and throw if not as mentioned here: https://github.com/testing-library/user-event/issues/178#issuecomment-608324794

doniyor2109: If we remove `element.focus()`, then how could we call `onFocus` callback will?

kentcdodds:
At some point this was implemented.

I apologize for not being more helpful here earlier. Thanks for your contribution anyway.

Issue #232, opened 2020-03-04
[feature request] type should support replacing the current input 


CodingDive:
An option like `replaceExisting: boolean` for the `type` event would be neat. It should mimic selecting all the content of the input field (sort of like `ctrl+a`) and only then start inserting the input into the text field.

kentcdodds:
This is now supported. You can either first call `clear()` or (probably better), do this:

```javascript
// select all:
input.setSelectionRange(0, input.value.length)
await userEvent.type(input, 'whatever new thing you want')
```

smriti-spotnana:
Why do we need an await before userEvent.type above, if there is no delay passed?

kentcdodds:
@smriti-spotnana you don't. My comment was made before that API change was made. The correct way to do this now is:

```js
// select all:
input.setSelectionRange(0, input.value.length)
userEvent.type(input, 'whatever new thing you want')
```

smriti-spotnana:
ok, thanks for the reply @kentcdodds 

my tests fail intermittently, if I don't use await when typing and then await waitFor(), waiting for a popup to show up. will you be able to comment on this? Thank You

Issue #226, opened 2020-02-20
Document how to delete text with type method


MattyBalaam:
The change to the type event to append to existing values has broken our tests which are written expecting to replace existing values.

Is it possible to document how to clear/delete existing values? I tried using the `\b` character but this resulted in some very strange behavior (I imagine this is not being translated until the test reporting.)

MattyBalaam:
The only reliable way I have at the moment is to clear using fireEvent. 

One other thing I tried was to use `userEvent.dblClick`to attempt to select all text.

bartgryszko:
A way this library is using in its own test files is by calling `input.value = '';`

https://github.com/testing-library/user-event/blob/38e3c227dd228c6ccc7624683eca3fae119a66e3/__tests__/react/type.js#L194

By calling that, you can basically simulate previous behavior.

If it's really important to test full user interaction, it would require writing custom `selectAll`/`clear` method I guess.

ivarprudnikov:
Slightly related #232

kentcdodds:
🎉 https://github.com/testing-library/user-event/issues/232#issuecomment-640791105

PR #233, opened 2020-03-11
feat: add userEvent.clear

wachunga:
This is a useful thing that Cypress provides. I've seen it requested a [few](https://github.com/testing-library/user-event/issues/232) [times](https://github.com/testing-library/user-event/issues/226) and recently missed it myself. 

This approach closely resembles how Cypress implements it: [{selectall}{backspace}](https://docs.cypress.io/api/commands/clear.html#Syntax). Because we don't have support for special keys yet, I simulated the relevant events but used double click for select all.

If you'd like, I can update the docs too.


wachunga:
I wrote this in a generic way to accommodate future support of a backspace key, but that's hurting the code coverage. (No way to test it because we don't expose that yet through userEvent.) Guess I'll simplify the backspace logic until we want `type` to support that.

wachunga:
@Gpx When you have a moment, could you review? Thanks!

Gpx:
Thank you 🙇

Issue #181, opened 2019-10-27
Simulate keyboard shortcuts


BenjaminBellamyJG:
Not sure if in this packages domain, but simulating shortcuts or hotkeys using this libary abstracting to something like `userEvent.shortcut('ctrl+c')` `userEvent.shortcut('ctrl+alt+enter')`

Gpx:
It sounds like something we are going to need at some point. Do you want to try implementing it?

Issue #172, opened 2019-09-19
Emulation of 'Backspace' and other non alphanumeric buttons pressing.


kernel72:
Hi, guys!

I have a next question, which has come during usage of this cool library. 
I have to simulate a user's "Backspace" button pressing and could not find how to do it with this library. 
Unfortunately, the native `testing-library` `fireEvent` is not works for me, because I have an input which filters pressed keys in `onKeyDown` event handler and it also has `onChange` event handler which makes additional validation and further component mutations.
I use `user-event` to check that both events  have been fires correclty and check the final value of input.

'Backspace' is also have to pass through this filter to erase typed data, but I couldn't find how to call it.

Could you please advice something?
Thank you.

Gpx:
Hi! I think this might be related to #31 which deals with adding support for all kinds of special characters.

Issue #31, opened 2018-10-21
Check type with special characters


Gpx:
`userEvent.type` writes text correctly. But I'm not sure how it behaves with special characters like `è` or 👍 .

It would be good to have tests covering these cases

weyert:
First of all thank you for merging my PR.

Yes, I have been thinking about this today. I think the behaviour of entering special character differ per language of keyboard being used. Do we just want to restrict this to the probably most used American keyboard? 

Beside of that do we want to enter capital letters correctly too? E.g. Shift + Letter? What about the enter key for a \n?

Gpx:
That's a good point. I think supporting every possible keyboard layout out there is too much work and out of the scope of this project. I would just assume that if a character is typed, it's because it was somewhere on the keyboard. That includes uppercase letters.

Cypress has support for special characters like enter. We could think to do the same https://docs.cypress.io/api/commands/type.html#Arguments

weyert:
While looking into a `react-testing-library` issue I discovered this package https://github.com/eventualbuddha/keysim.js . Maybe it's interesting to leverage for the `type`-behaviour?

Gpx:
The thing is this library is based on `fireEvent`, I don't know how that would fit with keysim

weyert:
Yeah, that's a good point. Maybe I borrow some ideas from that library :)

weyert:
Sorry, I haven't had a need yet to support special characters. I might consider with adding support {backspace} / {ctrl} {meta} etc first. I will have a closer look at how it works in Cypress again.

hatched-danny:
Hi, is there any progress on this? What have people been doing to simulate enter key presses if `    userEvent.type(textArea, "Testing!{enter}");` doesn't work?

duranmla:
There is no on going work related to support _"press enter"_ action? I will be happy to contribute if there is nothing in place yet, nevertheless, looks odd because it should be a very common usage case. Isn't? What does the people do now?

I have tried with:

```js
           fireEvent.keyDown(searchInput, {
                charCode: 13,
                code: 13,
                key: 'Enter',
                keyCode: 13,
            })
```

And it doesn't work to trigger the submit of the form I have

PR #235, opened 2020-03-15
feat: add special keys to userEvent.type()

malwilley:
Hello! I hope it's okay if I submit a PR for this feature. It seems to be highly requested (https://github.com/testing-library/user-event/issues/181, https://github.com/testing-library/user-event/issues/172, https://github.com/testing-library/user-event/issues/31) and it would make this library much more useful for some of my own use cases. Was hoping this would be a smaller change, but the scope got away from me a little bit!

This adds support for some new key presses in `userEvent.type()`. I am following the same syntax that cypress uses in [their .type() command](https://docs.cypress.io/api/commands/type.html#Syntax).

Commands added:
- `{enter}`
- `{esc}`
- `{backspace}`
- `{shift}`
- `{ctrl}`
- `{alt}`
- `{meta}`

This does not yet track cursor position or selection range, so keys like `{leftarrow}` have not been implemented. This work could be done in a followup PR, but I wanted to keep the scope of this relatively small so I stopped here. 

Code notes:

- The code for the `type()` command was getting pretty unwieldy so I've broken it out into its own file.
- Testing that events were being called with the right properties was resulting in some hairy test code. To combat this I've added `toHaveBeenCalledWithEventAtIndex` which helps assert that the event handlers are being called with the correct events.

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Typings (N/A)
- [x] Ready to be merged

### DIFF
--- a/src/type.js
+++ b/src/type.js
@@ -1,0 +1,182 @@
+import {fireEvent} from '@testing-library/dom'
+
+const wait = time => {
+  return new Promise(resolve => {
+    setTimeout(() => resolve(), time)
+  })
+}
+
+const specialKeyMap = {
+  '{enter}': {
+    key: 'Enter',
+    code: 13,
+    typed: '\n',
+  },
+  '{esc}': {
+    key: 'Escape',
+    code: 27,
+    skipPressEvent: true,
+  },
+  '{backspace}': {
+    key: 'Backspace',
+    code: 8,
+    inputType: 'deleteWordBackward',
+    skipPressEvent: true,
+  },
+  '{shift}': {
+    key: 'Shift',
+    code: 16,
+    modifier: 'shiftKey',
+  },
+  '{ctrl}': {
+    key: 'Control',
+    code: 17,
+    modifier: 'ctrlKey',
+  },
+  '{alt}': {
+    key: 'Alt',
+    code: 18,
+    modifier: 'altKey',
+  },
+  '{meta}': {
+    key: 'OS',
+    code: 91,
+    modifier: 'metaKey',
+  },
+}
+
+const parseIntoKeys = text =>
+  text
+    .split(/({[^{}]+?})/)
+    .map(part => {
+      if (specialKeyMap[part]) {
+        return {
+          typed: '',
+          skipPressEvent: false,
+          ...specialKeyMap[part],
+        }
+      }
+
+      return Array.from(part).map(char => {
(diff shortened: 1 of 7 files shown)


malwilley:
@kentcdodds @Gpx happy to see a flurry of activity on this repo!

I'd be happy to fix the conflicts on this PR if the changes are still wanted. Since this is such a big feature, however, I would not at all be offended if you preferred to tackle this in a different way. Let me know, and thank you for taking time to maintain this library!

kentcdodds:
Hi! Thanks so much for this. Sorry for messing up the git history for you a bit 😬

I hadn't noticed this PR, but I'm definitely a fan. If you'd like to get it updated, I'll help get it merged. Thanks!

malwilley:
@kentcdodds all cleaned up and ready for review!

Note that there are some prettier changes in `index.js` due to a previous PR getting merged without the new formatting requirements.

Also, code coverage is technically down in `index.js`, but I believe that's only because I reduced the number of lines in that file when I extracted the `type` function.

kentcdodds:
Thanks for your work on this. Just want to let you know that I've seen this. I need more time to review it though. I hope to get back to you soon though.

src/type.js
```diff
@@ -0,0 +1,182 @@
...
+  },
+  '{esc}': {
+    key: 'Escape',
+    code: 27,
+    skipPressEvent: true,
+  },
+  '{backspace}': {
+    key: 'Backspace',
+    code: 8,
+    inputType: 'deleteWordBackward',
```
klujanrosas, line 23: Should this be `deleteContentBackward`?

malwilley: Good catch! Yes it should

kentcdodds:
Hi @malwilley,

I promise I'm still thinking about this PR. It's in my inbox and I think about it multiple times a day. I really feel like there's something we could do to improve the design and reduce code complexity, but I think I need to experiment with it myself a little bit. I may work on it a little bit today on my live stream at https://kcd.im/discord if you want to join in later.

malwilley:
@kentcdodds No problem at all! I too think there is a lot of room for improvement -- I'm all ears if there is a more elegant solution to be found.

kentcdodds:
Thanks for all your work on this! I've opened a new PR that I think you'll really like. Would love your feedback: #308

PR #238, opened 2020-03-20
fix(tab): Keep focus on the document if there are no enabled, focusable elements

rtessier:
Hi @Gpx and @kentcdodds. 👋 

I ran into an issue where `tab()` blows up if the case under test has not rendered any elements which are focusable. Probably a bit of a corner case, but it would be nice to have it covered.

Hope the change is okay. This is my first open-source contribution! 😅


benmonro:
@kentcdodds @Gpx can we get this merged?

PR #239, opened 2020-03-21
fix:type no longer clips final character with negative maxlength

### LINKED ISSUE #217: Issue with `.type` when maxlength property is not defined on input
It seems that when the maxlength property is not defined on an input component, JS returns a value of -1 for it, which causes `computedText` to be incorrect (specifically the last character is consistently missing). Wondering if anyone else has observed this behavior?

cbacon91:
Resolves #217.

I noticed this issue when writing tests in a private repo. 

I'm not very familiar with `jest`, so I'm not sure if it's something with that or something else, but I noticed that elements created in the existing specs don't run into the `maxLength` problem mentioned in the linked issue, so I handled it a bit heavy-handedly in the spec.

I do know that creating a bland element with no attrs from a Chrome console can show the max length "bug".

```javascript
element = document.createElement('input')
element.maxLength
// -1
element = document.createElement('textarea')
element.maxLength
// -1
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -183,10 +183,7 @@ const userEvent = {
     };
     const opts = Object.assign(defaultOpts, userOpts);
 
-    const maxLength =
-      element.maxLength && element.maxLength >= 0
-        ? element.maxLength
-        : text.length;
+    const maxLength = element.maxLength > 0 ? element.maxLength : text.length;
     const computedText = text.slice(0, maxLength);
 
     const previousText = element.value;
--- a/src/index.js
+++ b/src/index.js
@@ -183,7 +183,11 @@ const userEvent = {
     };
     const opts = Object.assign(defaultOpts, userOpts);
 
-    const computedText = text.slice(0, element.maxLength || text.length);
+    const maxLength =
+      element.maxLength && element.maxLength >= 0
+        ? element.maxLength
+        : text.length;
+    const computedText = text.slice(0, maxLength);
 
     const previousText = element.value;
 
--- a/__tests__/react/type.js
+++ b/__tests__/react/type.js
@@ -31,6 +31,24 @@ describe("userEvent.type", () => {
     expect(getByTestId("input")).toHaveProperty("value", "hello world");
   });
 
+  it.each(["input", "textarea"])(
+    "includes entire textstring with negative maxlength in <%s>",
+    type => {
+      const onChange = jest.fn();
+      const { getByTestId } = render(
+        React.createElement(type, {
+          "data-testid": "input",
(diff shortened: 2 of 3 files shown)


cbacon91:
@Gpx What do you think of this? it is one solution to #217

src/index.js
```diff
@@ -183,7 +183,11 @@ const userEvent = {
     };
     const opts = Object.assign(defaultOpts, userOpts);
 
-    const computedText = text.slice(0, element.maxLength || text.length);
+    const maxLength =
+      element.maxLength && element.maxLength >= 0
```
apalaniuk: `element.maxLength && element.maxLength >= 0` seems slightly misleading because `>= 0` makes it look like `element.maxLength` will be applied if it's `0`, which isn't true. Particularly relevant in that jsdom seems to use 0 as its default/unlimited value, which deviates from specs/browser implementations.

apalaniuk: `element.maxLength && element.maxLength >= 0` seems slightly misleading because `>= 0` makes it look like `element.maxLength` will be applied if it's `0`, which isn't true. Particularly relevant in that jsdom seems to use 0 as its default/unlimited maxlength value, which deviates from specs/browser implementations.

cbacon91: That's a really good point that I overlooked. I've updated this.

apalaniuk: For context, I opened https://github.com/jsdom/jsdom/issues/2927 since that's the real source of deviation

src/index.js
```diff
@@ -183,7 +183,8 @@ const userEvent = {
     };
     const opts = Object.assign(defaultOpts, userOpts);
 
-    const computedText = text.slice(0, element.maxLength || text.length);
+    const maxLength = element.maxLength > 0 ? element.maxLength : text.length;
```
patmaddox, line 186: From the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength):

> The maxlength attribute defines the maximum number of characters (as UTF-16 code units) the user can enter into an <input> or <textarea>. This must be an integer value 0 or higher. If no maxlength is specified, or an invalid value is specified, the input or textarea has no maximum length.

So it seems like we wouldn't want to check for `maxLength > 0`, because `0` is a valid value for maxLength. It needs to check for `>= 0` instead, which necessitates another test for `maxLength: 0`

@a-bates

cbacon91:
#263 resolves this; closing as dupe

kentcdodds:
Thanks for the PR @cbacon91! Sorry I never noticed it before. I'm only now starting to pick up maintenance of this package a bit.

PR #240, opened 2020-03-25
ignore maxLength -1 for userEvent.type

ThomasRoest:
### the problem
We use Karma as a test runner and run tests in the browser. When using userEvent.type, the last character of the value is sliced of. The default value for maxLength for inputs in browsers is -1. So this is true in all our tests. So for example `mytestvalue` becomes `mytestvalu`. A possible fix is ignoring maxLength -1. 

Wat do you think?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -185,7 +185,7 @@ const userEvent = {
 
     let computedText;
     if (!element.maxLength || element.maxLength === -1) {
-      computedText = text.slice(0, text.length);
+      computedText = text.slice(0, element.maxLength > 0 ? element.maxLength : text.length);
     } else {
       computedText = text.slice(0, element.maxLength);
     }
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -183,7 +183,12 @@ const userEvent = {
     };
     const opts = Object.assign(defaultOpts, userOpts);
 
-    const computedText = text.slice(0, element.maxLength || text.length);
+    let computedText;
+    if (!element.maxLength || element.maxLength === -1) {
+      computedText = text.slice(0, text.length);
```
apalaniuk: This could be simplified to

```suggestion
      computedText = text.slice(0, element.maxLength > 0 ? element.maxLength : text.length);
```

and avoid the conditional assignment. It's also slightly more spec conformant since any invalid value should result in no max length being applied (realizing that jsdom diverges by using `0` as its default/unlimited value)

cbacon91:
looks like #263 was merged which resolves this.

kentcdodds:
Sorry I missed this PR! Thanks for the contribution anyway.

PR #244, opened 2020-04-18
#232 type should support replacing the current input

hamidvirani:
This PR supports an extra option replaceExisting in the type function, which when set to true will replace the content of the input instead of appending.

The replaceExisting parameter will be optional and will be default to false.

### DIFF
--- a/__tests__/react/type.js
+++ b/__tests__/react/type.js
@@ -42,6 +42,34 @@ describe("userEvent.type", () => {
     expect(getByTestId("input")).toHaveProperty("value", "hello world");
   });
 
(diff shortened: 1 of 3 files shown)


kentcdodds:
Thanks for this, but I don't think this is the approach we should take. The user will first select the input's contents and then either delete it with the delete or backspace key or will type something else instead.

We may have support for this soon. If not, feel free to open an issue about it.

hamidvirani:
Thanks @kentcdodds for the update. At the time of opening the PR, there was no support for `clear`. Now that we have `clear` method, I agree we don't need this.

PR #250, opened 2020-04-24
fix: Add TypeScript userEvent.clear definition

bdh1011:
I found there is no userEvent.clear type definition in version 10.1.0 (https://github.com/testing-library/user-event/pull/233)
I just added userEvcent.clear

PR #251, opened 2020-04-24
feat(clear): Add clear() to the typescript definitions

pleunv:
Quick PR to add the new clear() method to the type defs.


pleunv:
Whoopsie, [someone was ahead of me](https://github.com/testing-library/user-event/pull/250)

PR #253, opened 2020-04-27
Check for null activeElements when acting

rbusquet:
Following https://html.spec.whatwg.org/multipage/interaction.html#dom-documentorshadowroot-activeelement and most recent JSDom versions, `document.activeElement` can return null. `fireEvent.click` will fail when the argument is null with `Error: Unable to fire a "mouseMove" event - please provide a DOM element.`.

----

I'm having a hard time identifying and reproducing the issue. It was related to focused elements unmounting asynchronously during test execution, causing the last focused element to have a nullable `root.` I'll go back and try to write tests. I can do it if I explicitly call `focus` on a reference to a detached component.

### DIFF
--- a/__tests__/react/click.js
+++ b/__tests__/react/click.js
@@ -434,14 +434,14 @@ describe("userEvent.click", () => {
         </button>
       ) : (
         <button key="btn2" data-testid="btn2">
-          Second buttom
+          Second button
         </button>
       );
     }
 
     const tree = render(<App />);
     const btn1 = tree.getByTestId("btn1");
-    fireEvent.click(btn1);
+    userEvent.click(btn1);
     btn1.focus();
     expect(() => {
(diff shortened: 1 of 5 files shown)


kentcdodds:
I figured out what's going on here (ran into it myself). Sorry, I only just started fixing up this repo. I'll be making several changes and will handle this as part of these cleanup changes.

rbusquet:
@kentcdodds thanks! I still didn't figure out how to reproduce it but it's not happening a lot. I'll be following the changes and wait for the fix. Let me know if there's anything more I can do to help!

kentcdodds:
This should be fixed in the latest version.

Sorry that I couldn't find a good way to get your changes in. But I'd like to thank you formally anyway :)

@all-contributors please add @rbusquet for bugs, code, and tests

Issue #256, opened 2020-05-04
[Typescript]: Missing "clear" type definition


Meemaw:
"clear" function type definition is not exported.

Meemaw:
Fixed in https://github.com/testing-library/user-event/pull/250

PR #257, opened 2020-05-04
[Typescript]: Export "clear" function def

Meemaw:
Fixes https://github.com/testing-library/user-event/issues/256


Meemaw:
I see this is a duplicate of https://github.com/testing-library/user-event/pull/250

PR #262, opened 2020-05-08
fix: clear function on non text inputs

Meemaw:
`element.setSelectionRange(0, element.value.length)` does not work on all input types (e.g. `type="number"` or `type"email"`). See https://stackoverflow.com/questions/33406169/failed-to-execute-setselectionrange-on-htmlinputelement-the-input-elements

Using `userEvent.clear` on such element will throw an error:
Screenshot 2020-05-08 at 15 16 08


src/index.js
```diff
@@ -106,29 +106,40 @@ function fireChangeEvent(event) {
...
+      inputType: "deleteContentBackward",
     });
     element.value = ""; // when we add special keys to API, will need to respect selected range
   }
 }
 
 function selectAll(element) {
   userEvent.dblClick(element); // simulate events (will not actually select)
+  const elementType = element.type;
+  if (isInputElement(element)) {
```
Meemaw, line 132: We have to check this because type is a read only property on textarea.

package.json
```diff
@@ -33,25 +33,25 @@
...
+    "lint-staged": "^10.2.2",
+    "prettier": "^2.0.5",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
     "regenerator-runtime": "^0.13.5",
-    "semantic-release": "^17.0.4",
+    "semantic-release": "^17.0.7",
     "travis-deploy-once": "5.0.11",
     "vue": "^2.6.11",
     "vue-template-compiler": "^2.6.11"
```
kentcdodds, line 57: Was it necessary to update all dependencies for this PR?

Meemaw: No need, but I like to keep dependencies up to date. I can revert rhat if you'll like that.

Meemaw: ping @kentcdodds

Meemaw:
@kentcdodds I've removed the `package.json` changes

src/index.js
kentcdodds: Could we add a comment here indicating why changing the type to text is necessary?

Meemaw: Done 👍

PR #263, opened 2020-05-08
fix: type when element.maxLength is -1

pomber:
I'm using user-event in the browser, not sure if it's officially supported but it mostly works. 

There's a problem with `type`, it's removing the last char from the text because the browser returns -1 from `element.maxLength`. I changed it to only slice with maxLength when it's a positive number.


kentcdodds:
@all-contributors please add @pomber for code

PR #269, opened 2020-05-12
docs: fix README formatting

skovy:
**Problem:** had initially missed this method in the documentation when it was added because it wasn't formatted the same as the others.

**Solution:** format `clear` consistent with other methods.


kentcdodds:
@all-contributors please add @skovy for docs

PR #271, opened 2020-05-12
src: Add support for ctrl click.

Raynos:
We pass the `init` argument through so that the underlying
click event can be created with ctrlKey set to true if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,15 @@ test("click", () => {
 });
 ```
 
+You can also ctrlClick / shiftClick etc with
+
+```js
+userEvent.click(elem, { ctrlKey: true, shiftKey: true })
+```
+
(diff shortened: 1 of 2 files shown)


Raynos:
I found this library from a recent blog post by @kentcdodds ( https://kentcdodds.com/blog/common-mistakes-with-react-testing-library/#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find ).

This is a great library; I've added a small feature to allow testing multi-selection behavior in my app with ctrl and shift click behavior.

kentcdodds:
@all-contributors please add @Raynos for code

Raynos:
@kentcdodds @Gpx 

You guys are amazing and super responsive. This is fantastic code maintenance.

Raynos:
Forgot to add the type definition in the `types` file ( https://github.com/testing-library/user-event/pull/271 ).

PR #273, opened 2020-05-13
Reduce size of npm package by 99%

Raynos:
There exists a 12MB binary in the tarball published to npm called `cc-test-reporter`.

This looks like a happy accident to me. It's only neccessary on the CI machine and should be removed before publishing the library.

The npm tarball for `user-event` is 12.9MB ; of which 12.8MB is this `cc-test-reporter` binary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ script:
 after_success:
   - npm run codecov
   - npm run build
+  - rm ./cc-test-reporter
   - npm run travis-deploy-once "npm run semantic-release"
 branches:
   except:


Raynos:
Travis failed because of 429 rate limiting on github when trying to git clone the source code.

kentcdodds:
Thanks for this! I think this repo is due for some tooling improvements which I plan to make soon, but I'll get this merged now to fix this issue right now.

Raynos:
Npm says the tarball is now 112kb and not 12.9mb

Thanks !

kentcdodds:
Hooray! 🎉

PR #279, opened 2020-05-15
feat(upload): Add upload API

### LINKED ISSUE #267: File uploading API
I'm not sure that this is in the scope but would be great to have an API for uploading file like this:
```ts
function upload(element: TargetElement, file: File | File[]): void

await userEvent.upload(element, file)
await userEvent.upload(element, files)
```

At now I'm doing this:

```ts
fireEvent.change(input as HTMLInputElement, {
  target: { files: [file] },
})
```
What do you think?

vadimshvetsov:
Closes #267

I tried to implement it as close as possible to real user interaction.

There is [no `cypress` implementation](https://github.com/testing-library/user-event/pull/279) for uploading files at this moment to check.

At first I thought about emulating `input.files` with `DataTransfer` to get `FileList` but Node needs a polyfill. I haven't found reliable implementation of `DataTransfer`. I mocked `FileList` API on my own therefore.

Looking forward for reviews and suggestions. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,8 +181,7 @@ test("upload multiple files", () => {
 
 ### `clear(element)`
 
-Selects the text inside an `<input>` or `<textarea>` and deletes it. Also
-removes files from `<input type="file">`
+Selects the text inside an `<input>` or `<textarea>` and deletes it.
 
 ```jsx
 import React from "react";
--- a/README.md
+++ b/README.md
@@ -141,10 +141,11 @@ one character at the time. `false` is the default value.
 are typed. By default it's 0. You can use this option if your component has a
 different behavior for fast or slow users.
 
-### `upload(element, file)`
+### `upload(element, file, [{ clickInit, changeInit }])`
 
 Uploads file to an `<input>`. For uploading multiple files use `<input>` with
-`multiple` attribute and the second `upload` argument must be array then.
+`multiple` attribute and the second `upload` argument must be array then. Also
+it's possible to initialize click or change event with using third argument.
 
 ```jsx
 import React from "react";
--- a/README.md
+++ b/README.md
@@ -90,10 +90,11 @@ test("click", () => {
 You can also ctrlClick / shiftClick etc with
 
 ```js
-userEvent.click(elem, { ctrlKey: true, shiftKey: true })
+userEvent.click(elem, { ctrlKey: true, shiftKey: true });
 ```
 
-See the [`MouseEvent`](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/MouseEvent)
+See the
+[`MouseEvent`](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/MouseEvent)
 constructor documentation for more options.
 
 ### `dblClick(element)`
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -140,9 +141,48 @@ one character at the time. `false` is the default value.
 are typed. By default it's 0. You can use this option if your component has a
 different behavior for fast or slow users.
 
+### `upload(element, file)`
```
vadimshvetsov, line 144: The second argument might be single file or array of files. I'm not sure about that, may be second arg must be array only? But personally I think it's inconvenient to always remember about wrapping file to an array.

README.md
```diff
@@ -140,9 +141,48 @@ one character at the time. `false` is the default value.
...
+  expect(input.files[0]).toStrictEqual(files[0]);
+  expect(input.files[1]).toStrictEqual(files[1]);
+});
+```
+
 ### `clear(element)`
 
-Selects the text inside an `<input>` or `<textarea>` and deletes it.
+Selects the text inside an `<input>` or `<textarea>` and deletes it. Also
+removes files from `<input type="file">`
```
vadimshvetsov, line 185: Also I'm not sure about adding this kind of functionality to `clear` function. I've got tired when implementing mocking reseting file input element properly in my tests and maybe nice to have such `clear` capability in that cases.

src/index.js
```diff
@@ -294,6 +301,34 @@ const userEvent = {
     element.addEventListener("blur", fireChangeEvent);
   },
 
+  upload(element, fileOrFiles, init) {
+    if (element.disabled) return;
+    const focusedElement = element.ownerDocument.activeElement;
+
+    let files;
```
vadimshvetsov, line 308: This `files` variable depends on input `multiple` attribute. If it's true `fileOrFiles` arg assuming as array otherwise as singlefile. At first it relied on `Array.isArray` check but then I thought about passing test cases with multiple files and input without attribute `multiple`.

vadimshvetsov:
@kentcdodds The only way user can use this kind of `clear` behaviour with file input is by pressing reset button on a form. It really looks like this is out of the `clear` scope and should be implemented in user land.

src/index.js
```diff
@@ -294,6 +294,34 @@ const userEvent = {
...
+      files = element.multiple ? fileOrFiles : [fileOrFiles];
+      clickElement(element, focusedElement, init);
+    }
+
+    fireEvent.change(element, {
+      target: {
+        files: {
+          length: files.length,
+          item: (index) => files[index],
+          ...{ ...files },
```
kentcdodds, line 319:
```suggestion
          ...files,
```

src/index.js
```diff
@@ -294,6 +294,34 @@ const userEvent = {
     element.addEventListener("blur", fireChangeEvent);
   },
 
+  upload(element, fileOrFiles, init) {
```
kentcdodds, line 297: Would it not make sense to allow configurability of both events?

```suggestion
  upload(element, fileOrFiles, {clickInit, changeInit}) {
```

vadimshvetsov: Yep, looks reasonable but here I'm only following `clickElement` function API. It would require breaking changes to all functions which use `clickElement` then. Is it desirable?

src/index.js
vadimshvetsov: Was confused with array destructing to object and failing tests 😆

src/index.js
kentcdodds: I think the `clickElement` function is fine as is. No need to change it. It just seems odd to me that we'd allow you to initialize the `click` event and not the `change` event (which is actually what people will be thinking about when using this method).

vadimshvetsov: Now I've catched what you mean, thanks for guiding 😄

typings/index.d.ts
```diff
@@ -1,27 +1,39 @@
...
+  focusTrap?: Document | Element;
 }
 
 export type TargetElement = Element | Window;
 
+export type FilesArgument = File | File[];
+
+export type UploadInitArgument = {
+  clickInit?: MouseEventInit;
+  changeInit?: Event;
```
vadimshvetsov, line 18: I'm not sure in this type, took it from [there](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/testing-library__dom/events.d.ts#L86)

kentcdodds:
@all-contributors please add @vadimshvetsov for ideas, code, tests

PR #283, opened 2020-05-17
Fix maxLength handling over multiple type() calls

### LINKED ISSUE #281: userEvent.type does not respect maxlength when appending to existing text
When https://github.com/testing-library/user-event/commit/1e8d140b0e67aa9da089c832deb997c2a15d5a33 introduced a change to the way `.type()` behaves, it did not update `maxLength` handling to take previous text into account.

As a result, if you have an element like this:

```jsx
<textarea aria-label="input" rows="3" maxlength="10" />
```

doing the following
```js
const inputField = screen.getByLabelText('input');
await userEvent.type(inputField, '123456');
await userEvent.type(inputField, '123456789101112');
screen.debug();
```

ends up producing:

```jsx
<textarea aria-label="input" maxlength="10" rows="3" >
    1234561234567891
</textarea>
```

rather than what you would expect from the browser, namely:
```jsx
<textarea aria-label="input" maxlength="10" rows="3" >
    1234561234
</textarea>
```

9still:
Fixes #281 

When `type()` behavior was changed to append text, rather than replace it, `maxLength` handling was not updated to take any pre-existing text into account.  As a result, when using multiple `.type()` calls, the user is able to enter text in excess of `maxLength`.   More details in #281

### DIFF
--- a/__tests__/react/type.js
+++ b/__tests__/react/type.js
@@ -159,7 +159,7 @@ describe("userEvent.type", () => {
   );
 
   it.each(["input", "textarea"])(
-    "should type text in <%s> up to maxLength if provided",
+    "should enter text in <%s> up to maxLength if provided",
     (type) => {
       const onChange = jest.fn();
       const onKeyDown = jest.fn();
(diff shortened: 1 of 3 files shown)


kentcdodds:
@all-contributors please add @9still for code, tests, and bugs

Issue #268, opened 2020-05-11
`userEvent.clear()` does not fire `onChange` handlers


justinanastos:
Reproduction: https://github.com/justinanastos/user-event-clear-bug

I'm finding that `onChange` handlers are not being fired when using `userEvent.clear(getBy...)`.

This is both a bug report and a request for help. I'm not sure what I'm doing wrong here:

```js
import '@testing-library/jest-dom/extend-expect';
import { render, screen } from '@testing-library/react';
import React from 'react';
import userEvent from '@testing-library/user-event';

it('calling userEvent.clear() should call onChange on an input', () => {
  const onChange = jest.fn();

  render(
    <label>
      example
      <input defaultValue="default value" onChange={onChange} />
    </label>,
  );

  expect(screen.getByLabelText('example')).toHaveValue('default value');
  userEvent.clear(screen.getByLabelText('example'));
  expect(screen.getByLabelText('example')).toHaveValue('');

  // This test fails
  expect(onChange).toHaveBeenCalled();
});
```

rfoel:
I'm getting `InvalidStateError: The object is in an invalid state.` when I call the clear method 🤔

kentcdodds:
Looks like `clear` calls into `backspace` (after selecting all the contents) and this fires an `input` event:

https://github.com/testing-library/user-event/blob/ee8aa5c2a7457132de6ab06a13df7e37697a99be/src/index.js#L128-L130

I didn't write this or review these PRs so I'm not sure why it's working this way (I expect there's some reason for this). In general, I think our goal is to match what cypress does to do these kinds of things. If someone wants to investigate and make sure we're following what cypress does. This may be as simple as *also* firing a `change` event as well as the `input` event. But let's do some investigation.

justinanastos:
Thanks @kentcdodds . First, I verified performing `.type('{selectall}{backspace}')` both clears the `input` and fires the `onChange` handler in this demo repo: https://github.com/justinanastos/user-event-clear-bug-cypress

The `clear` command adds an event handler for `blur` in which it will execute the on change handler; I'm not seeing this fired even when adding a blur event to the test, which probably shouldn't be necessary if this is attempting to mirror the behavior of Cypress.

https://github.com/testing-library/user-event/blob/ee8aa5c2a7457132de6ab06a13df7e37697a99be/src/index.js#L215

I'm happy to open a PR to immediately fire an `onChange` event if that's a path you'd be willing to accept.

```diff
[...]

PR #286, opened 2020-05-19
Make `clear` method fire `onChange` handlers

justinanastos:
Resolves issue #268

Per the issue comments, I believed that the change would only require calling `fireEvent.change` in the `clear` method; but that turned out not to be the case.

React deduplicates events that it fires so that it doesn't emit multiple `onChange` events with the same value. Using `element.value = ""` causes React to expect a change event to have been fired, so calling `fireEvent.change` has will be considered a duplicate event and no event will be fired. I added tests to verify the expected behavior (they fail before the second commit, as one would expect).

The solution I landed on, and I'm super open to feedback, was to follow the logic I found in [stackoverflow](https://stackoverflow.com/questions/23892547/what-is-the-best-way-to-trigger-onchange-event-in-react-js). I abstracted a `setValue` method that will call the native value setter, fire the change event, and then call the prototype override value setter if one exists. I tried a few other solutions that had subtle issues.

The easier solution was to call the `onChange` event with the `target: { value: '' }` and then call `.value = ""`. This called the `onChange` handler, the `onChange` handler's `event.target.value` was correct, and after `clear` finishes, the `.value` would also be correct. However, if you check the original element's `value` inside of the `onChange` handler, it will be wrong because it'll be using React's prototype override; that override has an internal cache of the input `value`, which will be wrong. This is super subtle, but can be damning. I added a check to the tests `onChange` methods to make sure that `event.target.value` and `input.value` are correct when `onChange` is called.

Lastly, I added checks to the existing tests to make sure `onChange` is called when expected.

I expected this to be a much simpler fix; but this is the best I could come up with. I'd love to get some feedback!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -93,19 +93,38 @@ const Keys = {
 }
 
 function backspace(element) {
-  const eventOptions = {
+  const keyboardEventOptions = {
     key: Keys.Backspace.key,
     keyCode: Keys.Backspace.keyCode,
     which: Keys.Backspace.keyCode,
   }
-  fireEvent.keyDown(element, eventOptions)
-  fireEvent.keyUp(element, eventOptions)
+  fireEvent.keyDown(element, keyboardEventOptions)
+  fireEvent.keyUp(element, keyboardEventOptions)
 
   if (!element.readOnly) {
     fireEvent.input(element, {
       inputType: 'deleteContentBackward',
     })
-    element.value = '' // when we add special keys to API, will need to respect selected range
+
+    // We need to call `fireEvent.change` _before_ we change `element.value`
+    // beacuse `fireEvent.change` will use the element's native value setter
+    // (meaning it will avoid prototype overrides implemented by React). If we
+    // call `input.value = ""` first, React will swallow the change event (this
+    // is checked in the tests). `fireEvent.change` will only call the native
+    // value setter method if the event options include `{ target: { value }}`
+    // (https://github.com/testing-library/dom-testing-library/blob/8846eaf20972f8e41ed11f278948ac38a692c3f1/src/events.js#L29-L32).
+    //
+    // Also, we still must call `element.value = ""` after calling
+    // `fireEvent.change` because `fireEvent.change` will _only_ call the native
+    // `value` setter and not the prototype override defined by React, causing
+    // React's internal represetation of this state to get out of sync with the
+    // value set on `input.value`; calling `element.value` after will also call
+    // React's setter, keeping everything in sync.
+    //
+    // Comment either of these out or re-order them and see what parts of the
+    // tests fail for more context.
+    fireEvent.change(element, {target: {value: ''}})
+    element.value = ''
   }
 }
 
--- a/src/index.js
+++ b/src/index.js
@@ -107,7 +107,7 @@ function backspace(element) {
     })
 
     // We need to call `fireEvent.change` _before_ we change `element.value`
-    // beacuse `fireEvent.change` will use the element's native value setter
+    // because `fireEvent.change` will use the element's native value setter
     // (meaning it will avoid prototype overrides implemented by React). If we
     // call `input.value = ""` first, React will swallow the change event (this
     // is checked in the tests). `fireEvent.change` will only call the native
--- a/src/__tests__/clear.js
+++ b/src/__tests__/clear.js
@@ -3,7 +3,11 @@ import {render, screen} from '@testing-library/react'
 import userEvent from '../../src'
 
 test.each(['input', 'textarea'])('should clear text in <%s>', type => {
-  const onChange = jest.fn()
+  const onChange = jest.fn().mockImplementation(event => {
+    // Verify that `event.target`'s value is correct when the event handler is
+    // fired.
+    expect(event.target).toHaveProperty('value', '')
+  })
   render(
     React.createElement(type, {
       'data-testid': 'input',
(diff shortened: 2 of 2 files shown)


src/index.js
```diff
@@ -1,7 +1,7 @@
 import { fireEvent } from "@testing-library/dom";
 
 function wait(time) {
-  return new Promise(function (resolve) {
+  return new Promise(function(resolve) {
```
evocateur, line 4: With all of these irrelevant formatting changes, it's really hard to grok the real diff.

Is this caused by the `pre-commit` hook, or is this a wonky editor plugin?

justinanastos: Oh how embarrassing. Let me go through and figure out why I have whitespace changes.

justinanastos: Circling back this is totally the pre-commit hook. Let me separate out the actual changes I proposed from the prettier changes.

justinanastos:
@evocateur I separated out a commit that has the prettier updates applied by the pre-commit hook to the files I was about to change. Hopefully this makes reviewing commit-by-commit useful now. Sorry for not doing this to begin with.

src/index.js
```diff
@@ -115,20 +115,54 @@ const Keys = {
...
+  fireEvent.keyDown(element, keyboardEventOptions);
+  fireEvent.keyUp(element, keyboardEventOptions);
 
   if (!element.readOnly) {
     fireEvent.input(element, {
       inputType: "deleteContentBackward"
     });
-    element.value = ""; // when we add special keys to API, will need to respect selected range
+
+    setElementValue(element, "");
```
kentcdodds, line 165: Hmmm... Is there a reason we can't just use `fireEvent.change(element, {target: {value}})`? The `setElementValue` function is basically what `fireEvent.change` does: https://github.com/testing-library/dom-testing-library/blob/8846eaf20972f8e41ed11f278948ac38a692c3f1/src/events.js#L81-L96

justinanastos: Great question. I tried to explain this in the PR description but I obviously struggled and didn't get the point across. Also, I hadn't yet seen the `setNativeValue` implementation. Thanks for the patience of letting me try again.

My first attempt tried to use `fireEvent.change(element, { target: { value: '' }})` without realizing that they have a very similar function to the function that you linked to (https://github.com/testing-library/dom-testing-library/blob/8846eaf20972f8e41ed11f278948ac38a692c3f1/src/events.js#L81-L96) above https://github.com/testing-library/user-event/pull/286#discussion_r427617114.

If I replace this `setElementValue( ... )` with the `fireEvent.change`, the problem is that React's internal representation of state will fall out of sync with the actual `input.value` because React internally caches the `input` value https://github.com/facebook/react/blob/8b3b5c35245b3aa0b2802b51856fe9dad5bcb671/packages/react-dom/src/client/inputValueTracking.js#L51-L104. When we set the native `.value`, the React `value` setter isn't called, causing React to have a different internal value than the `input` itself. Fortunately the test additions I added will catch this. Changing the value will show that the value is not maintained.

This is the part of my abstraction that differs from the above linked: https://github.com/testing-library/user-event/pull/286/files#diff-1fdf421c05c1140f6d71444ea2b27638R144-R148, which calls the prototype overridden value setter after the event is dispatched.

However, there's a way simpler solution that I didn't see and is now obvious after looking at the `fireEvent.change` and these new tests. If I just call `input.value = ""` after calling `fireEvent.change` then everything works as we'd expect! I'll make this change.

vadimshvetsov:
It looks like your `prettier` config overrides repo one somehow. I've made changes in some PRs literally yesterday and I haven't got this format diffs. If you can't handle this pre-commit hook, you could try to use `git commit --no-verify' flag

justinanastos:
> It looks like your `prettier` config overrides repo one somehow. I've made changes in some PRs literally yesterday and I haven't got this format diffs. If you can't handle this pre-commit hook, you could try to use `git commit --no-verify' flag

I wish I had a good explanation of what I was doing wrong @vadimshvetsov. I checked out master and ran prettier after an `npm ci`. I have no idea why I was seeing discrepancies before; they are gone now. Thanks for your patience.

__tests__/react/clear.js
```diff
@@ -72,14 +78,18 @@ describe("userEvent.clear", () => {
     it.each(["input", "textarea"])(
       `should clear when <%s> is of type="${type}"`,
       (inputType) => {
+        const onChange = jest.fn().mockImplementation((event) => {
+          expect(input).toHaveProperty("value", "");
+          expect(event.target).toHaveProperty("value", "");
```
vadimshvetsov, line 83: I'm wondering is there a specific case why assertions placed inside a mocked implementation instead of test flow? Isn't `input` strict equal `event.target` here and already checked in the end of the test (on line 100)?

kentcdodds:
I super duper apologize here, but I just made a bunch of changes to the project to clean things up a bit and now there are some significant merge conflicts. Rather than fix the merge conflicts, it may be easier to create the PR anew.

I hope you accept my apology. I know that's totally not fun. Let me know if you'd rather I just do this myself 😬 Sorry!

__tests__/react/clear.js
justinanastos: Yes, `input` is strict equal to `event.target`.

I ran into an edge case with the implementation that I've since scrapped. The sequence was events was as follows:

* call `input`'s prototype `value` setter
* fired change event (wasn't using `fireEvent.change`)
* use `input.value` setter, which is overridden by React

In this case, when the event handler was fired, `input.value` would still be the previous value because `input.value`'s getter is also overridden by React and uses an internal copy of the state. If we didn't use `input.value = `to assign the value, then the internal state would not be updated. This is super subtle and I'm fine getting rid of it, but this does add a little bit of confidence. I'll add a comment in the meantime and remove if it that's what you want.

justinanastos:
> I super duper apologize here, but I just made a bunch of changes to the project to clean things up a bit and now there are some significant merge conflicts. Rather than fix the merge conflicts, it may be easier to create the PR anew.
> 
> I hope you accept my apology. I know that's totally not fun. Let me know if you'd rather I just do this myself 😬 Sorry!

Thanks for the offer @kentcdodds ; merge conflicts resolved, rebased, and pushed.

PR #288, opened 2020-05-20
fix(upload): FileList item function must return null instead undefined when it can't find a file

vadimshvetsov:
I was looking at `FileList` implementation and noted that `item` function must return `File` or `null` instead `undefined`. Here is [`jsdom` implementation](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/file-api/FileList-impl.js#L10) of it. Also fix wrong expectation in test that's not cause an error when it must. Also I removed unnecessary expectations in `upload` tests.

### DIFF
--- a/__tests__/react/upload.js
+++ b/__tests__/react/upload.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { cleanup, render, fireEvent } from "@testing-library/react";
+import { cleanup, render } from "@testing-library/react";
 import "@testing-library/jest-dom/extend-expect";
 import userEvent from "../../src";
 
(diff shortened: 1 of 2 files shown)


__tests__/react/upload.js
```diff
@@ -86,14 +86,6 @@ describe("userEvent.upload", () => {
     expect(input.files[0]).toStrictEqual(file);
     expect(input.files.item(0)).toStrictEqual(file);
     expect(input.files).toHaveLength(1);
-
-    fireEvent.change(input, {
-      target: { files: { item: () => {}, length: 0 } },
-    });
-
-    expect(input.files[0]).toBeUndefined();
```
vadimshvetsov, line 94: This tests do nothing special to `user-event` library. They only checks that `fireEvent` function works properly.

PR #289, opened 2020-05-20
fix: general `null` and `ownerDocument` fixes

kentcdodds:


kentcdodds:
Discovered [this issue](https://github.com/jsdom/jsdom/issues/2978) thanks to these bugs 😅

Issue #295, opened 2020-05-21
[feature request] Ability to change element focus between `type` events


kayleighridd:
Currently all the `keyUp`, 'keyPress` and `keyDown` events within `type` fire on the specified element, but it would be useful to be able to fire these on the currently active element so that focus can change between these events.

My use case is a multi input field where I change the focus as input reaches its max character length so that the value is added to the next input.
I do this by changing the input focus on the `keyDown`  event so that the `keyUp` event fires in another input.
When I am testing, I'd like to specify a `type` event on one input, but confirm that the focus has moved and the value added to the next input.

Here is a quick sandbox to illustrate my use case:
https://codesandbox.io/s/userevents-type-events-firing-on-same-element-not-active-element-229sk
(Code sandbox doesn't seem to respect the jsdom-sixteen test env so the error isn't correct, but hopefully it gives some context)

kentcdodds:
Hi @kayleighridd,

I can visualize the kind of experience you're describing... What if instead of firing all the events on the element that's passed, we first focus that element and then reference the `document.activeElement`? That's actually more like what the user's doing when they type and it should support the situation you're describing. What do you think?

kayleighridd:
@kentcdodds yes, that would be ideal 🎉 
I assumed this would be a feature that we can turn on through `options` so it's not a breaking change, but are you suggesting this should always be the behaviour?

kayleighridd:
Okay great. Yeah, I'm happy to raise a pull request for this, thanks!

kayleighridd:
👋  I've raised a PR for this here: https://github.com/testing-library/user-event/pull/299
I'm not sure how to link it to this issue.

PR #299, opened 2020-05-27
Bugix/fire type events on active element

kayleighridd:
**What**:
I raised an issue (#295) to allow focus to change between type events. This supports scenarios where focus is switched to a second input in between `keyDown` and `keyUp` events, for example.

This PR also contains a lot of formatting fixes that happened automatically on commit.

**Why**:
I personally, needed this to test a multiple input field, but this also more closely matches user behaviour.

**How**:
I've updated the `type` function to:
- Set focus on the `element` provided
- Reference the current `document.activeElement` when firing individual events
- Reference the current `document.activeElement`'s value and `maxLength` when checking `isTextPastThreshold` so that we check the relevant element's threshold before firing `input`
- Moved the `computedText` and `previousText` variables inside the `allAtOnce` check as it's no longer needed elsewhere. 
-`allAtOnce` functionality is unchanged.
(As there are a lot of formatting changes to sift through, the functionality changes described above can be found around this comment: https://github.com/testing-library/user-event/pull/299/files#r431147168)

**Checklist**:
- [ ] Documentation - N/A (I don't think this is required, but happy to add some if needed)
- [x] Tests
- [ ] Typings - N/A
- [x] Ready to be merged

The work in this PR looks like it conflicts with the work / relates to the conversations happening in https://github.com/testing-library/user-event/pull/231
So, this may be blocked until that PR is resolved.

### DIFF
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {Fragment} from 'react'
 import {render, screen} from '@testing-library/react'
 import userEvent from '../../src'
 
@@ -256,3 +256,63 @@ test.each(['input', 'textarea'])(
     expect(onKeyUp).not.toHaveBeenCalled()
   },
 )
+
+test('should fire events on the currently focussed element', async () => {
+  const changeFocusLimit = 7
+  const onKeyDown = jest.fn(event => {
+    if (event.target.value.length === changeFocusLimit) {
+      screen.getByTestId('input2').focus()
+    }
+  })
+
+  render(
+    <Fragment>
+      <input data-testid="input1" onKeyDown={onKeyDown} />
+      <input data-testid="input2" />
+    </Fragment>,
+  )
+
+  const text = 'Hello, world!'
+
+  const input1 = screen.getByTestId('input1')
+  const input2 = screen.getByTestId('input2')
+
+  await userEvent.type(input1, text)
+
+  expect(input1).toHaveValue(text.slice(0, changeFocusLimit))
+  expect(input2).toHaveValue(text.slice(changeFocusLimit))
+  expect(input2).toHaveFocus()
+})
+
+test('should enter text up to maxLength of the current element if provided', async () => {
+  const changeFocusLimit = 7
+  const input2MaxLength = 2
+
+  const onKeyDown = jest.fn(event => {
+    if (event.target.value.length === changeFocusLimit) {
+      screen.getByTestId('input2').focus()
+    }
+  })
+
+  render(
+    <Fragment>
+      <input data-testid="input" onKeyDown={onKeyDown} />
+      <input data-testid="input2" maxLength={input2MaxLength} />
+    </Fragment>,
+  )
+
+  const text = 'Hello, world!'
+  const input2ExpectedValue = text.slice(
+    changeFocusLimit,
+    changeFocusLimit + input2MaxLength,
+  )
+
+  const input1 = screen.getByTestId('input')
+  const input2 = screen.getByTestId('input2')
+
+  await userEvent.type(input1, text)
+
+  expect(input1).toHaveValue(text.slice(0, changeFocusLimit))
+  expect(input2).toHaveValue(input2ExpectedValue)
+  expect(input2).toHaveFocus()
+})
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -290,21 +322,27 @@ function clear(element) {
```
kayleighridd, line 322: Functionality changes below. (As there are a lot of formatting changes, this is just a comment to anchor to)

src/index.js
```diff
@@ -290,21 +322,27 @@ function clear(element) {
 
 async function type(element, text, {allAtOnce = false, delay} = {}) {
   if (element.disabled) return
```
kayleighridd, line 324: I haven't duplicated this disabled check in the loop to check `document.activeElement.disabled`, because I don't think the focus can change to a disabled element.
But, if it feels safer to check I'm happy to update.

kayleighridd:
I can see that my Code Coverage report has failed and is a blocking issue, but I'm not quite sure how to tackle it. I would appreciate any advice / suggestions 😄

src/index.js
```diff
@@ -313,28 +351,28 @@ async function type(element, text, {allAtOnce = false, delay} = {}) {
...
+        const pressEvent = fireEvent.keyPress(currentElement(), {
           key,
           keyCode,
           charCode: keyCode,
         })
 
         const isTextPastThreshold =
-          (actuallyTyped + key).length > (previousText + computedText).length
+          (actuallyTyped() + key).length >
+          (currentElement().maxLength || text.length)
```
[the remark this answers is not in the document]
kayleighridd: I've updated this to the way I think it should work, but I think I may have misunderstood.
The way I understand this is....
Previously we could calculate how many total characters we can type based on the value that was already in the input before we started using `previousText`. We could then use this as a check against what we've `actuallyTyped`.
Now we are using the current value (`actuallyTyped`) of the element (because the `previousValue` may no longer apply to the current element), so we're computing how many more characters we can add at any point (not total).

I had originally assumed the `isTextPastThreshold` check would be:
```const isTextPastThreshold =
          (currentValue() + key).length >
          (currentValue() + computeText()).length```
But this felt wrong, because we're checking how many more characters we can add and then adding it to the current value and that doesn't necessarily mean it's the correct length.

So, I think what I've done is correct, but let me know if I've completely missed the point 😄 
(Also, to note, where I'm saying `actuallyTyped` I've renamed it to `currentValue` in the code because this felt more reflective of the actual usage now)

File: src/index.js (hunk 2 above)
Comment (kayleighridd) [reply src/index.js:336]: No worries, updated now.

File: src/index.js [hunk 4]
```diff
@@ -363,6 +366,8 @@ async function typeImpl(element, text, {allAtOnce = false, delay} = {}) {
       // eslint-disable-next-line no-await-in-loop
       if (delay > 0) await wait(delay)
 
+      if (currentElement.disabled) return
```
kentcdodds, line 369: This should be a function call right?

```suggestion
      if (currentElement().disabled) return
```

src/__tests__/type.js
```diff
@@ -256,3 +256,63 @@ test.each(['input', 'textarea'])(
     expect(onKeyUp).not.toHaveBeenCalled()
   },
 )
+
+test('should fire events on the currently focussed element', async () => {
```
kentcdodds, line 260:
```suggestion
test('should fire events on the currently focused element', async () => {
```

src/index.js
kayleighridd: 🤦‍♀️  😂  Yes, thank you!

kentcdodds:
@all-contributors please add @kayleighridd for bugs, code, and tests

kayleighridd:
🎉 🙌  Thanks!

kentcdodds:

ScottG489:
FYI this broke some of my tests coming from `11.0.0` since I believe this is the only change in `11.0.1`. I know this is an old change, but it was a little frustrating to see a patch version change break tests. I just wanted to call it out for visibility because I've used testing-library for a while and it's a quality project, so I'm sure you'd like to know.

ScottG489:
FYI this broke some of my tests coming from `11.0.0` since I believe this is the only change in `11.0.1`. I know this is an old change, but it was a little frustrating to see a patch version change break tests. I just wanted to call it out for visibility because I've used testing-library for a while and it's a quality project, so I'm sure you'd like to know.

PR #300, opened 2020-05-28
Implement arrow interactions as arrowPress

juanca:
**What**: Add arrow interactions to the user-event API

**Why**: Arrow interactions emit key down event, maybe a key press event (I need help on this), and a key up event -- in that particular order. Arrow interactions are crucial in implementing accessibility support for UI widgets. However, there are more than just arrows -- we also need a HOME and END (to name a couple) keys support.  

**How**: Adds an `arrowPress` function to the module API -- however, perhaps it makes more sense to expose these interactions as a different API (and for better long-term purposes)

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Typings
- [ ] Ready to be merged
      

This is not ready to be merged. As I developed this, I figured I should get more feedback on the API design and testing design. I can follow up on documentation and typing -- though, I will be honest, I don't know how typing are defined in this repository -- and any additional PR checklist/revisions/what-have-yous. 

Thank you for your time!

### DIFF
--- a/src/__tests__/arrowpress.js
+++ b/src/__tests__/arrowpress.js
@@ -1,0 +1,55 @@
+import React from 'react'
+import {render, screen} from '@testing-library/react'
+import userEvent from '../../src'
+
+test('should handle ArrowDown', () => {
+  const onKeyDown = jest.fn()
+  const onKeyUp = jest.fn()
+
+  render(<div data-testid="element" onKeyDown={onKeyDown} onKeyUp={onKeyUp} />)
+
+  const div = screen.getByTestId('element')
+
+  userEvent.arrowPress(div, 'ArrowDown')
+  expect(onKeyDown).toHaveBeenCalledWith(expect.objectContaining({}))
+  expect(onKeyUp).toHaveBeenCalledWith(expect.objectContaining({}))
+})
+
+test('should handle ArrowLeft', () => {
+  const onKeyDown = jest.fn()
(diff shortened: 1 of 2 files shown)


juanca:
Uh, looks like my IDE removed a bunch of things. 🤔 I didn't see those changes in `git add --patch`

nickmccurdy:
This seems similar to #354, what do you think about it? I started writing tests but you can help out with the implementation if you'd like.

juanca:
I'm open to collaborating. Let me know the repo#branch combo and I can start digging later today / tomorrow as I balance some work priorities with downtime.

nickmccurdy:
Sorry I got distracted and my test doesn't really have any useful assertions (just a snapshot that's currently incorrect), but here's what I started with as an example:

```js
test('{right} on a radio button', () => {
  const {element: button, getEventCalls} = setup(
    <>
      <input type="radio" name="example" defaultChecked />
      <input type="radio" name="example" />
      <input type="radio" name="example" />
    </>,
  )

  userEvent.type(button, '{right}')
})
```

After this only the second radio should be checked.

juanca:
I ended up going a different route: using the event snapshot. See https://github.com/testing-library/user-event/pull/371

Feedback on the PR would be appreciated. Let me know if this is aligned with your thoughts on the user-event API -- I may have done one too many assumptions.

PR #306, opened 2020-06-04
feat(type): replace selected text

### LINKED ISSUE #301: userEvent.type does not replace selected input value
- `@testing-library/user-event` version: 10.4.0
- Testing Framework and version: jest 24.9.0
- DOM Environment: jsdom 14.1.0, 16.2.2

Relevant code or config

```javascript
test('click, auto-focus, type-to-replace', async () => {
  render(<InputThatSelectsOwnValueOnFocus initialValue="whatever" />);
  const input = screen.getByLabelText(/input/i);

  input.focus();
  await user.type(input, 'works'); // should replace selected text

  expect(input).toHaveValue('works'); // but it doesn't, it appends
});
```

```
Expected the element to have value:
  works
Received:
  whateverworks
```

What you did:

I'm trying to test behavior of an input that automatically selects its contents on focus (so the user can immediately start typing to replace the value, rather than having to clear it manually).

What happened:

Typing after focusing the input is appending to the value rather than replacing it (while actual behavior is working as intended).

Reproduction repository:

https://codesandbox.io/s/user-event-type-to-replace-issue-9wu97?file=/src/__tests__/App.test.js

> Note: it's unclear if CSB is properly using `jest-environment-jsdom-sixteen`, but the error is the same locally for jsdom v14 and v16.

Problem description:

Calling `input.focus()` on an "auto-select-value-on-focus" input followed by `await userEvent.type(input, 'whatever')` just appends `whatever` to the existing input value rather than replacing it.

Suggested solution:

Not sure. Is this just a slightly different manifestation of #236?

wKovacs64:
Closes #301

**What**: allowing `type` to replace selected text

**Why**: currently, `type` always appends to the end of the current value (ignoring selected text, which is replaced when a real user starts typing in the browser)

**How**: check for `element.selectionEnd > element.selectionEnd` and replace that segment instead of appending

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Typings
- [ ] Ready to be merged
      

@kentcdodds This change (as currently implemented) breaks the ESLint complexity threshold (clocks in at 15 when the limit is 14). Do we need to simplify it somehow, adjust the rule, disable the rule, other?

Thanks.

### DIFF
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -43,8 +43,9 @@ test('should replace selected text one by one', async () => {
       onChange={onChange}
     />,
   )
-  screen.getByTestId('input').selectionStart = 'hello '.length
-  screen.getByTestId('input').selectionEnd = 'hello world'.length
+  screen
+    .getByTestId('input')
+    .setSelectionRange('hello '.length, 'hello world'.length)
   await userEvent.type(screen.getByTestId('input'), 'friend')
   expect(onChange).toHaveBeenCalledTimes('friend'.length)
   expect(screen.getByTestId('input')).toHaveValue('hello friend')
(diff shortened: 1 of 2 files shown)


wKovacs64:
@kayleighridd would you mind reviewing this, too? I don't _think_ I broke your recent work (tests pass), but another set of eyes wouldn't hurt if you have the time. 🙂 Thanks.

kentcdodds:
I'm actively working on this right now (https://kcd.im/discord). I'll see if I can merge your PR, but I'm making some significant changes to how type works.

wKovacs64:
Oh! Okay.

PR #308, opened 2020-06-04
feat(type): support special charaters (like {enter})

### LINKED ISSUE #31: Check type with special characters

### LINKED ISSUE #172: Emulation of 'Backspace' and other non alphanumeric buttons pressing.

### LINKED ISSUE #181: Simulate keyboard shortcuts

### LINKED ISSUE #301: userEvent.type does not replace selected input value

kentcdodds:
**What**: feat(type): support special characters (like `{enter}`)

**Why**:

Closes #306
Closes #301
Closes #235
Closes #181
Closes #172
Closes #31

**How**:

Oh boy... Totally reworked the `type` functionality. Also made some pretty cool utilities for the tests which might be useful elsewhere.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Typings
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ change the state of the checkbox.
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
+
 - [Installation](#installation)
 - [API](#api)
   - [`click(element)`](#clickelement)
@@ -158,8 +159,8 @@ import userEvent from '@testing-library/user-event'
 test('type', async () => {
   render(<textarea />)
 
-  await userEvent.type(screen.getByRole('textbox'), 'Hello, World!')
-  expect(screen.getByRole('textbox')).toHaveAttribute('value', 'Hello, World!')
+  await userEvent.type(screen.getByRole('textbox'), 'Hello,{enter}World!')
+  expect(screen.getByRole('textbox')).toHaveValue('Hello,\nWorld!')
 })
 ```
 
@@ -169,6 +170,32 @@ one character at the time. `false` is the default value.
 `options.delay` is the number of milliseconds that pass between two characters
 are typed. By default it's 0. You can use this option if your component has a
 different behavior for fast or slow users.
+
+#### Special characters
+
+The following special character strings are supported:
+
+| Text string   | Key       | Modifier   | Notes                                                                              |
+| ------------- | --------- | ---------- | ---------------------------------------------------------------------------------- |
+| `{enter}`     | Enter     | N/A        | Will insert a newline character (`<textarea />` only).                             |
+| `{esc}`       | Escape    | N/A        |                                                                                    |
+| `{backspace}` | Backspace | N/A        | Will delete the previous character (or the characters within the `selectedRange`). |
+| `{shift}`     | Shift     | `shiftKey` | Does **not** capitalize following characters.                                      |
+| `{ctrl}`      | Control   | `ctrlKey`  |                                                                                    |
+| `{alt}`       | Alt       | `altKey`   |                                                                                    |
+| `{meta}`      | OS        | `metaKey`  |                                                                                    |
+
+> **A note about modifiers:** Modifier keys (`{shift}`, `{ctrl}`, `{alt}`,
+> `{meta}`) will activate their corresponding event modifiers for the duration
+> of type command or until they are closed (via `{/shift}`, `{/ctrl}`, etc.).
+
+<!-- space out these notes -->
+
+> We take the same
+> [stance as Cypress](https://docs.cypress.io/api/commands/type.html#Modifiers)
+> in that we do not simulate the behavior that happens with modifier key
+> combinations as different operating systems function differently in this
+> regard.
 
 ### `upload(element, file, [{ clickInit, changeInit }])`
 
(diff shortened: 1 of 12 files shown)


src/type.js
```diff
@@ -0,0 +1,312 @@
...
+
+          fireEvent.keyPress(currentElement(), {
+            key,
+            keyCode,
+            charCode: keyCode,
+            ...eventOverrides,
+          })
+        }
+
+        if (currentElement().tagName === 'BUTTON') {
```
malwilley, line 83: Good thinking here! I didn't even think about non-text elements. Now that I am though, would it be appropriate to also handle `<input type="button" />`, `<input type="submit" />`, etc?

src/__tests__/type-modifiers.js
```diff
@@ -0,0 +1,321 @@
...
+
+// This also means that '{shift}a' will fire an input event with the shiftKey,
+// but will not capitalize "a".
+
+test('{esc} triggers typing the escape character', async () => {
+  const {element: input, getEventCalls} = setup('input')
+
+  await userEvent.type(input, '{esc}')
+
+  expect(getEventCalls()).toMatchInlineSnapshot(`
```
malwilley, line 23: This is such a slick method of testing. I ran into this same problem where tests became super hairy and difficult to reason about, but this is awesome :exploding_head:

README.md
```diff
@@ -170,6 +171,32 @@ one character at the time. `false` is the default value.
...
+| `{esc}`       | Escape    | N/A        |                                                                                    |
+| `{backspace}` | Backspace | N/A        | Will delete the previous character (or the characters within the `selectedRange`). |
+| `{shift}`     | Shift     | `shiftKey` | Does **not** capitalize following characters.                                      |
+| `{ctrl}`      | Control   | `ctrlKey`  |                                                                                    |
+| `{alt}`       | Alt       | `altKey`   |                                                                                    |
+| `{meta}`      | OS        | `metaKey`  |                                                                                    |
+
+> **A note about modifiers:** Modifier keys (`{shift}`, `{ctrl}`, `{alt}`,
+> `{meta}`) will activate their corresponding event modifiers for the duration
+> of type command or until they are closed (via `{/shift}`, `{/ctrl}`, etc.).
```
malwilley, line 190: I do like the extra control we get here by adding `{/shift}` (compared to cypress), but worry that maybe it's a bit too verbose for most use cases.

Perhaps you could release modifier keys at the end of the type statement if no closing modifiers are provided? That could be the best of both worlds. I don't think there's any use case for keeping these keys held down at the end of a type statement (since you don't get the event modifiers after it's done).

src/type.js
```diff
@@ -0,0 +1,313 @@
...
+
+        fireEvent.keyDown(currentElement(), {
+          key,
+          keyCode,
+          which: keyCode,
+          ...eventOverrides,
+          ...newEventOverrides,
+        })
+
+        return {eventOverrides: newEventOverrides}
```
malwilley, line 287: I think you want to return both `eventOverrides` AND `newEventOverrides` here. Otherwise multiple held modifiers don't work correctly (such as `userEvent.type('{ctrl}{meta}{enter}')`)

kentcdodds:
@malwilley, love all the feedback. Would you like to make a pull request to implement it all?

malwilley:
Sure thing!

src/type.js
kentcdodds: Actually, we merge the old eventOverrides with the new ones elsewhere so we're good here 👍

kentcdodds:
We're moving user-event to within dom-testing-library and I just added the auto-release feature :) https://github.com/testing-library/dom-testing-library/commit/bebfc74b3d9685ab33def5737d7a86984cfe515a

If you'd like, there's plenty of opportunity to help still: https://github.com/testing-library/dom-testing-library/pull/616

Issue #316, opened 2020-06-05
type causes InvalidStateError with <input type="email">


kelvinlzhang:
Occurs in master of `@testing-library/user-event` version: 11.2.1

Here's an example failing test, not really up to this library's standards but I wanted to get as close to the source as possible.

STR:

1. Add the following block to src/\_\_tests\_\_/type.js
```
const create = () => {
    const utils = render(<input aria-label="input" type="email"/>)
    const input = utils.getByLabelText('input')
    return {
        input,
        ...utils,
    }
}

it('types text in email input', async () => {
    const { input } = create()
    await userEvent.type(input, 'testing@gmail.com')
    expect(input.value).toBe('testing@gmail.com')
})
```
2. Run test
3. `InvalidStateError`

What happened: I wanted to follow up on this comment and reproduce: https://github.com/testing-library/user-event/issues/307#issuecomment-639226751. Sorry @kentcdodds, the `type=` was abstracted away and it was pretty silly of me to assume it was text. Seems like I ran into 2 completely different issues now, but hopefully I'm just missing something obvious.

Test fails on `InvalidStateError`, adding an `onChange` handler with a `console.log(event.target.value)` shows that `type` only enters in the first char before breaking.

kelvinlzhang:
Yeah, something funky going on with validated text input types.

kentcdodds:
Thanks for this @kelvinlzhang! Would you be interested in digging in and figuring out the issue?

kentcdodds:
@tpict you're welcome to try as well 😁

kelvinlzhang:
@kentcdodds see https://stackoverflow.com/questions/21177489/selectionstart-selectionend-on-input-type-number-no-longer-allowed-in-chrome

Seems like a lot of people aren't too happy about this :(

tpict:
Looks like `user-event` handles this elsewhere by simply [switching out the `type` of the input for the function's lifetime](https://github.com/testing-library/user-event/blob/5c11411b2564603c4a28673dbe089516364f26ec/src/index.js#L217-L229). Would that be a reasonable fix here, or is it off the table on account of `type` being asynchronous?

kentcdodds:
That's what I was thinking. Want to give it a try?

Meemaw:
This is the same problem as https://github.com/testing-library/user-event/pull/262 -- setSelectionRange issue. That function should probably be extracted and reused.

nvh95:
[...]

PR #317, opened 2020-06-06
Fix typing into validated input types

tpict:
**What**:

Temporarily sets the target element's type to `text` until the `type` function resolves

**Why**:

The `type` implementation depends on `setSelectionRange` which [raises an error for validated input types](https://github.com/testing-library/user-event/issues/316) like `email`, `number`

**How**:

Take a guess 😝 

**Checklist**:

- [ ] Documentation N/A (?)
- [x] Tests
- [ ] Typings N/A
- [x] Ready to be merged

I did try replacing `type` with a getter property to log a warning along the lines of "hey, you're accessing `type` while `type`ing and the result will be inaccurate", but it turns out that it's referenced by testing machinery so frequently as to be useless. Does this behavior need to be documented anywhere?

### DIFF
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -496,15 +496,20 @@ test('ignored {backspace} in controlled input', async () => {
   `)
 })
 
-test('typing in an input with validated type works', async () => {
+test.each([
+  ['email', String],
+  ['password', String],
+  ['number', Number],
+  ['text', String],
+])(`typing in an input of type="%s" works`, async (inputType, resultType) => {
   const onChange = jest.fn()
   const {
     container: {firstChild: input},
   } = render(
-    <input type="number" onChange={onChange} aria-label="Test input" />,
+    <input type={inputType} onChange={onChange} aria-label="Test input" />,
   )
 
   await userEvent.type(input, '5550690')
   expect(onChange).toHaveBeenCalledTimes(7)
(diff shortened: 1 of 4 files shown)


src/__tests__/type.js
```diff
@@ -495,3 +495,16 @@ test('ignored {backspace} in controlled input', async () => {
     keyup: 4 (52)
   `)
 })
+
+test('typing in an input with validated type works', async () => {
```
Meemaw, line 499: Consider adding a case for multiple input types, like here https://github.com/testing-library/user-event/blob/master/src/__tests__/clear.js#L77

kentcdodds:
I'm afraid this really isn't going to work. Changing the input's type will violate the principle of least surprise (especially with regard to async code and event handlers that may rely on the `type` being set to something specific).

In addition, it's not going to help people who try to set the selection range before typing.

I'm thinking the solution will be to assume that when the `selectionStart` is `null`, it's actually always at the end of the input. I think that's the best we can do.

Would you be able to add a test to make sure this works on an input with a `number`? `await userEvent.type('12{backspace}34')`. If it does not, could you make it work?

kentcdodds:
Hey @tpict, so I was working on `type` fixing several bugs and I fixed this on my own. Thanks for your help anyway!

PR #320, opened 2020-06-08
Support customElements

### LINKED ISSUE #319: Support firing events inside of custom element shadow DOM
- `@testing-library/user-event` version: 11.2.1
- Testing Framework and version:
  - `@open-wc/testing` version 2.5.17
  - `lit-html` version 1.2.1
  - `mocha` version 6.2.3
- DOM Environment: Chrome@83.0.4103.97

Relevant code or config

```javascript
const el = render('custom-element');
const input = el.shadowRoot.querySelector('input');
userEvent.type(input, 'hello');
```

What you did:
I used `userEvent.type` with on an input-element inside a custom element's shadow DOM (in order to test its internal event listeners).

What happened:
Events did not fire on the input.

Reproduction repository:

Problem description:
The `currentElement`-helper returns `element.ownerDocument.activeElement` which return the custom element host instead of the focused element inside its shadow DOM. I'm guessing this is intended ;) but it is causing problems in this case, and it's not how the real DOM behaves when writing in an input-element inside a custom element.
 
Suggested solution:
Modify the `currentElement`-helper to return `activeElement` inside shadow DOM.

PR on its way :)

hontas:
**What**: fixes #319

**Why**: in order to support custom elements

**How**: modify the `currentElement` helper to return nested `activeElement` inside shadow DOM

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] Typings - I'm not familiar with TS - not sure if I need to do anything here
- [x] Ready to be merged
      

There were more changes required in the testing helpers than in the actual library - I'm open to suggestions on how to improve my suggested solution :)

### DIFF
--- a/src/__tests__/helpers/utils.js
+++ b/src/__tests__/helpers/utils.js
@@ -41,7 +41,6 @@ function setup(ui) {
   const {
     container: {firstChild: element},
   } = render(ui)
-
   element.previousTestData = getTestData(element)
 
   const {getEventCalls, clearEventCalls} = addListeners(element)
--- a/src/__tests__/helpers/utils.js
+++ b/src/__tests__/helpers/utils.js
@@ -37,21 +37,15 @@ function addEventListener(el, type, listener, options) {
   el.addEventListener(type, hijackedListener, options)
 }
 
-function setup(ui, {shadowRootSelector} = {}) {
-  let hostElement
-  let {
+function setup(ui) {
+  const {
     container: {firstChild: element},
   } = render(ui)
-
-  if (shadowRootSelector) {
-    hostElement = element
-    element = element.shadowRoot.querySelector(shadowRootSelector)
-  }
 
   element.previousTestData = getTestData(element)
 
   const {getEventCalls, clearEventCalls} = addListeners(element)
-  return {element, hostElement, getEventCalls, clearEventCalls}
+  return {element, getEventCalls, clearEventCalls}
 }
 
 function addListeners(element) {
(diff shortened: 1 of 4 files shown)


src/__tests__/helpers/utils.js
```diff
@@ -37,14 +37,21 @@ function addEventListener(el, type, listener, options) {
...
+    hostElement = element
+    element = element.shadowRoot.querySelector(shadowRootSelector)
+  }
+
   element.previousTestData = getTestData(element)
 
   const {getEventCalls, clearEventCalls} = addListeners(element)
-  return {element, getEventCalls, clearEventCalls}
+  return {element, hostElement, getEventCalls, clearEventCalls}
 }
```
kentcdodds, line 55: Rather than complicating the `setup` function (especially for what will likely be only a very small number of tests that need it), let's just not use the setup function in the test and instead use `addListeners`.

src/type.js
```diff
@@ -17,14 +17,23 @@ async function type(...args) {
...
 }
 
+const getActiveElement = element => {
+  const activeElement = element.activeElement
+  if (activeElement.shadowRoot) {
+    return getActiveElement(activeElement.shadowRoot) || activeElement
+  } else {
+    return activeElement
+  }
+}
```
kentcdodds, line 27:
```suggestion
const getActiveElement = document => {
  const activeElement = document.activeElement
  if (activeElement.shadowRoot) {
    return getActiveElement(activeElement.shadowRoot) || activeElement
  } else {
    return activeElement
  }
}
```

src/__tests__/helpers/utils.js
hontas: yea good idea! fixing...

kentcdodds:
@all-contributors please add @hontas for code and tests

hontas:
Thank you for your super-duper quick feedback and merge! :)

PR #330, opened 2020-06-09
fix(tab): make tab respect radio groups (fix #207)

### LINKED ISSUE #207: tab with radio buttons
just discovered an issue w/ `tab()`.  it does not account for radio buttons properly and needs to more closely resemble the browser in this respect.

nickmccurdy:
**What**: make tab respect radio groups

**Why**: fixes bugs / inaccurate simulation with `tab()` naively focusing all radio buttons

**How**: ignore radio buttons that aren't checked or at the ends within their named groups

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Typings N/A
- [x] Ready to be merged
      

The algorithm I used to prune irrelevant radio buttons is a bit ugly and hacky because of all the edge cases, so I'd appreciate any ideas to improve it. We also might need to carefully document this change as it could cause tests to fail if they're relying on the incorrect focusing behavior (though changing incorrect functionality to fix a bug is technically not a breaking change in semver).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -408,15 +408,39 @@ function tab({shift = false, focusTrap = document} = {}) {
 
       return diff === 0 ? a.idx - b.idx : diff
     })
+    .map(({el}) => el)
 
-  const index = orderedElements.findIndex(
-    ({el}) => el === el.ownerDocument.activeElement,
+  if (shift) orderedElements.reverse()
+
+  // keep only the checked or first element in each radio group
+  const prunedElements = []
+  for (const el of orderedElements) {
+    if (el.type === 'radio' && el.name) {
+      const replacedIndex = prunedElements.findIndex(
+        ({name}) => name === el.name,
+      )
+
+      if (replacedIndex === -1) {
+        prunedElements.push(el)
+      } else if (el.checked) {
+        prunedElements.splice(replacedIndex, 1)
+        prunedElements.push(el)
+      }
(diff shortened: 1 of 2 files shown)


nickmccurdy:
@benmonro I'm hoping this is what you meant the problem was in #207, if not let's update that issue any I may give it another try.

src/index.js
```diff
@@ -408,15 +408,39 @@ function tab({shift = false, focusTrap = document} = {}) {
...
+      } else if (el.checked) {
+        prunedElements.splice(replacedIndex, 1)
+        prunedElements.push(el)
+      }
+    } else {
+      prunedElements.push(el)
+    }
+  }
+
+  if (shift) prunedElements.reverse()
```
benmonro, line 434: Is it already reversed since you reversed the ordered elements above?

nickmccurdy: Yes. I'm reversing it twice (before and after the pruning) when using `shift` because I couldn't figure out a clean way to make the pruning logic reverse directions without duplicating it.

nickmccurdy:
Sorry I force pushed this to clean up the commit message, not realizing you reviewed. If you're okay with this for now we can merge, unless you want the behavior documented somewhere first.

src/index.js
benmonro: Nope that's cool just checking

PR #331, opened 2020-06-09
feat(hover): Add hover API (fixes #254)

### LINKED ISSUE #254: Hover
A way to hover on elements to e.g. test that tooltips are working correctly.

nickmccurdy:
**What**: Add `hover()` for hovering elements

**Why**: Testing tooltips and other behavior involving mouse hovering

**How**: Calling various mouse events

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Typings
- [ ] Ready to be merged
      

__Note:__ I intentionally left the test failing because I could use some help figuring out how to fix it. For some reason, `mouseout` is not added to the list of events, despite being called in the function and `onMouseOut` working in a React test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,20 +407,20 @@ Hovers over `element`.
 import React from 'react'
 import {render, screen} from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
+import Tooltip from '../tooltip'
 
 test('hover', async () => {
-  const handler = jest.fn()
+  const messageText = 'Hello'
   render(
-    <button
-      data-testid="button"
-      onMouseEnter={handler}
-      onMouseOver={handler}
-      onMouseMove={handler}
-    />,
+    <Tooltip messageText={messageText}>
+      <TrashIcon aria-label="Delete" />
+    </Tooltip>,
   )
 
-  await userEvent.hover(screen.getByTestId('button'))
-  expect(handler).toHaveBeenCalledTimes(3)
+  await userEvent.hover(screen.getByLabelText(/delete/i))
+  expect(screen.getByText(messageText)).toBeInTheDocument()
+  await userEvent.unhover(screen.getByLabelText(/delete/i))
+  expect(screen.queryByText(messageText)).not.toBeInTheDocument()
 })
 ```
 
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -451,6 +451,14 @@ function tab({shift = false, focusTrap = document} = {}) {
   }
 }
 
+function hover(element, init) {
```
kentcdodds, line 454: Since everything's going to eventually be async (#312), maybe we should make sure new utils are async and have a tick?

For now, I think this would be:

```javascript
await tick()
fireEvent...
await tick()
fireEvent..

..etc.
```

But eventually it'll just be:

```javascript
await fireEvent..
await fireEvent..
await fireEvent..
```

where `fireEvent` is our own wrapper around the dtlFireEvent:

```javascript
async function fireEvent(...args) {
  await tick()
  return fireEvent(...args)
}
```

src/__tests__/hover.js
```diff
@@ -0,0 +1,35 @@
...
+test('hover', () => {
+  const {element, getEventCalls} = setup(<button />)
+
+  userEvent.hover(element)
+  expect(getEventCalls()).toMatchInlineSnapshot(`
+    mouseenter: Left (0)
+    mouseover: Left (0)
+    mousemove: Left (0)
+    mouseout: Left (0)
+    mouseleave: Left (0)
```
kentcdodds, line 15: I'm guessing a lot of people who want this will want to assert that a popover/tooltip showed up on hover. If we immediately mouseout and mouseleave before they get a chance to make their assertions then this won't help them...

Maybe we should have a `hover` and an `unhover` or something?

kentcdodds:
Thanks for working on this!

src/index.js
nickmccurdy: If you don't have a strong preference, I figured it would be easier to make that change separately.

kentcdodds: Right, all I'm saying is that as we're introducing something new, we should make it behave like it will in the near future (it should be async and there should be a tick between every fireEvent call).

README.md
```diff
@@ -398,6 +399,56 @@ it('should cycle elements in document tab order', () => {
 })
 ```
 
+### `async hover(element)`
+
```
kentcdodds, line 403: Let's try and make these examples more real-world like. I don't think they need to actually run with real implantation of components, but if they could just demonstrate the expected type of component to be tested with this that would be great.

May also help to note that this will not trigger pseudo states for css (I'm not sure whether that's possible).

src/__tests__/hover.js
```diff
@@ -0,0 +1,14 @@
...
+import {setup} from './helpers/utils'
+
+test('hover', async () => {
+  const {element, getEventCalls} = setup(<button />)
+
+  await userEvent.hover(element)
+  expect(getEventCalls()).toMatchInlineSnapshot(`
+    mouseover: Left (0)
+    mouseenter: Left (0)
+    mousemove: Left (0)
```
kentcdodds, line 12: Got the order of these via: https://jsbin.com/sulogop/1/edit?js,output

kentcdodds:
@all-contributors please add @nickmccurdy for code and tests

src/__tests__/hover.js
nickmccurdy: That's strange, I got the enter/over/move order from React, does it fire events in a different order than the DOM itself?

kentcdodds: It's possible. React does weird things to events, which is why my playground doesn't use react

PR #344, opened 2020-06-13
fix(selectOptions): don't select option if select is disabled

ben-dyer:
**What**: Prevent a select's value from changing with `selectOptions` if it is disabled

**Why**: It's possible to change the value of a disabled select using `selectOptions`. For instance this is passing:
```
  const {
    container: {firstChild: select},
  } = render(
    <select disabled>
      <option>No value selected</option>
      <option value="1">1</option>
      <option value="2">2</option>
      <option value="3">3</option>
    </select>,
  )

  expect(select.selectedIndex).toBe(0)

  userEvent.selectOptions(select, '1')
  expect(select.selectedIndex).toBe(1)
```

**How**: Exit early in `selectOptions` if the element is disabled, similar to what is done in `clear` and `upload`

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] Typings N/A
- [x] Ready to be merged
      

AFAIK it is ready to be merged, however I am not experienced with this project (only as a user) and there may be very good reasons why this was not already done this way.

### DIFF
--- a/src/__tests__/selectoptions.js
+++ b/src/__tests__/selectoptions.js
@@ -318,3 +318,19 @@ test('does not select anything if no matching options are given', () => {
   userEvent.selectOptions(select, 'Matches nothing')
   expect(select.selectedIndex).toBe(0)
 })
+
+test('does not select anything if select is disabled', () => {
+  const {
+    container: {firstChild: select},
+  } = render(
+    <select disabled>
+      <option>No value selected</option>
+      <option value="1">1</option>
+      <option value="2">2</option>
+      <option value="3">3</option>
+    </select>,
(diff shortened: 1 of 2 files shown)


ben-dyer:
Thanks for taking a look, I'm happy to expand it a bit to include options. Will update

kentcdodds:
You know, it's probably better to just leave this as-is and instead make a pull request to my branch that moves user event into DOM Testing Library: https://github.com/testing-library/dom-testing-library/pull/616

kentcdodds:
@all-contributors please add @ben-dyer for code and tests

Issue #365, opened 2020-06-17
Document body should be a focusable element


juanca:
- `@testing-library/user-event` version: `@testing-library/user-event@10.4.0`
- Testing Framework and version: `jest@25.1.0`
- DOM Environment: `jsdom@16.2.0`
- node: `node/10.20.1`

Relevant code or config (I'm also using React)

```javascript
it('manages the page tab sequence', () => {
  function TestComponent() {
    return (<ul role="listbox">
      <li role="option" tabindex="0">First option</li>
      <li role="option" tabindex="-1">Second option</li>
    </ul>);
  }

  render(<TestComponent />);

  expect(document.body).toHaveFocus();

  userEvent.tab();
  expect(screen.getByText('First option')).toHaveFocus();

  userEvent.tab();
  expect(document.body).toHaveFocus();

  userEvent.tab({ shift: true });
  expect(screen.getByText('First option')).toHaveFocus();
});
```

What you did:

Ran the test.

What happened:

The test failed. 

```
    expect(element).toHaveFocus()

    Expected
      <body><div><ul role="listbox"><li role="option" tabindex="0">First option</li><li role="option" tabindex="-1">Second option</li></ul></div></body>
    Received:
      <li role="option" tabindex="0">First option</li>

      76 | 
      77 |       userEvent.tab();
    > 78 |       expect(document.body).toHaveFocus();
         |                             ^
      79 | 
      80 |       userEvent.tab({ shift: true });
      81 |       expect(screen.getByText('First option')).toHaveFocus();
```

Reproduction repository:

Problem description:

It looks like the `tab` method does not consider the body as a tabable element. That is actually not true -- at least from doing a quick experiment in Chrome.

The following screenshot is the experiement. 

1. Render a button
1. Look at active element (body)
1. Press tab
1. Look at active element (button)
1. Press tab
1. Look at active element (body)

<img width="1566" alt="Screen Shot 2020-06-17 at 12 21 03 PM" src="https://user-images.githubusercontent.com/841084/84941849-c8488400-b096-11ea-8c52-3c95eb121e9a.png">

Suggested solution:

Assuming the focus trap is on the document, always set the body as the first tabable element.

juanca:
Possibly related: 

Looks like we need to integrate `act` in `userEvent.tab`. Otherwise, we should recommend using `act` with `tab`. I noticed the user event was not waiting for my stateful component updates. I can look into reproducing this with the same example: use state to manage tab indices (focus => change indices structure).

[...]

PR #368, opened 2020-06-19
Cycle focus between the body and page tab sequence

juanca:
Resolves https://github.com/testing-library/user-event/issues/365

**What**: Add `document.body` to the cycled elements in tab sequence

**Why**: The browser allows the active element to be `document.body` when the next selected element is outside the page tab sequence. When the active element is the body (e.g. initial state) then the next reasonable selected element is the first element in the page tab sequence. 

**How**: 

- Consolidate logic around getting the next element to focus
- If the current active element is on the edges of the page tab sequence then focus the `document.body` accordingly

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Typings
- [x] Ready to be merged
      

1. I was assuming that invoking `document.body.focus` would be undesirable -- opting for just `previousElement.blur` as a means of setting active element to the `body`. However, it seems that `document.body.focus` is working well enough. Do we want to add this kind of logic?

1. All tests pass when [not mutating `tabindex` for `tab`](https://github.com/testing-library/user-event/blob/f4eae9cf3c4bb5765dcdb7ce7f7652ae16e2f9a7/src/tab.js#L93-L96). Do we want to clean that up in this PR? Or is there some version of jsdom this is intended to support?

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,7 +402,11 @@ it('should cycle elements in document tab order', () => {
 
   userEvent.tab()
 
-  // cycle goes back to first element
+  // cycle goes back to the body element
+  expect(document.body).toHaveFocus()
+
+  userEvent.tab()
+
   expect(checkbox).toHaveFocus()
 })
 ```
@@ -545,6 +549,7 @@ Thanks goes to these people ([emoji key][emojis]):
 
 <!-- markdownlint-enable -->
 <!-- prettier-ignore-end -->
+
 <!-- ALL-CONTRIBUTORS-LIST:END -->
 
 This project follows the [all-contributors][all-contributors] specification.
--- a/src/__tests__/tab.js
+++ b/src/__tests__/tab.js
@@ -111,7 +111,11 @@ test('should cycle elements in document tab order', () => {
 
   userEvent.tab()
 
-  // cycle goes back to first element
+  // cycle goes back to the body
+  expect(document.body).toHaveFocus()
+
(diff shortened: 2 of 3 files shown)


nickmccurdy:
> I was assuming that invoking `document.body.focus` would be undesirable -- opting for just `previousElement.blur` as a means of setting active element to the `body`. However, it seems that `document.body.focus` is working well enough. Do we want to add this kind of logic?

It's consistent with JSDOM semantics, so it should be fine. Not sure if this would break if someone wasn't using JSDOM, but that's pretty unlikely and we could always patch this later if it becomes an issue.

> All tests pass when not mutating tabindex for tab. Do we want to clean that up in this PR? Or is there some version of jsdom this is intended to support?

Not sure I understand the context here, is it only an issue in some versions of JSDOM?

juanca:
>> All tests pass when not mutating tabindex for tab. Do we want to clean that up in this PR? Or is there some version of jsdom this is intended to support?

> Not sure I understand the context here, is it only an issue in some versions of JSDOM?

I would guess so. But I also haven't investigated much (besides commenting out relevant pieces of code). I found out because I was trying to understand what was happening in the `tab` method. Perhaps something to think about later (or not).

nickmccurdy:
I'm not sure, it's your call

kentcdodds:
@all-contributors please add @juanca for code and tests

kentcdodds:

PR #370, opened 2020-06-20
fix: update onChange/input events fired on selected option

### LINKED ISSUE #358: userEvent.selectOptions does not trigger onChange event
- `@testing-library/user-event` version: 12.0.0
- Testing Framework and version: jest@24.7.1
  
- DOM Environment: jsdom@14.1.0
  

Relevant code or config

```javascript
// Select.js
import React from "react";

const Select = props => {
  const handleChange = e => {
    console.log(e.target.value);
    props.test(e.target.value);

  };
  return (
    <select data-testid="select" onChange={handleChange}>
      <option data-testid="first" value="first">
        first
      </option>
      <option data-testid="second" value="second">
        second
      </option>
      <option data-testid="third" value="third">
        third
      </option>
    </select>
  );
};
export default Select;

// Select.test.js
  import React from 'react';
  import Select from './Select';
  import userEvent from '@testing-library/userEvent';
  import render from '@testing-library/react';


  test("`userEvent.selectOptions` should trigger an onChange event", () => {
    const mockTest = jest.fn().mockImplementation(val => console.log(val));
    const { getByTestId, debug } = render(<Select test={mockTest} />);
    const select = getByTestId("select");
    const second = getByTestId("second");

    userEvent.selectOptions(select, ["second"]);
    debug();
    console.log(select.value)
    expect(mockTest).toHaveBeenCalledTimes(1);
    expect(mockTest).toHaveBeenCalledWith('second');
    expect(second.selected).toBe(true);
  });
```

What you did:
I first noticed this behavior in a larger project, so I made a small repo to reproduce. 

I made a `<select/>` element in React that fires a function received from props onChange. This behavior works as expected in the browser. When testing the behavior, I can successfully assert that a mocked version of that onChange function is called when I trigger a change event (via importing `fireEvent` from RTL), but cannot assert that the same mock is called when using `userEvent.selectOptions`. 

What happened:
Screen Shot 2020-06-16 at 9 28 24 AM

Reproduction repository: 
https://github.com/khalidwilliams/select-test-demo

Problem description:
`userEvent.selectOptions` doesn't trigger a change event. onChange handlers won't fire in tests that use this method, leading to problems with testing any change-dependent behavior. 

Suggested solution:
I'm not sure what the best solution is, it seems like [this line](https://github.com/testing-library/user-event/blob/master/src/select-options.js#L64) may not be firing as expected? Please let me know if there's something I'm missing here -- I'm happy to keep digging!

marcosvega91:
fix #358 

**What**:

This PR will fix the issue that events (onChage/input) are not fired on ancestors.
This issue was related to a problem with React because it fires event on bubbling phase. 
As we can see from the images below 

the change event is an `Event` of type `change` 
image

the input event is an `Event` of type `input`
image

Both events are of a generic event type. 

**Why**:

Because in particular with React the change event is fired but the onChange listener didn't work

**How**:

By calling 
```
fireEvent(
  select,
  createEvent('input', select, {
    bubbles: true,
    cancelable: false,
    composed: true,
    ...init,
  }),
)
fireEvent.change(select, init)
```

**Checklist**:

- [ ] Documentation
- [X] Tests
- [ ] Typings
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/select-options.js
+++ b/src/__tests__/select-options.js
@@ -133,3 +133,19 @@ test('does not select anything if options are disabled', () => {
   expect(o2.selected).toBe(false)
   expect(o3.selected).toBe(false)
 })
+
+test('should call onChange bubbling up the event when a new option is selected', () => {
+  const {select, form} = setupSelect({multiple: true})
+  const onChangeSelect = jest.fn()
+  const onChangeForm = jest.fn()
+  select.addEventListener('change', onChangeSelect)
+  form.addEventListener('change', onChangeForm)
+
+  expect(onChangeSelect.mock.calls).toHaveLength(0)
+  expect(onChangeForm.mock.calls).toHaveLength(0)
+
+  userEvent.selectOptions(select, ['1'])
+
+  expect(onChangeForm.mock.calls).toHaveLength(1)
+  expect(onChangeSelect.mock.calls).toHaveLength(1)
+})
--- a/src/__tests__/select-options.js
+++ b/src/__tests__/select-options.js
@@ -134,18 +134,26 @@ test('does not select anything if options are disabled', () => {
   expect(o3.selected).toBe(false)
 })
 
-test('should call onChange bubbling up the event when a new option is selected', () => {
+test('should call onChange/input bubbling up the event when a new option is selected', () => {
   const {select, form} = setupSelect({multiple: true})
   const onChangeSelect = jest.fn()
   const onChangeForm = jest.fn()
+  const onInputSelect = jest.fn()
+  const onInputForm = jest.fn()
   select.addEventListener('change', onChangeSelect)
+  select.addEventListener('input', onInputSelect)
   form.addEventListener('change', onChangeForm)
+  form.addEventListener('input', onInputForm)
 
   expect(onChangeSelect.mock.calls).toHaveLength(0)
   expect(onChangeForm.mock.calls).toHaveLength(0)
+  expect(onInputSelect.mock.calls).toHaveLength(0)
+  expect(onInputForm.mock.calls).toHaveLength(0)
 
   userEvent.selectOptions(select, ['1'])
 
   expect(onChangeForm.mock.calls).toHaveLength(1)
   expect(onChangeSelect.mock.calls).toHaveLength(1)
+  expect(onInputSelect.mock.calls).toHaveLength(1)
+  expect(onInputForm.mock.calls).toHaveLength(1)
 })
--- a/src/__tests__/select-options.js
+++ b/src/__tests__/select-options.js
@@ -140,20 +140,22 @@ test('should call onChange/input bubbling up the event when a new option is sele
   const onChangeForm = jest.fn()
   const onInputSelect = jest.fn()
   const onInputForm = jest.fn()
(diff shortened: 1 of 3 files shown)


src/__tests__/select-options.js
```diff
@@ -133,3 +133,27 @@ test('does not select anything if options are disabled', () => {
...
+  expect(onChangeForm.mock.calls).toHaveLength(0)
+  expect(onInputSelect.mock.calls).toHaveLength(0)
+  expect(onInputForm.mock.calls).toHaveLength(0)
+
+  userEvent.selectOptions(select, ['1'])
+
+  expect(onChangeForm.mock.calls).toHaveLength(1)
+  expect(onChangeSelect.mock.calls).toHaveLength(1)
+  expect(onInputSelect.mock.calls).toHaveLength(1)
+  expect(onInputForm.mock.calls).toHaveLength(1)
```
nickmccurdy, line 158: These could probably use `expect(something).toHaveBeenCalledTimes(number)`. Alternatively, this test could also use an inline snapshot, though you can only have one per test.

src/__tests__/select-options.js
```diff
@@ -133,3 +133,27 @@ test('does not select anything if options are disabled', () => {
...
   expect(o3.selected).toBe(false)
 })
+
+test('should call onChange/input bubbling up the event when a new option is selected', () => {
+  const {select, form} = setupSelect({multiple: true})
+  const onChangeSelect = jest.fn()
+  const onChangeForm = jest.fn()
+  const onInputSelect = jest.fn()
+  const onInputForm = jest.fn()
+  select.addEventListener('change', onChangeSelect)
```
kentcdodds, line 143: The test utils exposes an addEventListener helper that should be used instead (to avoid memory leaks by benefiting from automatic cleanups).

marcosvega91:
Thanks for your reviews 😄 . I have fixed what was wrong

kentcdodds:
@nickmccurdy, I'll let you merge this one. You should be able to once the build is successful. Just make sure to use squash and merge and follow the commit message convention so we get the right release for it (I think it should be good without change actually). Actually you'll want to add the text "Closes #358" somewhere in the body of the commit to semantic release will add a release comment to that issue.

Thanks!

fabb:
Will this fix this issue where `userEvent.selectOptions` does not correctly update Formik state?
https://codesandbox.io/s/usereventselectoptions-bug-gn8wb

Or should I create a separate bug?

kentcdodds:
@fabb, I don't know. Can you try when this is released?

kentcdodds:
Hey @marcosvega91, you've been added as an official member of the GitHub org :) Thanks for all you do!

marcosvega91:
> Hey @marcosvega91, you've been added as an official member of the GitHub org :) Thanks for all you do!

Thank you @kentcdodds. I don't have other words, I started to contribute some months ago after one of your tweets and now I'm a member of testing lib. I discovered a world and I'm learning a lot thanks to you. 

Thank you for this opportunity and for your trust 🎉 

Regards 
Marco

fabb:
> @fabb, I don't know. Can you try when this is released?

@kentcdodds yes, it's fixed, thanks!

Issue #354, opened 2020-01-07
arrow keys?


benmonro:
for the life of me I can't seem to be able to fire an event for a keypress with arrow keys on a radio button, is this a problem in jsdom, or am I crazy?

 ```js
   fireEvent.keyPress(myRadioButton, { key: 'ArrowUp', which: 38, keyCode: 38 });
```

brandonpittman:
Just found myself looking for this feature too.

PR #371, opened 2020-06-20
Implement key press functionality from a user perspective

juanca:
**What**: Implement user interaction of pressing a single key with some modifiers

**Why**: This is especially useful for testing accessibility interactions (keyboard interactions) on UI elements.

**How**: 

- Interaction is solely on the active element because a user will always be on _some_ element
- Any key is "pressable" given the key name
- The 4 popular modifiers (alt, ctrl, meta, shift) are implemented and fired around the pressed key
  - Looks like there might be more key to implement but I think we can work on these in the future

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Typings
- [ ] Ready to be merged
      

Closes https://github.com/testing-library/user-event/issues/354

A couple of notes:

- The event is always fired on the active element since a user will always perform an interaction on the active element
- Event `which` key is not implement because [it is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which)
- Event `code` / `keyCode` key is not implement because it can vary on type of key (`KeyA`, `Digit1`, `AltLeft`, etc), opting for a more simple interface

### DIFF
--- a/src/__tests__/key-press.js
+++ b/src/__tests__/key-press.js
@@ -1,0 +1,264 @@
+import userEvent from '../'
+import {setup} from './helpers/utils'
+
+test('presses arrow keys on active element', () => {
+  const {element, getEventSnapshot} = setup('<div tabindex="0"/>')
+  userEvent.click(element)
+  userEvent.keyPress('ArrowDown')
+  userEvent.keyPress('ArrowLeft')
+  userEvent.keyPress('ArrowRight')
+  userEvent.keyPress('ArrowUp')
+  expect(getEventSnapshot()).toMatchInlineSnapshot(`
+    Events fired on: div
+
+    div - pointerover
+    div - pointerenter
+    div - mouseover: Left (0)
+    div - mouseenter: Left (0)
+    div - pointermove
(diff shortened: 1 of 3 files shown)


juanca:
I can whip up documentation if we're okay with this new API. 

I am not sure what to do with typings. Is there some work I need for that particular checklist item?

kentcdodds:
I'd prefer this functionality be built in via the `type` utility. So: `type(input, '{ArrowDown})`...

nickmccurdy:
For reference: #354

juanca:
I started digging into this and I think I discovered unplanned features:

I am thinking the PR will add support for [navigation keys](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values#Navigation_keys). It's pretty simple if I am only concerned about keyboard navigation for accessibility (simple keydown + keyup). However, it gets more complex: the `type` method involves managing the value and cursor/selection state in various elements; the navigation keys also have a side effect of changing the cursors or selection range.

I'll give it another go on Friday.

Issue #356, opened 2020-06-15
activeElement might change during userEvent.type causing runtime errors


beckerei:
- `@testing-library/user-event` version: >= 11.1.0
- `@testing-library/dom` version: 5.10.1
- Testing Framework and version: jest@26.0.1
- DOM Environment: jsdom jest-environment-jsdom-global@2.0.2
- node: 13.7.0

### What you did:

Using `userEvent.type`. We have like > 1k tests which mostly use `fireEvent` but slowly upgrading to `userEvent`. All run fine, execept for one I encountered while updating our dependencies.

As a tmp fix one can use `fireEvent.change`.

### What happened:
I narrowed it down to start with the release of https://github.com/testing-library/user-event/releases/tag/v11.1.0
```
    TypeError: Cannot read property 'length' of undefined

      at calculateNewValue (node_modules/@testing-library/user-event/dist/type.js:312:43)
      at typeCharacter (node_modules/@testing-library/user-event/dist/type.js:366:43)
      at typeImpl (node_modules/@testing-library/user-event/dist/type.js:230:29)
      at node_modules/@testing-library/user-event/dist/type.js:20:14
      at node_modules/@testing-library/react/dist/pure.js:58:16
```
### Problem description:
We currently have 90 occurrences for `await userEvent.type(` in our codebase, yet only a single one has this issue. I don't understand (yet) what could cause this.
But I figured out the `activeElement` is always the body, so calling `currentValue()` will be undefined, as `body` has no value prop.

I will provide more details if I find more.

tsmithATX:
I'm seeing this happen with portals where the portal component opens and has an active element but once you try to focus back to a text area to type the activeElement is the root of the portal.

tsmithATX:
v12.0.14 addressed our issues, no longer seeing this happen!

Junhao6503:
I am still encountering this issue due to portals. It gives me errors like:
```
TypeError: Cannot read property 'length' of undefined
      at setSelectionRange (node_modules/@testing-library/user-event/dist/type.js:240:70)
      at fireInputEventIfNeeded (node_modules/@testing-library/user-event/dist/type.js:269:5)
      at handleDel (node_modules/@testing-library/user-event/dist/type.js:613:5)
      at node_modules/@testing-library/user-event/dist/type.js:144:29
      at typeImpl (node_modules/@testing-library/user-event/dist/type.js:159:4)
      at type (node_modules/@testing-library/user-event/dist/type.js:73:14)
[...]

PR #375, opened 2020-06-21
fix: add message if type/paste function is called with an invalid element

marcosvega91:
**What**:

I have thrown a new `TypeError` if the user calls `userEvent.type` or `userEvent.paste` with an element that has no value. 
This PR is part of issue #356.

**Why**:

Because as it was before the PR the error was not very helpful

image

**How**:

By checking if the currentElement has value property. 
The message thrown is like this `the current element is of type BODY and doesn't have a valid value`

**Checklist**:

- [ ] Documentation
- [X] Tests
- [ ] Typings
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/paste.js
+++ b/src/__tests__/paste.js
@@ -94,3 +94,12 @@ test('should replace selected text all at once', () => {
   userEvent.paste(element, 'friend')
   expect(element).toHaveValue('hello friend')
 })
+
+test('should give error if we are trying to call paste on an invalid element', () => {
+  const {element} = setup('<div  />')
+  expect(() =>
+    userEvent.paste(element, "I'm only a div :("),
+  ).toThrowErrorMatchingInlineSnapshot(
+    `"the current element is of type DIV and doesn't have a valid value"`,
+  )
+})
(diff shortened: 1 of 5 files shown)


marcosvega91:
I'm looking the code better and maybe there could be other problems like this
https://github.com/testing-library/user-event/blob/50e330bc1d7249eea618405f8e3f70e32077ef07/src/type.js#L253-L254
https://github.com/testing-library/user-event/blob/50e330bc1d7249eea618405f8e3f70e32077ef07/src/type.js#L286-L287

Maybe we should use `currentValue` also in the other parts?

kentcdodds:
Whatever you think makes the most sense 👍

marcosvega91:
I have made the changes. I have updated also paste function with this logic

kentcdodds:
@all-contributors please add @marcosvega91 for code and tests

kentcdodds:

marcosvega91:
thank you @kentcdodds, it's awesome. I'm really happy to be part of another team of testing library 🥳 .
I like a lot the project and I hope to do always good thing 🦇 

Thank you for all the things that you are doing for the community and the opportunity that you give to all of us ❤️

PR #381, opened 2020-06-25
fix: React onChange/onInput events not triggering on input type="file" file upload

### LINKED ISSUE #380: input type="file" - upload() doesn't trigger onChange in React
- `@testing-library/user-event` version: 12.0.0+
- Testing Framework and version: jest@26.1.0
- DOM Environment: jsdom@16.2.2

Relevant code or config:

All relevant code is in the codesandbox, no special config needed.

What you did:

I have a project with a form that includes a `<input type="file"/>`. I wrote a test for the form that was working well up until I upgraded from v11 to v12. I created a simple codesandbox that reproduces the issue.
The repro includes a simple form with just the relevant input element and a form clear button because the value is read-only and I wanted to make sure the test is clean. The test calls `userEvent.upload` to upload a fake file and checks if the callback inside of `onChange` was called.

What happened:

In the small repro sandbox the test is passing on v11 (11.4.2 specifically, which is the latest) but fails on 12.0.0.

image

Reproduction repository:

https://codesandbox.io/s/testing-library-user-event-file-input-change-event-728w8

Problem description:

In React, `onChange` is not triggered when using `userEvent.upload(...)` to add a file to a `<input type="file"/>`.
When I came to file this issue I found [this similar issue](https://github.com/testing-library/user-event/issues/358) that already has a merged fix PR so I suspect the solution should be pretty straight forward requiring a similar for `upload` as was changed for `selectOptions`. Considering this is the second function that broke in the same way, it's probably worth it to check if there are more similar cases as well.

Suggested solution:

The solution should be very similar to [this merged PR](https://github.com/testing-library/user-event/pull/370). I can try to get a PR for it this evening but I figured I'd file the issue now in case someone else can get it done earlier.

ybentz:
Fixes #380.

**What**:
This PR fixes a bug in the `.upload()` function that's very similar to #370. The cause of the bug and the fix are basically the same.

**Why**:
In React, starting with user-event v12 the `onChange` and `onInput` callbacks stopped working for `<input type="file" />`.

**How**:
Very similar to #370:
- Input event: passing the `bubbles: true` option to the event's options.
- Change event: Use `fireEvent.change` instead of `fireEvent(input, createEvent('change', ...`.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] Typings N/A
- [x] Ready to be merged

**Notes**:
As I mentioned in the [related issue](#380), we should probably check the rest of the changes in case some other similar functionality broke with the recent changes.

### DIFF
--- a/src/__tests__/upload.js
+++ b/src/__tests__/upload.js
@@ -61,6 +61,8 @@ test('should fire the correct events with label', () => {
     label[for="element"] - click: Left (0)
     input#element[value=""] - click: Left (0)
     input#element[value=""] - focusin
+    input#element[value=""] - input
+    input#element[value=""] - change
   `)
 })
 
--- a/src/__tests__/upload.js
+++ b/src/__tests__/upload.js
@@ -1,5 +1,5 @@
 import userEvent from '../'
-import {setup} from './helpers/utils'
+import {setup, addListeners} from './helpers/utils'
 
 test('should fire the correct events for input', () => {
   const file = new File(['hello'], 'hello.png', {type: 'image/png'})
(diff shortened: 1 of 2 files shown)


kentcdodds:
@all-contributors please add @ybentz for code and tests

kentcdodds:

PR #392, opened 2020-06-26
Disable package lock

nickmccurdy:
**What**: Disable generation of `package-lock.json` completely in npm

**Why**: We are only ignoring the lock from being committed, which is not enough to disable it. The package lock still exists, so issues can occur with npm trying to deal with changing dependency versions across branches and clones, thinking that the lock is being committed everywhere. This also causes the lock to be unnecessarily generated in CI, only to be deleted after the build completes.

**How**: Add `package-lock=false`, just like the other repositories in this organization

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests
- [ ] Typings N/A
- [ ] Ready to be merged

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 registry=http://registry.npmjs.org/
+package-lock=false


nickmccurdy:
Going to merge this since it passes CI, I don't see any disadvantages given all our other projects are already doing this. Feel free to discuss it though.

PR #397, opened 2020-07-08
fix: submit form on ENTER pressed

marcosvega91:
**What**: I have fixed a bug with type function

**Why**: As for the [tweet](https://twitter.com/jacobmparis/status/1280875565885554688) when you run `type` function on an `input` inside a `form` it should `submit` the form if it has a listener. 

**How**: If the `currentElement` is an input and if it has a form I have fired a `submit` event on form

**Checklist**:

- [ ] Documentation
- [X] Tests
- [ ] Typings
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -758,14 +758,18 @@ test('typing on input type file should not result in an error', () => {
 })
 
 test('should submit a form when ENTER is pressed on input', () => {
-  const {element} = setup(`
+  const handleSubmit = jest.fn()
+  const {element} = setup(
+    `
     <form>
       <input type='text'/>
     </form>
-  `)
-  const onSubmit = jest.fn()
-  element.addEventListener('submit', onSubmit)
+  `,
+    {
+      eventHandlers: {submit: handleSubmit},
+    },
+  )
   userEvent.type(element.querySelector('input'), '{enter}')
 
-  expect(onSubmit).toHaveBeenCalledTimes(1)
+  expect(handleSubmit).toHaveBeenCalledTimes(1)
 })
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -756,3 +756,16 @@ test('typing on input type file should not result in an error', () => {
   expect.assertions(0)
   return userEvent.type(element, 'bar').catch(e => expect(e).toBeUndefined())
 })
+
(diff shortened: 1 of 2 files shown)


src/__tests__/type.js
```diff
@@ -715,3 +715,16 @@ test('should give error if we are trying to call type on an invalid element', as
...
+  const {element} = setup(`
+    <form>
+      <input type='text'/>
+    </form>
+  `)
+  const onSubmit = jest.fn()
+  element.addEventListener('submit', onSubmit)
+  userEvent.type(element.querySelector('input'), '{enter}')
+
+  expect(onSubmit).toHaveBeenCalledTimes(1)
```
kentcdodds, line 729:
```suggestion
  const handleSubmit = jest.fn()
  const {element} = setup(`<form><input type="text" /></form>`, {
    eventHandlers: {submit: handleSubmit},
  })
  userEvent.type(element.querySelector('input'), '{enter}')

  expect(handleSubmit).toHaveBeenCalledTimes(1)
```

kentcdodds: Hit the wrong button. This won't work after all...

However, the way it's currently coded doesn't clean up the event handler.... One sec.

marcosvega91: right I always forget that this setup has this param 🤦  sorry

kentcdodds:
Ok, sorry for the confusing emails I'm sure you received. I cleaned up my comments and my suggestion is valid :)

kentcdodds:
Sorry I lost track of this one. Could you resolve the conflicts and we can get this merged.

marcosvega91:
I have rebased the current branch :)

Jessidhia:
(this is not quite the implicit submission algorithm; see https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission for the full story; but it's good enough if you're not actually testing the spec or relying on submit blockers)

kentcdodds:
Thanks for that @Jessidhia! We should probably improve this to match what browsers actually do... @marcosvega91, would you be interested in doing that investigation?

marcosvega91:
Hi @Jessidhia thanks for raising it.

I think that the only problem that we have is that if a form has more than one field we can't trigger click without a `submit` button. I'll open a new PR for this

PR #400, opened 2020-07-10
fix: type into date input

### LINKED ISSUE #399: userEvent.type doesn't seem to support `input` fields of `type="date"` (native datepicker)
- `@testing-library/user-event` version: 12.0.11
- Testing Framework and version: jest (out of the box with Codesandbox MUI issue template)
  
- DOM Environment: Out of the box with Codesandbox MUI issue template
  

Relevant code or config

For an MUI native datepicker such as (which ultimately renders in the DOM as an `input` component of `type="date"`):

```javascript
<TextField
    id={`nativeDatepicker`}
    type="date"
    defaultValue={moment().toISOString()}
    onChange={event => handleDateValueUpdate(event.target.value)}
    error={false}
    helperText={""}
    disabled={false}
    inputProps={{
      "data-testid": `inputId-nativeDatepicker`
    }}
    required={true}
  />
```

`userEvent.type` will not call onChange with the specified `event.target.value`:

```javascript
it("should update with userEvent.type", async () => {
    const handleDateValueUpdate = jest.fn();
    render(<NativeDatepicker handleDateValueUpdate={handleDateValueUpdate} />);
    const nativeDatepicker = await screen.findByTestId(
      "inputId-nativeDatepicker"
    );
    userEvent.type(nativeDatepicker, "2020-01-02");
    expect(handleDateValueUpdate).toHaveBeenCalledWith("2020-01-02");
  });
```

Results:
```
expect(jest.fn()).toHaveBeenCalledWith(expected)

Expected mock function to have been called with:
  ["2020-01-02"]
But it was called with:
  [""]

  11 |       "inputId-nativeDatepicker"
  12 |     );
  13 |     userEvent.type(nativeDatepicker, "2020-01-02");
> 14 |     expect(handleDateValueUpdate).toHaveBeenCalledWith("2020-01-02");
  15 |   });
  16 | });
```

What you did:
I put the relevant code on a sandbox here:
https://codesandbox.io/s/heuristic-sun-dqfby?file=/src/NativeDatepicker.test.js

What happened:
I'd expect the testing library to be able to update the date field with the provided value, but it doesn't seem to happen. 

Results:
```
expect(jest.fn()).toHaveBeenCalledWith(expected)

Expected mock function to have been called with:
  ["2020-01-02"]
But it was called with:
  [""]

  11 |       "inputId-nativeDatepicker"
  12 |     );
  13 |     userEvent.type(nativeDatepicker, "2020-01-02");
> 14 |     expect(handleDateValueUpdate).toHaveBeenCalledWith("2020-01-02");
  15 |   });
  16 | });
```

Reproduction repository:

https://codesandbox.io/s/heuristic-sun-dqfby?file=/src/NativeDatepicker.test.js

Problem description:
I want to be able to update an input date field in my integration tests using userEvent, but not sure if the `type` method supports input fields of type date. 

Suggested solution:
Support input fields of type date.

lourenci:
**What**:

It closes #399.

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Typings
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -602,11 +602,11 @@ test('can type into an input with type `email`', () => {
 
 test('can type into an input with type `date`', () => {
   const {element, getEventSnapshot} = setup('<input type="date" />')
-  const date = '2020-07-09'
+  const date = '2020-06-29'
   userEvent.type(element, date)
   expect(element).toHaveValue(date)
   expect(getEventSnapshot()).toMatchInlineSnapshot(`
-    Events fired on: input[value="2020-07-09"]
+    Events fired on: input[value="2020-06-29"]
 
     input[value=""] - pointerover
     input[value=""] - pointerenter
@@ -623,45 +623,37 @@ test('can type into an input with type `date`', () => {
     input[value=""] - click: Left (0)
     input[value=""] - keydown: 2 (50)
     input[value=""] - keypress: 2 (50)
-    input[value=""] - input
     input[value=""] - keyup: 2 (50)
     input[value=""] - keydown: 0 (48)
     input[value=""] - keypress: 0 (48)
-    input[value=""] - input
     input[value=""] - keyup: 0 (48)
     input[value=""] - keydown: 2 (50)
     input[value=""] - keypress: 2 (50)
-    input[value=""] - input
     input[value=""] - keyup: 2 (50)
     input[value=""] - keydown: 0 (48)
     input[value=""] - keypress: 0 (48)
-    input[value=""] - input
     input[value=""] - keyup: 0 (48)
     input[value=""] - keydown: - (45)
     input[value=""] - keypress: - (45)
-    input[value=""] - input
     input[value=""] - keyup: - (45)
     input[value=""] - keydown: 0 (48)
     input[value=""] - keypress: 0 (48)
-    input[value=""] - input
     input[value=""] - keyup: 0 (48)
-    input[value=""] - keydown: 7 (55)
-    input[value=""] - keypress: 7 (55)
-    input[value=""] - input
-    input[value=""] - keyup: 7 (55)
+    input[value=""] - keydown: 6 (54)
+    input[value=""] - keypress: 6 (54)
+    input[value=""] - keyup: 6 (54)
     input[value=""] - keydown: - (45)
     input[value=""] - keypress: - (45)
-    input[value=""] - input
     input[value=""] - keyup: - (45)
-    input[value=""] - keydown: 0 (48)
-    input[value=""] - keypress: 0 (48)
-    input[value=""] - input
-    input[value=""] - keyup: 0 (48)
+    input[value=""] - keydown: 2 (50)
+    input[value=""] - keypress: 2 (50)
+    input[value=""] - keyup: 2 (50)
     input[value=""] - keydown: 9 (57)
     input[value=""] - keypress: 9 (57)
-    input[value="2020-07-09"] - input
-      "{CURSOR}" -> "{CURSOR}2020-07-09"
-    input[value="2020-07-09"] - keyup: 9 (57)
+    input[value="2020-06-29"] - input
+      "{CURSOR}" -> "{CURSOR}2020-06-29"
+    input[value="2020-06-29"] - change
+    input[value="2020-06-29"] - keyup: 9 (57)
   `)
 })
 
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -600,6 +600,13 @@ test('can type into an input with type `email`', () => {
   expect(element).toHaveValue(email)
 })
 
+test('can type into an input with type `date`', () => {
+  const {element} = setup('<input type="date" />')
+  const date = '2020-07-09'
+  userEvent.type(element, date)
+  expect(element).toHaveValue(date)
+})
+
 // https://github.com/testing-library/user-event/issues/336
 test('can type "-" into number inputs', () => {
   const {element, getEventSnapshot} = setup('<input type="number" />')
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -601,10 +601,68 @@ test('can type into an input with type `email`', () => {
 })
 
 test('can type into an input with type `date`', () => {
-  const {element} = setup('<input type="date" />')
+  const {element, getEventSnapshot} = setup('<input type="date" />')
   const date = '2020-07-09'
   userEvent.type(element, date)
   expect(element).toHaveValue(date)
+  expect(getEventSnapshot()).toMatchInlineSnapshot(`
+    Events fired on: input[value="2020-07-09"]
+
+    input[value=""] - pointerover
+    input[value=""] - pointerenter
+    input[value=""] - mouseover: Left (0)
+    input[value=""] - mouseenter: Left (0)
+    input[value=""] - pointermove
+    input[value=""] - mousemove: Left (0)
+    input[value=""] - pointerdown
+    input[value=""] - mousedown: Left (0)
+    input[value=""] - focus
+    input[value=""] - focusin
+    input[value=""] - pointerup
+    input[value=""] - mouseup: Left (0)
+    input[value=""] - click: Left (0)
+    input[value=""] - keydown: 2 (50)
+    input[value=""] - keypress: 2 (50)
+    input[value=""] - input
+    input[value=""] - keyup: 2 (50)
+    input[value=""] - keydown: 0 (48)
+    input[value=""] - keypress: 0 (48)
+    input[value=""] - input
+    input[value=""] - keyup: 0 (48)
+    input[value=""] - keydown: 2 (50)
+    input[value=""] - keypress: 2 (50)
+    input[value=""] - input
+    input[value=""] - keyup: 2 (50)
+    input[value=""] - keydown: 0 (48)
+    input[value=""] - keypress: 0 (48)
+    input[value=""] - input
+    input[value=""] - keyup: 0 (48)
+    input[value=""] - keydown: - (45)
+    input[value=""] - keypress: - (45)
+    input[value=""] - input
+    input[value=""] - keyup: - (45)
+    input[value=""] - keydown: 0 (48)
+    input[value=""] - keypress: 0 (48)
+    input[value=""] - input
+    input[value=""] - keyup: 0 (48)
+    input[value=""] - keydown: 7 (55)
+    input[value=""] - keypress: 7 (55)
+    input[value=""] - input
+    input[value=""] - keyup: 7 (55)
+    input[value=""] - keydown: - (45)
+    input[value=""] - keypress: - (45)
+    input[value=""] - input
+    input[value=""] - keyup: - (45)
+    input[value=""] - keydown: 0 (48)
+    input[value=""] - keypress: 0 (48)
+    input[value=""] - input
+    input[value=""] - keyup: 0 (48)
+    input[value=""] - keydown: 9 (57)
+    input[value=""] - keypress: 9 (57)
+    input[value="2020-07-09"] - input
+      "{CURSOR}" -> "{CURSOR}2020-07-09"
+    input[value="2020-07-09"] - keyup: 9 (57)
+  `)
 })
 
 // https://github.com/testing-library/user-event/issues/336
--- a/src/type.js
+++ b/src/type.js
@@ -243,8 +243,16 @@ function fireInputEventIfNeeded({
   newValue,
   newSelectionStart,
   eventOverrides,
+  textToBeTyped,
 }) {
   const prevValue = currentElement().value
+
+  if (
+    currentElement().type === 'date' &&
+    !isValidDateValue(currentElement(), textToBeTyped)
+  )
+    return {prevValue}
+
   if (
     !currentElement().readOnly &&
     !isClickable(currentElement()) &&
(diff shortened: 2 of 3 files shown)


src/type.js
```diff
@@ -222,6 +226,10 @@ async function typeImpl(
           newEntry = `${prevValue}.${char}`
         }
 
+        if (currentElement().type === 'date' && charPosition === 9) {
```
kentcdodds, line 229: Hmmmmm.... This seems to be coded for a specific use case (9 characters) rather than being very general. Are we sure that we'll be firing all the right events with these changes?

lourenci: It's a first naive implementation just to make its behavior green. :) I tried to not call any attention with a draft PR, sorry for that.

The problem with the issue is: we just can type the entire value, the input doesn't accept any invalid date. I think it was the issue with the input types too (number with `.` for example).
As for the date type, I don't have any current value in the interim to keep it around it.

I don't know yet what are the correct events to be triggered here. I think we can dispatch all of them (numbers and dashes in the same order passed to `type` (the `text`). What do you think?

Thank you.

kentcdodds: I think it's fine to do something like this do long as we fire all the same events a user would fire as they type into the date input. You might test it out here to see what events should be fired (it's possible what you have is correct): https://codesandbox.io/s/user-event-playground-eo909

I'm mostly concerned about 9 characters hard coded here. Will it always be exactly 9 characters?

lourenci: The events triggered in this PR are a little different from the playground: the browser triggers `input` (together with a `change`) events as soon as I start to type a valid date:

My input is `dd/07/2020`, I just need to type the day:

- when I type 2: `02/07/2020`:
```
input#date[value="2020-07-02"] - input
input#date[value="2020-07-02"] - change

```
- when I type 2 again: `22/07/2020`:
```
input#date[value="2020-07-22"] - input
input#date[value="2020-07-22"] - change
```

lourenci: Together with the `input` event, the browsers trigger a `change` event as soon as the date is valid. A little odd compared to the others.

kentcdodds: Is this different for other locales? I think this might be ok, though maybe it would be more resilient to try something like this:

```javascript
function isValidValue(el, value) {
  const clone = el.cloneNode()
  clone.value = value
  return clone.value === value
}
```

If the value is valid then we know we don't need to do anything special. Thoughts?

lourenci: Sorry for the delay.

I think it should work to avoid the hardcoded `charPosition`, but we'll have to keep some var around to control the `value` (from `isValueValud`) since we cannot type the `char` (what we have inside the `runCallbacks`) in a date input. Is't make sense for you or I'm missing something?

As for the events, just to make it clear: date input just triggers an `input` event (plus a `change` event) on a valid date according to the user-event playground. It's different from other events, though it's easy to implement it. Let's for it?

Thanks.

kentcdodds:
I'll wait for you mark it as ready to review before merging.

lourenci:
Hi, @kentcdodds. Did you read [this comment](https://github.com/testing-library/user-event/pull/400#discussion_r456793687)?  I'm not sure about the events yet since the events we are triggering here differ from the user-events playground. Anyway, I'll mark this to review, feel free to call a discussion about the events.

src/type.js
kentcdodds: Yeah, try that out and we'll see what it looks like. Thanks!

src/type.js
```diff
@@ -220,17 +225,28 @@ async function typeImpl(
...
-            data: key,
-            inputType: 'insertText',
-            ...eventOverrides,
-          },
-        })
-        prevValue = inputEvent.prevValue
+        if (
+          !(currentElement().type === 'date') ||
+          isValidDateValue(currentElement(), textToBeTyped)
+        ) {
```
lourenci, line 235: [`date` input doesn't trigger an `input` event when a date is not valid.](https://codesandbox.io/s/user-event-playground-91jlf).

My first move was doing this conditional inside the `fireInputEventIfNeeded`, but the conditional there is more general, so I keep it here.

src/type.js
```diff
@@ -220,17 +225,28 @@ async function typeImpl(
...
+              inputType: 'insertText',
+              ...eventOverrides,
+            },
+          })
+          prevValue = inputEvent.prevValue
+        }
+
+        if (isValidDateValue(currentElement(), textToBeTyped)) {
+          fireEvent.change(currentElement(), {target: {value: textToBeTyped}})
+        }
```
lourenci, line 249: [`date` input triggers a `change` event together with `input` event when a date is valid.](https://codesandbox.io/s/user-event-playground-91jlf)

src/type.js
```diff
@@ -192,12 +195,14 @@ async function typeImpl(
...
       prevWasPeriod = false,
       prevValue = '',
+      typedValue = '',
       eventOverrides,
     },
   ) {
     const key = char // TODO: check if this also valid for characters with diacritic markers e.g. úé etc
     const keyCode = char.charCodeAt(0)
     let nextPrevWasMinus, nextPrevWasPeriod
+    const textToBeTyped = typedValue + char
```
lourenci, line 205: We need to keep some var around to concat the chars since the `date` input only accepts valid dates for its value.

src/__tests__/type.js
```diff
@@ -600,6 +600,63 @@ test('can type into an input with type `email`', () => {
...
+    input[value=""] - keydown: - (45)
+    input[value=""] - keypress: - (45)
+    input[value=""] - keyup: - (45)
+    input[value=""] - keydown: 2 (50)
+    input[value=""] - keypress: 2 (50)
+    input[value=""] - keyup: 2 (50)
+    input[value=""] - keydown: 9 (57)
+    input[value=""] - keypress: 9 (57)
+    input[value="2020-06-29"] - input
+      "{CURSOR}" -> "{CURSOR}2020-06-29"
```
kentcdodds, line 654: Is this the right spot for the cursor? I would expect it at the end.

lourenci: I hadn't think about that but using it, it seems to keep the last typed part selected. Maybe we could force the selection of the day (the last part of `type('2020-06-29'`))? 

ezgif com-video-to-gif

kentcdodds: This is where things get complicated... For you the date input accepts `dd/mm/yyyy`, but for me it's `mm/dd/yyyy`. So what the user types is different for everyone. Additionally, here's what happens when I type `10181988`  in the date field in [the playground](https://codesandbox.io/s/user-event-playground-eo909):

```
input#date[value=""] - pointerdown
input#date[value=""] - mousedown
input#date[value=""] - focusin
input#date[value=""] - pointerup
input#date[value=""] - mouseup
input#date[value=""] - click
input#date[value=""] - keydown
input#date[value=""] - keypress
input#date[value=""] - keyup
input#date[value=""] - keydown
input#date[value=""] - keypress
input#date[value=""] - keyup
input#date[value=""] - keydown
input#date[value=""] - keypress
input#date[value=""] - keyup
input#date[value=""] - keydown
input#date[value=""] - keypress
input#date[value=""] - keyup
input#date[value=""] - keydown
input#date[value=""] - keypress
input#date[value="0001-10-18"] - input
input#date[value="0001-10-18"] - change
input#date[value="0001-10-18"] - keyup
input#date[value="0001-10-18"] - keydown
input#date[value="0001-10-18"] - keypress
input#date[value="0019-10-18"] - input
input#date[value="0019-10-18"] - change
input#date[value="0019-10-18"] - keyup
input#date[value="0019-10-18"] - keydown
input#date[value="0019-10-18"] - keypress
input#date[value="0198-10-18"] - input
input#date[value="0198-10-18"] - change
input#date[value="0198-10-18"] - keyup
input#date[value="0198-10-18"] - keydown
input#date[value="0198-10-18"] - keypress
input#date[value="1988-10-18"] - input
input#date[value="1988-10-18"] - change
input#date[value="1988-10-18"] - keyup
input#date[value="1988-10-18"] - focusout
```

Notice that the `input` and `change` events are fired as soon as I get to the year input.

So what the user types depends on the locale, which we can retrieve if we're given a date object: `date.toLocaleDateString()`.

So what if we accept a `Date` object instead, and then type whatever we get from `date.toLocaleDateString()`? We'll want to make sure to handle all these input/change events too I think.

lourenci: I don't know whether I got it.

[`input#value` (in date types) only accepts `yyyy-mm-dd` format](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#Value). I think this might add complexity since we have to format to `yyyy-mm-dd` anyway.

lourenci: As for the `input` and `change` getting fired as soon as we got a valid date, I don't know how to handle that (yet) neither if we need to handle that because we don't know what is the last part of the date to be typed.

If you type last the day (you have typed the year and the month already), the browser will fire the events as soon as you type the day.
If we have a date in format `yyyy-mm-dd` (2020-06-29). We already have typed the year and month, as soon as we type `2`  (from 29):
```
input#date[value=""] - keydown
input#date[value=""] - keypress
input#date[value="2020-06-02"] - input
input#date[value="2020-06-02"] - change
input#date[value="2020-06-02"] - keyup
input#date[value="2020-06-02"] - keydown
input#date[value="2020-06-02"] - keypress
input#date[value="2020-06-29"] - input
input#date[value="2020-06-29"] - change
input#date[value="2020-06-29"] - keyup
```

What do you think?

kentcdodds: Even though the browser gives the value of an input as `yyyy-mm-dd` format, the *user* will type it in differently and that's the point I was trying to make. We should type it in the same way the user would. Which is why accepting a date and then formatting it with `toLocaleDateString` to know how to type it would probably be best.

Another thing to investigate is how Cypress handles this use case.

marcosvega91: I think that cypress is supporting only `yyyy-mm-dd`

lourenci: [Cypress only uses `yyyy-mm-dd` format](https://docs.cypress.io/api/commands/type.html#Date-Inputs).

kentcdodds: That's fine, but what does Cypress do when typing into the input?

I just tested it out with:

```javascript
describe('temp', () => {
  it('fills in dates', () => {
    cy.visit('https://eo909.csb.app/')

    // Comparing what it's like with a regular field first.
    // Using a delay so it's more obvious what the difference is
    cy.findByLabelText(/email/i).type('hello@gmail.com', {delay: 100})

    cy.findByLabelText(/date input/i).type('2020-06-28', {delay: 100})
  })
})
```

Screen Recording 2020-07-23 at 1 45 16 PM

It looks like they do basically the same thing and don't bother "typing" and instead just set the value. This is surprising to me, but I guess I'm fine with following their lead here.

Unfortunately, for some reason, when running the playground in cypress, it doesn't appear to log events, so I'm not sure what events are being run when cypress conducts its interactions. But I'm guessing we'll want to fire the ones that the user would fire if they typed the value in one character at a time with the exception of change/input events I guess 🤷‍♂️

lourenci: > It looks like they do basically the same thing and don't bother "typing" and instead just set the value. This is surprising to me, but I guess I'm fine with following their lead here.

Are you talking about the `value` format or the events triggered by Cypress?

I think this PR is near to what happens in a browser when the user types into it than Cypress, but we can change the PR if you want.

Thank you.

kentcdodds: Yeah, I'm thinking this PR is about ready.

I just realized that I merged a PR that results in conflicts with this one. I'm sorry about that. Do you mind fixing those and we can get this merged. Thanks!

lourenci: [I had to move some code around to avoid the `complexity` eslint rule](https://github.com/testing-library/user-event/pull/400/commits/1b34e2400cf3beab455ddf2cd8631324d8a56e75).

kentcdodds: If that rule ever forces you to make the code *more* complex or confusing then feel free to disable it 😅

visualjerk:
Sure, will review it as soon as possible 😊

src/type.js
```diff
@@ -240,8 +243,16 @@ function fireInputEventIfNeeded({
...
   eventOverrides,
+  textToBeTyped,
 }) {
   const prevValue = currentElement().value
+
+  if (
+    currentElement().type === 'date' &&
+    !isValidDateValue(currentElement(), textToBeTyped)
+  )
+    return {prevValue}
```
visualjerk, line 254: How about keeping this special case outside of `fireInputEventIfNeeded` and instead add the following at the end of the `calculateNewValue` util?

```js
if (element.type === 'date' && !isValidDateValue(element, newValue)) {
  newValue = value
}
```

This would keep things a bit easier to reason about.

src/type.js
```diff
@@ -297,6 +310,8 @@ function typeCharacter(
         newEntry = `-${char}`
       } else if (prevWasPeriod) {
         newEntry = `${prevValue}.${char}`
+      } else if (isValidDateValue(currentElement(), textToBeTyped)) {
```
visualjerk, line 313: Should this be done whenever we have a valide date value? If so, we should put it in a plain `if` after the prevWas checks.

src/utils.js
```diff
@@ -196,6 +196,14 @@ function eventWrapper(cb) {
   return result
 }
 
+function isValidDateValue(element, value) {
+  if(element.type !== 'date') return false
+
+  const clone = element.cloneNode()
+  clone.value = value
+  return clone.value === value
```
visualjerk, line 204: Nice one 🎉

src/type.js
lourenci: This would not avoid a `change` event to be triggered. We could move the condition to outside, in `typeCharacter` scope, but the linter will throw a `complexity` violation.

>As for the events, just to make it clear: date input just triggers an input event (plus a change event) on a valid date according to the user-event playground

visualjerk: It should actually do the same as the current implementation. We already check for `newValue = currentElement.value()` in `fireInputEventIfNeeded`. So when `calculateNewValue` returns the elements current value, the input event will not be fired.

Regarding the `change` event: right now it looks like we only fire it inside `typeCharacter` if we have a valid date on a date input. This behavior would not be affected.

lourenci: My bad, I typo the events, I meant the `input` event: it should be avoided when a date is not valid.

Oh, I overlooked that, I think it might work If we change the `calculateNewValue` function to receive the new value as you have said, though I'm now sure about this move since `calculateNewValue` is more concerning about the new value in a "selection aware". What do you think about it, @kentcdodds?

Thanks

kentcdodds: I'm ok with the suggested change

lourenci: Done. I've to agree, it's better. :)

PR #403, opened 2020-07-18
Feature: Implement cursor navigation with navigation keys

juanca:
Note: WIP! I am running into a few problems and would like advice.

**What**:

Implements [navigation keys](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values#Navigation_keys) support in `type` method:

- Arrows
- Home and end
- Page up and down

**Why**:

It is reasonable for a user to use keyboard navigation while interacting with an interactive element. An interactive element could be an input (where a user would type into) or an accessible UI with keyboard support (no typing but it handles key events).

**How**:

- Handles new key events while typing: e.g. `Brek{arrowleft}a{arrowright}fast` -> `Breakfast`
- Handles new key events with modifier while typing: e.g. `Breakfast{shift}{arrowleft}{/shift}` -> `Breakfas{selection}t{/selection}`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Typings
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -715,3 +715,77 @@ test('should give error if we are trying to call type on an invalid element', as
     `"the current element is of type BODY and doesn't have a valid value"`,
   )
 })
+
+test('navigation key: {arrowleft} and {arrowright} moves the cursor', () => {
+  const {element, getEventSnapshot} = setup('<input />')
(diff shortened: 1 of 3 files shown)


juanca:
I meant to open a draft. Closing this and making a new one.

PR #404, opened 2020-07-18
Feature: Implement cursor navigation with navigation keys

juanca:
Note: WIP! I am running into a few problems and would like advice.

**What**:

Implements [navigation keys](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values#Navigation_keys) support in `type` method:

- Arrows
- Home and end
- Page up and down

**Why**:

It is reasonable for a user to use keyboard navigation while interacting with an interactive element. An interactive element could be an input (where a user would type into) or an accessible UI with keyboard support (no typing but it handles key events).

**How**:

- Handles new key events while typing: e.g. `Brek{arrowleft}a{arrowright}fast` -> `Breakfast`
- Handles new key events with modifier while typing: e.g. `Breakfast{shift}{arrowleft}{/shift}` -> `Breakfas{selection}t{/selection}`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Typings
- [ ] Ready to be merged
      

Closes: https://github.com/testing-library/user-event/issues/354

### DIFF
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -1025,9 +1025,9 @@ test('should not type inside a contenteditable=false div', () => {
 
 test('navigation key: {arrowleft} and {arrowright} moves the cursor', () => {
   const {element, getEventSnapshot} = setup('<input />')
-  userEvent.type(element, 'Brea{arrowleft}k{arrowright}fast')
+  userEvent.type(element, 'b{arrowleft}a{arrowright}c')
   expect(getEventSnapshot()).toMatchInlineSnapshot(`
-    Events fired on: input[value="Brekafast"]
+    Events fired on: input[value="abc"]
 
     input[value=""] - pointerover
     input[value=""] - pointerenter
@@ -1042,57 +1042,27 @@ test('navigation key: {arrowleft} and {arrowright} moves the cursor', () => {
     input[value=""] - pointerup
     input[value=""] - mouseup: Left (0)
     input[value=""] - click: Left (0)
-    input[value=""] - keydown: B (66)
-    input[value=""] - keypress: B (66)
-    input[value="B"] - input
-      "{CURSOR}" -> "B{CURSOR}"
-    input[value="B"] - keyup: B (66)
-    input[value="B"] - keydown: r (114)
-    input[value="B"] - keypress: r (114)
-    input[value="Br"] - input
-      "B{CURSOR}" -> "Br{CURSOR}"
-    input[value="Br"] - keyup: r (114)
-    input[value="Br"] - keydown: e (101)
-    input[value="Br"] - keypress: e (101)
-    input[value="Bre"] - input
-      "Br{CURSOR}" -> "Bre{CURSOR}"
-    input[value="Bre"] - keyup: e (101)
-    input[value="Bre"] - keydown: a (97)
-    input[value="Bre"] - keypress: a (97)
-    input[value="Brea"] - input
-      "Bre{CURSOR}" -> "Brea{CURSOR}"
-    input[value="Brea"] - keyup: a (97)
-    input[value="Brea"] - keydown: ArrowLeft (37)
-    input[value="Brea"] - select
-    input[value="Brea"] - keyup: ArrowLeft (37)
-    input[value="Brea"] - keydown: k (107)
-    input[value="Brea"] - keypress: k (107)
-    input[value="Breka"] - input
-      "Bre{CURSOR}a" -> "Breka{CURSOR}"
-    input[value="Breka"] - select
-    input[value="Breka"] - keyup: k (107)
-    input[value="Breka"] - keydown: ArrowRight (39)
-    input[value="Breka"] - select
-    input[value="Breka"] - keyup: ArrowRight (39)
-    input[value="Breka"] - keydown: f (102)
-    input[value="Breka"] - keypress: f (102)
-    input[value="Brekaf"] - input
-      "Breka{CURSOR}" -> "Brekaf{CURSOR}"
-    input[value="Brekaf"] - keyup: f (102)
-    input[value="Brekaf"] - keydown: a (97)
-    input[value="Brekaf"] - keypress: a (97)
-    input[value="Brekafa"] - input
-      "Brekaf{CURSOR}" -> "Brekafa{CURSOR}"
-    input[value="Brekafa"] - keyup: a (97)
-    input[value="Brekafa"] - keydown: s (115)
-    input[value="Brekafa"] - keypress: s (115)
-    input[value="Brekafas"] - input
-      "Brekafa{CURSOR}" -> "Brekafas{CURSOR}"
-    input[value="Brekafas"] - keyup: s (115)
-    input[value="Brekafas"] - keydown: t (116)
-    input[value="Brekafas"] - keypress: t (116)
-    input[value="Brekafast"] - input
-      "Brekafas{CURSOR}" -> "Brekafast{CURSOR}"
-    input[value="Brekafast"] - keyup: t (116)
+    input[value=""] - keydown: b (98)
+    input[value=""] - keypress: b (98)
+    input[value="b"] - input
+      "{CURSOR}" -> "b{CURSOR}"
+    input[value="b"] - keyup: b (98)
+    input[value="b"] - keydown: ArrowLeft (37)
+    input[value="b"] - select
+    input[value="b"] - keyup: ArrowLeft (37)
+    input[value="b"] - keydown: a (97)
+    input[value="b"] - keypress: a (97)
+    input[value="ab"] - input
+      "{CURSOR}b" -> "ab{CURSOR}"
+    input[value="ab"] - select
+    input[value="ab"] - keyup: a (97)
+    input[value="ab"] - keydown: ArrowRight (39)
+    input[value="ab"] - select
+    input[value="ab"] - keyup: ArrowRight (39)
+    input[value="ab"] - keydown: c (99)
+    input[value="ab"] - keypress: c (99)
+    input[value="abc"] - input
+      "ab{CURSOR}" -> "abc{CURSOR}"
+    input[value="abc"] - keyup: c (99)
   `)
 })
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -1022,3 +1022,77 @@ test('should not type inside a contenteditable=false div', () => {
     div - click: Left (0)
   `)
 })
+
+test('navigation key: {arrowleft} and {arrowright} moves the cursor', () => {
+  const {element, getEventSnapshot} = setup('<input />')
+  userEvent.type(element, 'Brea{arrowleft}k{arrowright}fast')
+  expect(getEventSnapshot()).toMatchInlineSnapshot(`
+    Events fired on: input[value="Brekafast"]
+
+    input[value=""] - pointerover
+    input[value=""] - pointerenter
+    input[value=""] - mouseover: Left (0)
+    input[value=""] - mouseenter: Left (0)
+    input[value=""] - pointermove
+    input[value=""] - mousemove: Left (0)
+    input[value=""] - pointerdown
+    input[value=""] - mousedown: Left (0)
+    input[value=""] - focus
+    input[value=""] - focusin
+    input[value=""] - pointerup
+    input[value=""] - mouseup: Left (0)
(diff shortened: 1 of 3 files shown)


src/__tests__/type.js
```diff
@@ -715,3 +715,77 @@ test('should give error if we are trying to call type on an invalid element', as
...
+    input[value="Brea"] - input
+      "Bre{CURSOR}" -> "Brea{CURSOR}"
+    input[value="Brea"] - keyup: a (97)
+    input[value="Brea"] - keydown: ArrowLeft (37)
+    input[value="Brea"] - select
+    input[value="Brea"] - keyup: ArrowLeft (37)
+    input[value="Brea"] - keydown: k (107)
+    input[value="Brea"] - keypress: k (107)
+    input[value="Breka"] - input
+      "Bre{CURSOR}a" -> "Breka{CURSOR}"
```
juanca, line 764: I tried looking into this but got a bit lost. It looks like the event snapshot has a bug where the `{CURSOR}` is always at the end. I think to get an accurate representation of the navigation keys in our tests then the event snapshot should move the cursor around as well. 

Any advice or tips on getting this fixed?

src/keys/navigation-key.js
```diff
@@ -0,0 +1,58 @@
+import {fireEvent} from '@testing-library/dom'
```
juanca, line 1: What are your opinions on using this kind of code organization? I noticed there is 3 files for typing:

1. utils
1. type
1. type-modifiers

I think navigation keys will end up adding a chunk (but not massive) amount of code. I opted for a new file but I am willing to move things around.

src/__tests__/type.js
juanca: I think I'm figuring out how the test utils work. 

After learning a bit more, I think I'm going to implement a change for arrows (?):

```
    input[value="Brea"] - keydown: ArrowLeft (37)
    input[value="Brea"] - select
      "Brea{CURSOR}" -> "Bre{CURSOR}a"
    input[value="Brea"] - keyup: ArrowLeft (37)
```

Though, I'm not sure if it should be on the select event or the keyup.

kentcdodds: I'm not sure. Based on [the playground](https://codesandbox.io/s/user-event-playground-eo909?file=/src/index.js) we shouldn't fire a `select` event at all I think (though it may be unfortunately unavoidable).

src/keys/navigation-key.js
kentcdodds: I probably wouldn't have bothered making the new file myself, but I'm ok with it.

src/__tests__/type.js
juanca: Looks like it is a bug in jsdom's `HTMLInputElement-impl.js`:

```
set value(val) {
    switch (valueAttributeMode(this.type)) {
      // https://html.spec.whatwg.org/multipage/input.html#dom-input-value-value
      case "value": {
        const oldValue = this._value;
        this._value = sanitizeValueByType(this, val);
        this._dirtyValue = true;

        if (oldValue !== this._value) {
          this._selectionStart = this._selectionEnd = this._getValueLength();
          this._selectionDirection = "none";
        }
        break;
      }

```

where `val = "Breka"` and `oldValue = "Brea"`

juanca: Okay, nevermind. This is an issue with the HTML standard. See more information @ https://github.com/whatwg/html/issues/5743

I am going to resolve this and ignore the odd cursor position.

src/__tests__/type.js
```diff
@@ -715,3 +715,77 @@ test('should give error if we are trying to call type on an invalid element', as
...
+    input[value="Bre"] - input
+      "Br{CURSOR}" -> "Bre{CURSOR}"
+    input[value="Bre"] - keyup: e (101)
+    input[value="Bre"] - keydown: a (97)
+    input[value="Bre"] - keypress: a (97)
+    input[value="Brea"] - input
+      "Bre{CURSOR}" -> "Brea{CURSOR}"
+    input[value="Brea"] - keyup: a (97)
+    input[value="Brea"] - keydown: ArrowLeft (37)
+    input[value="Brea"] - select
```
juanca, line 759: I am thinking a select event should show the before and after of the cursor position:

```
    input[value="Brea"] - keydown: ArrowLeft (37)
    input[value="Brea"] - select
      "Brea{CURSOR}" -> "Bre{CURSOR}a"
    input[value="Brea"] - keyup: ArrowLeft (37)
```

What do y'all think?

juanca: For posterity's sake: there was an earlier comment and it got [1 response](https://github.com/testing-library/user-event/pull/404#discussion_r456944718). 

> I'm not sure. Based on the playground we shouldn't fire a select event at all I think (though it may be unfortunately unavoidable).

I made a new comment thread to leave the unexpected `{CURSOR}` position bits out of this.

kentcdodds: I agree

CWSites:
Any update here? We could really use the ability to use type the arrow key for a couple reasons.
1. Test/verify accessibility issues
2. Get around complex components such as Material UI autocomplete by using our arrow key to select an option since we are unable to pick a specific option any other way.

DovRine:
Agree with CWSites. This would greatly simplify testing things like: MaterialUI TextField[type="time"].

kentcdodds:
I'm sure @juanca would be more than happy to have either of you pick up where they left off and address the merge conflicts and other issues in this PR :)

jesujcastillom:
Hi! This is something I recently came across with, and I'd be happy to help if it needs picking up 😃

kentcdodds:
Go for it 👍

juanca:
@jesujcastillom Thanks for the offer! I would greatly appreciate the help. I've been nerd-sniped into other projects for my free time. 

Let me know if anyone needs any context, happy to provide it whenever.

src/__tests__/type.js
```diff
@@ -1022,3 +1023,83 @@ test('should not type inside a contenteditable=false div', () => {
...
+      "Brekafas{CURSOR}" -> "Brekafast{CURSOR}"
+    input[value="Brekafast"] - keyup: t (116)
+  `)
+})
+
+test('navigationKey throws if an invalid key is passed', () => {
+  expect(() => {
+    navigationKey('SOME_INVALID_KEY')
+  }).toThrowError(TypeError)
+})
```
jesujcastillom, line 1105: I've modified the function to hit all the lines (`default` wasn't reachable) now the question is, should I include this test to make sure we throw a `TypeError` (and also throw it in the function)` if we don't get a valid key, or leave it without throwing, since it's an internal function anyway.

In summary, removing the last commit entirely, we still get 💯 coverage, but it's not really covered if an invalid key is passed, any thoughts? @juanca @kentcdodds

jesujcastillom: Note, modified the commit because I didn't have my git config setup properly in this  machine, but yeah, same changes 😅

src/type.js
```diff
@@ -44,6 +45,8 @@ const modifierCallbackMap = {
 }
 
 const specialCharCallbackMap = {
+  '{arrowleft}': navigationKey('ArrowLeft'),
+  '{arrowright}': navigationKey( 'ArrowRight'),
```
kentcdodds, line 49:
```suggestion
  '{arrowright}': navigationKey('ArrowRight'),
```

src/__tests__/type.js
```diff
@@ -1022,3 +1023,83 @@ test('should not type inside a contenteditable=false div', () => {
...
   `)
 })
+
+test('navigation key: {arrowleft} and {arrowright} moves the cursor', () => {
+  const {element, getEventSnapshot} = setup('<input />')
+  userEvent.type(element, 'Brea{arrowleft}k{arrowright}fast')
+  expect(getEventSnapshot()).toMatchInlineSnapshot(`
+    Events fired on: input[value="Brekafast"]
+
+    input[value=""] - pointerover
```
kentcdodds, line 1033: I just tried this in [the playground](https://codesandbox.io/s/user-event-playground-eo909) and here's what I got for the events (Note, I ignored mouse/pointer move events because they're too noisy in the playground):

```
textarea#textarea[value=""] - pointerdown
textarea#textarea[value=""] - mousedown
textarea#textarea[value=""] - focusin
textarea#textarea[value=""] - pointerup
textarea#textarea[value=""] - mouseup
textarea#textarea[value=""] - click
textarea#textarea[value=""] - keydown
textarea#textarea[value=""] - keypress
textarea#textarea[value="b"] - input
textarea#textarea[value="b"] - keyup
textarea#textarea[value="b"] - keydown
textarea#textarea[value="b"] - keypress
textarea#textarea[value="br"] - input
textarea#textarea[value="br"] - keyup
textarea#textarea[value="br"] - keydown
textarea#textarea[value="br"] - keypress
textarea#textarea[value="bre"] - input
textarea#textarea[value="bre"] - keyup
textarea#textarea[value="bre"] - keydown
textarea#textarea[value="bre"] - keypress
textarea#textarea[value="brea"] - input
textarea#textarea[value="brea"] - keyup
textarea#textarea[value="brea"] - keydown // <- here's the arrow down to go left
textarea#textarea[value="brea"] - keyup
textarea#textarea[value="brea"] - keydown
textarea#textarea[value="brea"] - keypress
textarea#textarea[value="breka"] - input
textarea#textarea[value="breka"] - keyup
textarea#textarea[value="breka"] - keydown // <- here's the arrow down to go right
textarea#textarea[value="breka"] - keyup
textarea#textarea[value="breka"] - keydown
textarea#textarea[value="breka"] - keypress
textarea#textarea[value="brekaf"] - input
textarea#textarea[value="brekaf"] - keyup
textarea#textarea[value="brekaf"] - keydown
textarea#textarea[value="brekaf"] - keypress
textarea#textarea[value="brekafa"] - input
textarea#textarea[value="brekafa"] - keyup
textarea#textarea[value="brekafa"] - keydown
textarea#textarea[value="brekafa"] - keypress
textarea#textarea[value="brekafas"] - input
textarea#textarea[value="brekafas"] - keyup
textarea#textarea[value="brekafas"] - keydown
textarea#textarea[value="brekafas"] - keypress
textarea#textarea[value="brekafast"] - input
textarea#textarea[value="brekafast"] - keyup
```

Notice that there's no "select" event in there. I'm not sure whether it's possible to avoid that event, but it would be good to avoid if we can.

Also, I think this test could be simplified if we change the input to: `b{arrowleft}a{arrowright}c` (which would result in `abc`, and much fewer events for the snapshot).

src/__tests__/type.js
kentcdodds: Is this even possible? We're the only ones who call `navigationKey` and we always call it properly. I'd just remove that logic.

kentcdodds (commented):
Thanks for picking this up! Here's some feedback for you.

src/__tests__/type.js
jesujcastillom: No, it isn't I'll remove it :)

src/__tests__/type.js
jesujcastillom: At least from what I could investigate, the [setSelectionRange](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-textarea/input-setselectionrange] will fire a `select` event, and I can't really seem to find another way to move the cursor.

src/type.js
jesujcastillom: Done!

kentcdodds (approved):
Looks good to be, but I don't think I can merge this until @juanca removes the draft status of this pr.

kentcdodds:
@All-contributors please add @juanca for code and tests

@All-contributors please add @jesujcastillom  for code and tests

kentcdodds:
@all-contributors please add @juanca for code and tests

kentcdodds:
@all-contributors please add @jesujcastillom for code and tests

brandonpittman:
Forgive my ignorance, but why only implement arrowright and arrowleft? Why not include up and down?

kentcdodds:
Up and down would be extremely difficult 😬

brandonpittman:
@kentcdodds What makes up and down harder than left and right?

kentcdodds:
It comes down to where the cursor goes. In a textarea it depends on the size of the textarea as well as the font size. This would take a ridiculous amount of work (would probably require an environment that supports layout, so jsdom is out). Just too much work in afraid.

Issue #406, opened 2020-07-21
Native button should trigger click on keydown space


visualjerk:
For of all, thank you for this handy library 😄 

- `@testing-library/user-event` version: 12.0.11
- Testing Framework and version: jest - 26.1.0
  
- DOM Environment: react-dom - 16.13.1
  

Relevant code or config

```javascript
import React from "react";
import { render, screen } from "@testing-library/react";
import userEvent from "@testing-library/user-event";

describe("test", () => {
  it("emits click on button type space", async () => {
    const handleClick = jest.fn();
    render(<button onClick={handleClick}>foo</button>);
    screen.getByText("foo").focus();
    await userEvent.type(screen.getByText("foo"), " ", { skipClick: true });
    expect(handleClick).toHaveBeenCalledTimes(1);
  });
});

```

Reproduction repository:

https://codesandbox.io/s/angry-bas-bttix?file=/src/user-event.test.js

Problem description:

A native button triggers a click event on keydown space.
Currently only keydown enter is handled properly by `userEvent.type`.

Suggested solution:

Add the same behavior that keydown enter has for keydown space.

kentcdodds:
That makes sense to me, though I think having a modifier (like `{space}`) would be useful. Would you be interested in adding support for that?

PeteJodo:
just to add to this convo, I have a test right now with an element that has `role` set to `button` and was trying to test that pressing `space` triggers the expected behavior but I'm hitting [this error](https://github.com/testing-library/user-event/blob/master/src/type.js#L60-L64) in the code. I wonder if by introducing a `{space}` modifier, we can avoid that branch in the code somehow.

There's nothing I can find in the ARIA `button` role documentation that requires an ARIA-equivalent attribute to `value` being set so I feel like what I'm doing is a legitimate use case

marcosvega91:
what you're saying is regarding the other issue opened by @visualjerk https://github.com/testing-library/user-event/issues/407

visualjerk:
> That makes sense to me, though I think having a modifier (like `{space}`) would be useful. Would you be interested in adding support for that?

I would be glad to add support for it. Just to be clear:

I would expect that `userEvent.type(el, '{space}')` = `userEvent.type(el, ' ')`

marcosvega91:
I think that the behaviour should be different because `{space}` on a button should press the button if the button has the focus while `' '` should add a value to the button equal to ` `

kentcdodds:
[...]

PR #409, opened 2020-07-22
fix: button trigger space

visualjerk:
**What**:

This fixes https://github.com/testing-library/user-event/issues/406

**Why**:

Because clickable inputs and buttons should be handled differently by `type`.

**How**:

- Added clickable input checks in `type` where it was necessary
- Added support for a special character {space}
- Refactored `type.js` to keep it understandable with these additions

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Typings (not needed)
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/type-modifiers.js
+++ b/src/__tests__/type-modifiers.js
@@ -443,6 +443,92 @@ test('{enter} on a button', () => {
   `)
 })
 
+test('{space} on a button', () => {
+  const {element, getEventSnapshot} = setup('<button />')
+
+  userEvent.type(element, '{space}')
+
+  expect(getEventSnapshot()).toMatchInlineSnapshot(`
+    Events fired on: button
+
+    button - pointerover
+    button - pointerenter
+    button - mouseover: Left (0)
+    button - mouseenter: Left (0)
+    button - pointermove
+    button - mousemove: Left (0)
+    button - pointerdown
+    button - mousedown: Left (0)
+    button - focus
+    button - focusin
+    button - pointerup
+    button - mouseup: Left (0)
+    button - click: Left (0)
+    button - keydown: (32)
+    button - keypress: (32)
+    button - keyup: (32)
+    button - click: Left (0)
+  `)
+})
+
+test(`' ' on a button`, () => {
+  const {element, getEventSnapshot} = setup('<button />')
+
+  userEvent.type(element, ' ')
+
+  expect(getEventSnapshot()).toMatchInlineSnapshot(`
+    Events fired on: button
+
+    button - pointerover
+    button - pointerenter
+    button - mouseover: Left (0)
+    button - mouseenter: Left (0)
+    button - pointermove
+    button - mousemove: Left (0)
+    button - pointerdown
+    button - mousedown: Left (0)
+    button - focus
+    button - focusin
+    button - pointerup
+    button - mouseup: Left (0)
+    button - click: Left (0)
+    button - keydown: (32)
+    button - keypress: (32)
+    button - keyup: (32)
+    button - click: Left (0)
+  `)
+})
+
+test('{space} on an input', () => {
+  const {element, getEventSnapshot} = setup(`<input value="" />`)
+
+  userEvent.type(element, '{space}')
+
+  expect(getEventSnapshot()).toMatchInlineSnapshot(`
+    Events fired on: input[value=" "]
+
+    input[value=""] - pointerover
+    input[value=""] - pointerenter
+    input[value=""] - mouseover: Left (0)
+    input[value=""] - mouseenter: Left (0)
+    input[value=""] - pointermove
+    input[value=""] - mousemove: Left (0)
+    input[value=""] - pointerdown
+    input[value=""] - mousedown: Left (0)
+    input[value=""] - focus
+    input[value=""] - focusin
+    input[value=""] - pointerup
+    input[value=""] - mouseup: Left (0)
+    input[value=""] - click: Left (0)
+    input[value=""] - keydown: (32)
+    input[value=""] - keypress: (32)
+    input[value=" "] - input
+      "{CURSOR}" -> " {CURSOR}"
+    input[value=" "] - keyup: (32)
+  `)
+})
+
+
 test('{enter} on a textarea', () => {
   const {element, getEventSnapshot} = setup('<textarea></textarea>')
 
--- a/src/__tests__/type-modifiers.js
+++ b/src/__tests__/type-modifiers.js
@@ -528,6 +528,95 @@ test('{space} on an input', () => {
   `)
 })
 
+test('{enter} on an input type="color"', () => {
+  const {element, getEventSnapshot} = setup(
+    `<input value="#ffffff" type="color" />`,
+  )
+
+  userEvent.type(element, '{enter}')
+
+  expect(getEventSnapshot()).toMatchInlineSnapshot(`
+    Events fired on: input[value="#ffffff"]
+
+    input[value="#ffffff"] - pointerover
+    input[value="#ffffff"] - pointerenter
+    input[value="#ffffff"] - mouseover: Left (0)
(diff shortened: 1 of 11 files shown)


visualjerk:
Unfortunately the CI fails 😄 

During implementation I needed to update the focus and blur behavior, because of a recent fix in jsdom: https://github.com/jsdom/jsdom/issues/2708

I also updated `@testing-library/dom` and `@testing-library/jest-dom` versions to make sure the new jsdom in the CI.

However it looks like Travis is still loading a cached version. Maybe the new jsdom version is not defined in jest-dom.

Do you know how to tell Travis to rerun with a cleared cache?

kentcdodds:
Awesome! Just triggered a run of CI without the cache. This is a big change! I'll try to review it now.

src/__tests__/type-modifiers.js
```diff
@@ -443,6 +443,181 @@ test('{enter} on a button', () => {
...
+    button - mouseenter: Left (0)
+    button - pointermove
+    button - mousemove: Left (0)
+    button - pointerdown
+    button - mousedown: Left (0)
+    button - focus
+    button - focusin
+    button - pointerup
+    button - mouseup: Left (0)
+    button - click: Left (0)
```
kentcdodds, line 466: I'm thinking this is probably a special case where simulating mousing over and clicking wouldn't make sense right? Because anyone trying to do: `userEvent.type(button, '{space}')` is probably *not* wanting to fire *two* click events... Hmmm... Should we make this a special case?

src/__tests__/type-modifiers.js
```diff
@@ -443,6 +443,181 @@ test('{enter} on a button', () => {
...
+    button - keydown: (32)
+    button - keypress: (32)
+    button - keyup: (32)
+    button - click: Left (0)
+  `)
```
kentcdodds, line 499: Could we make it clear that all we're really testing is that `' '` is the same as `'{space}'`?

```suggestion
test(`' ' on a button is the same as '{space}'`, () => {
  const {element: e1, getEventSnapshot: snap1} = setup('<button />')
  const {element: e2, getEventSnapshot: span2} = setup('<button />')

  userEvent.type(e1, ' ')
  userEvent.type(e2, '{space}')

  expect(snap1()).toBe(snap2())
```

src/__tests__/type-modifiers.js
```diff
@@ -443,6 +443,181 @@ test('{enter} on a button', () => {
...
+    input[value="#ffffff"] - mouseup: Left (0)
+    input[value="#ffffff"] - click: Left (0)
+    input[value="#ffffff"] - keydown: (32)
+    input[value="#ffffff"] - keypress: (32)
+    input[value="#ffffff"] - keyup: (32)
+    input[value="#ffffff"] - click: Left (0)
+  `)
+})
+
+test('" " on an input type="color"', () => {
```
kentcdodds, line 591: Let's do the same to this one as we do with the button one above.

src/__tests__/type-modifiers.js
```diff
@@ -443,6 +443,181 @@ test('{enter} on a button', () => {
...
+    input[value="#ffffff"] - mouseup: Left (0)
+    input[value="#ffffff"] - click: Left (0)
+    input[value="#ffffff"] - keydown: Enter (13)
+    input[value="#ffffff"] - keypress: Enter (13)
+    input[value="#ffffff"] - click: Left (0)
+    input[value="#ffffff"] - keyup: Enter (13)
+  `)
+})
+
+test('{space} on an input type="color"', () => {
```
kentcdodds, line 561: Could we make it clear (probably via a comment or the test title) what about this test is important? It's hard to tell as it looks just like all the other tests. Is there some special reason we want to test the color type? (I just tried out an input[type=color] and realized in chrome it renders a button with a color picker so that makes sense why we've got this test here, but a comment would be helpful).

src/__tests__/type-modifiers.js
```diff
@@ -443,6 +443,181 @@ test('{enter} on a button', () => {
...
+    input[value=""] - click: Left (0)
+    input[value=""] - keydown: (32)
+    input[value=""] - keypress: (32)
+    input[value=" "] - input
+      "{CURSOR}" -> " {CURSOR}"
+    input[value=" "] - keyup: (32)
+  `)
+})
+
+test('{enter} on an input type="color"', () => {
```
kentcdodds, line 531: Same thing as the space on a type color. Why is type color so special it needs a test? When this test fails, how does the developer know what part of the test is important for fixing?

src/type.js
```diff
@@ -167,106 +158,215 @@ async function typeImpl(
...
+  }
+  return {
+    callback: specialCharCallbackMap[specialChar],
+    remainingString: remainingString.slice(specialChar.length),
+  }
+}
+
+function getTypeCallback(remainingString) {
+  const character = remainingString[0]
+  const callback = createTypeCharacter(character)
```
kentcdodds, line 218: I'm not sure we need the createTypeCharacter function. This is a quick and easy one-liner.

```suggestion
  const callback = context => typeCharacter(character, context)
```

src/type.js
```diff
@@ -167,106 +158,215 @@ async function typeImpl(
...
-            nextPrevWasPeriod = prevWasPeriod
-          } else {
-            nextPrevWasPeriod = newEntry === '.'
-          }
-        }
-      }
-    }
+function createTypeCharacter(character) {
+  return context => typeCharacter(character, context)
+}
```
kentcdodds, line 284: With the definition inline above we can remove this.

```suggestion
```

kentcdodds:
Oh, and the CI failures are ESLint related:

```
[lint] 
[lint] /home/travis/build/testing-library/user-event/src/__tests__/click.js
[lint]    82:19  error  Use not.toBeChecked() instead of toHaveProperty('checked', false)  jest-dom/prefer-checked
[lint]   110:19  error  Use toBeChecked() instead of toHaveProperty('checked', true)       jest-dom/prefer-checked
[lint]   128:19  error  Use not.toBeChecked() instead of toHaveProperty('checked', false)  jest-dom/prefer-checked
[lint] 
[lint] ✖ 3 problems (3 errors, 0 warnings)
[lint]   3 errors and 0 warnings potentially fixable with the `--fix` option.
[lint] 
[lint] npm run lint --silent exited with code 1
```

If you could fix those too that'd be super :) Thanks!

kentcdodds:
Fantastic. Thank you!

kentcdodds:
@all-contributors please add @visualjerk for code, tests, and docs

src/__tests__/type-modifiers.js
visualjerk: I was asking myself the same question. For the sake of consistency I decided it should behave the same way it does with other elements.

Maybe we could add a section about this to the documentation? So it is clear that the most popular way to test keyboard events on a button would be:

```js
button.focus()
userEvent.type(button, '{space}', { skipClick: true })
```

Another idea would be to use element.focus() instead of userEvent.click(element) as a default for `type`.

The benefit would be, that it is less "magic" and we could get rid of `skipClick` altogether.
Also the `.focus()` would just fire additional events, if the element doesn't have focus already. So it would just be a minimal safety net.

For tests that require a `click` beforehand, it would still be easy to do that:

```js
userEvent.click(input)
userEvent.type(input, 'foo')
```

kentcdodds:

src/type.js
visualjerk: Sorry, just had time to read this. That's true. Initially I used createTypeCharacter in more places, but now it is a bit too much 😄

visualjerk:
@kentcdodds Thanks for your kind words 😄 

As this is already merged, I can add the improvements you mentioned to the next PR.

kentcdodds:
Oh, whoops 😅 I didn't check and just assumed you'd made the improvements in that one commit. Thanks!

PR #415, opened 2020-07-25
feat: support contenteditable div

marcosvega91:
**What**: I have added support for content editable

**Why**: Because user can't type into it 

**How**: Triggering the same events as textarea. Div does not have a value but a textContent
 
I didn't update `getEventSnapshot` because there is not selector for textContent

**Checklist**:

- [ ] Documentation
- [X] Tests
- [ ] Typings
- [X] Ready to be merged

### DIFF
--- a/src/click.js
+++ b/src/click.js
@@ -2,6 +2,7 @@ import {fireEvent} from '@testing-library/dom'
 import {
   getMouseEventOptions,
   isLabelWithInternallyDisabledControl,
+  isContentEditable,
 } from './utils'
 import {hover} from './hover'
 import {blur} from './blur'
@@ -63,7 +64,10 @@ function clickElement(element, init, {clickCount}) {
     const shouldFocus = element.ownerDocument.activeElement !== element
     if (continueDefaultHandling) {
       if (previousElement) blur(previousElement, init)
-      if (shouldFocus) focus(element, init)
+      if (shouldFocus) {
+        if (element.tagName !== 'DIV' || isContentEditable(element))
+          focus(element, init)
+      }
     }
   }
   fireEvent.pointerUp(element, init)
--- a/src/click.js
+++ b/src/click.js
@@ -2,7 +2,6 @@ import {fireEvent} from '@testing-library/dom'
 import {
   getMouseEventOptions,
   isLabelWithInternallyDisabledControl,
-  isContentEditable,
 } from './utils'
 import {hover} from './hover'
 import {blur} from './blur'
@@ -64,10 +63,7 @@ function clickElement(element, init, {clickCount}) {
     const shouldFocus = element.ownerDocument.activeElement !== element
     if (continueDefaultHandling) {
       if (previousElement) blur(previousElement, init)
-      if (shouldFocus) {
-        if (element.tagName !== 'DIV' || isContentEditable(element))
-          focus(element, init)
-      }
+      if (shouldFocus) focus(element, init)
     }
   }
   fireEvent.pointerUp(element, init)
--- a/src/type.js
+++ b/src/type.js
@@ -10,6 +10,9 @@ import {
   setSelectionRangeIfNecessary,
   isClickable,
   isValidDateValue,
+  getSelectionRange,
+  getValue,
+  isContentEditable,
 } from './utils'
 import {click} from './click'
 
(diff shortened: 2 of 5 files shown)


src/click.js
```diff
@@ -63,7 +64,10 @@ function clickElement(element, init, {clickCount}) {
     const shouldFocus = element.ownerDocument.activeElement !== element
     if (continueDefaultHandling) {
       if (previousElement) blur(previousElement, init)
-      if (shouldFocus) focus(element, init)
+      if (shouldFocus) {
+        if (element.tagName !== 'DIV' || isContentEditable(element))
+          focus(element, init)
+      }
```
kentcdodds, line 70: I don't think we need this change because the `focus` utility will exit early if the given element is not focusable. 

```suggestion
      if (shouldFocus) focus(element, init)
```

src/utils.js
```diff
@@ -94,8 +94,38 @@ function getActiveElement(document) {
   }
 }
 
+function getSelectionRange(element) {
+  if (isContentEditable(element)) {
+    const range = document.getSelection().getRangeAt(0)
+
+    return {selectionStart: range.startOffset, selectionEnd: range.endOffset}
```
kentcdodds, line 101: We should probably also make sure that the document's selection range is within the given element as well.

src/type.js
```diff
@@ -85,6 +88,14 @@ async function typeImpl(
...
   if (!skipClick) click(element)
 
+  if (isContentEditable(element)) {
+    if (document.getSelection().rangeCount === 0) {
+      const range = document.createRange()
+      range.setStart(element, 0)
+      range.setEnd(element, 0)
+      document.getSelection().addRange(range)
+    }
+  }
```
kentcdodds, line 98: Is this not all handled by the `setSelectionRangeIfNecessary` call below?

src/click.js
marcosvega91: I have tried on the div without the content editable but the focus was trigfered. I'll take a look again

marcosvega91: The problem here was the a div is not in the list of focusable elements. I have moved the check here https://github.com/testing-library/user-event/blob/2335842d247a6ae2e761f96f81e99fd1628e0a3f/src/utils.js#L167-L172.

src/type.js
marcosvega91: The problem is that when the div has the focus it's possible that `getSelection` is elsewhere. Maybe we can put this logic in the `getSelectionRange`. If the element of `getSelection` is not the current element we can change the range

src/click.js
kentcdodds: Let's change `isFocusable` to support this use case so we don't have to make this block of code more complex.

src/type.js
kentcdodds: I wouldn't want `getSelectionRange` to perform a side-effect.

How is it possible for the element with focus to not be the one with selection?

marcosvega91: Good question. I need to create some use cases on playground because there are more thing to consider

marcosvega91: `jsdom` behaviour is very strange. 

1) jsdom https://repl.it/repls/DutifulLividBytecode#index.js
2) without jsdom https://jsfiddle.net/41sabm23/1/

In `user-event` focus is working right but after it `getSelection().rangeCount` is 0

kentcdodds: Seems like a jsdom bug you could report (using those exact same reproduction links).

I think we can merge this for now though.

kentcdodds:
Could you resolve the merge conflicts?

marcosvega91:
I have rebased the branch :) 👍

src/utils.js
```diff
@@ -187,7 +230,7 @@ const FOCUSABLE_SELECTOR = [
 function isFocusable(element) {
   return (
     !isLabelWithInternallyDisabledControl(element) &&
-    element?.matches(FOCUSABLE_SELECTOR)
+    (isContentEditable(element) || element?.matches(FOCUSABLE_SELECTOR))
```
kentcdodds, line 233: Actually, could we change this to update the `FOCUSABLE_SELECTOR` variable to include `[contenteditable=""]` and `[contenteditable="true"]`. Then we don't need to make any changes here.

kentcdodds: The benefit of this is that then the `FOCUSABLE_SELECTOR` will really be able to tell us whether something is focusable (with the exception of a label with internally disabled control... I wonder whether that would be possible to handle in the selector as well).

src/utils.js
```diff
@@ -180,6 +223,8 @@ const FOCUSABLE_SELECTOR = [
   'button:not([disabled])',
   'select:not([disabled])',
   'textarea:not([disabled])',
+  'div[contenteditable=""]',
+  'div[contenteditable="true"]',
```
kentcdodds, line 227: Does it have to be a `div`? Couldn't it be a `span` or anything else?

```suggestion
  '[contenteditable=""]',
  '[contenteditable="true"]',
```

marcosvega91: 😓

kentcdodds:
Thanks @marcosvega91! This is a great new feature 👏

PR #427, opened 2020-08-07
feat: 🎸 add ignoreActiveElement option to type

DanielRHayes:
**What**:

* Adds a `ignoreActiveElement` option to `type`. Causing `type` to type into the `element` passed in instead of the current active element.

**Why**:

* I'm using the library in a browser extension to complete forms on the behalf of users. In my case there is a dialog displaying to users which is always the active element.

**How**:

* in `typeImpl` added the option which if set true will cause `currentElement()` to always return the `element` passed

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Typings  (I'm not strong on typescript if someone could check that change)
- [x] Ready to be merged

Thanks for all the work on this library!

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,12 +175,16 @@ are typed. By default it's 0. You can use this option if your component has a
 different behavior for fast or slow users. If you do this, you need to make sure
 to `await`!
 
-> To be clear, `userEvent.type` *always* returns a promise, but you *only* need
-> to `await` the promise it returns if you're using the `delay` option. Otherwise
-> everything runs synchronously and you can ignore the promise.
+> To be clear, `userEvent.type` _always_ returns a promise, but you _only_ need
+> to `await` the promise it returns if you're using the `delay` option.
+> Otherwise everything runs synchronously and you can ignore the promise.
 
 `type` will click the element before typing. To disable this, set the
 `skipClick` option to `true`.
+
+`type` will type into the document's active element. To change this behavior and
+type into the `element` regardless of the active element you can set the
+`ignoreActiveElement` option to `true`.
 
 #### Special characters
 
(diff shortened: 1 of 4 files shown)


kentcdodds:
I'm confused. How can the end-user type if the element isn't the active element? In the browser, the element that receives the key events is always the active element.

DanielRHayes:
True.  It does seem like right now the method works like `type(document, text)` or `type(activeElement, text)` rather than typing into the particular element you pass. It would make more sense to me if the current behavior was scoped to the whole document rather than an element,  more like `type(text)` or `keyboard.type(text)`.

```
import {click, type, keyboard} from 'user-event'

// typing scoped to the element
type(element1, 'some text')

// typing scoped to the document (active element)
click(element2);
keyboard.type('some text') 
```

`type(element1, 'some text')` wouldn't type into `element2` while `keyboard.type('some text')` could type into several inputs (or maybe none).

Overall the question is what  user event is the developer trying to simulate when calling the method?  It would be more clear if there was a separate keyboard for triggering key events to the focused element and having the `type(element, text)` method type into the passed element.

It would be a larger change. I would be willing to write a PR.

kentcdodds:
I think it would be fair to criticize having to pass the `element` as an argument to `type` except part of what `userEvent.type` does is click on and focus the given element (which should make it the `activeElement` if it's focusable, or you could pass a label and the focus would automatically change to the element that it's labeling in which case that would work as well).

You could absolutely create your own utility around userEvent that exposes a `type('some text')` and just expects that you set the active element before calling it (such a utility would either pass `document.activeElement` for the element or pass `document.body` and utilize the `skipClick` option).

> Overall the question is what user event is the developer trying to simulate when calling the method?

The stated goal of user-event is:

> Fire events the same way the user does

So that's what the user event developer is trying to simulate when calling type. And the way the user fires events when typing is: firing keyboard events on the active element.

I hope that helps clear things up :)

PR #428, opened 2020-08-07
fix: improve submit form on ENTER pressed 

marcosvega91:
Follow up #397 

**What**: improve what I have done in the previous PR

**Why**: As suggested by @Jessidhia something was wrong with the current implementation.
A form should be submitted pressing enter in an input only when:
1) A `form` is present
2) The form contains only one input or multiple inputs and a submit button (`input[type="submit"]` or `button[type="submit"]`)

**How**: Updating the current implementation in the modifier

**Checklist**:

- [ ] Documentation
- [X] Tests
- [ ] Typings
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -853,10 +853,9 @@ test('typing on input type file should not result in an error', () => {
   return userEvent.type(element, 'bar').catch(e => expect(e).toBeUndefined())
 })
 
-test('should submit a form with only one input when ENTER is pressed on it', () => {
+test('should submit a form containing multiple text inputs and an input of type submit', () => {
   const handleSubmit = jest.fn()
-  let utils
-  utils = setup(
+  const {element} = setup(
     `
     <form>
       <input type='text'/>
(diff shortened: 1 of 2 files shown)


src/__tests__/type.js
```diff
@@ -853,12 +853,63 @@ test('typing on input type file should not result in an error', () => {
   return userEvent.type(element, 'bar').catch(e => expect(e).toBeUndefined())
 })
 
-test('should submit a form when ENTER is pressed on input', () => {
+test('should submit a form with only one input when ENTER is pressed on it', () => {
```
kentcdodds, line 856: I think it would make sense to split this test up into three different tests. Is there a reason they're combined?

marcosvega91: The first time I have made three tests but because they are related to the same thing I have merged them in a single one

kentcdodds: Often I'm a fan of [fewer, longer tests](https://kentcdodds.com/blog/write-fewer-longer-tests), but in this case I think these make complete sense to be separate tests. Thanks!

marcosvega91: No problem 🙂

PR #438, opened 2020-08-25
Move @testing-library/dom from peerDependencies to dependencies

pustovalov:
**What**:

Make this package more consistent with others, for example:
https://github.com/testing-library/react-testing-library/blob/master/package.json#L47

I installed this package before the documentation was updated, and just noticed this warning:
`warning " > @testing-library/user-event@12.1.1" has unmet peer dependency "@testing-library/dom@>=7.21.4".`

**How**:

`@testing-library/dom` was moved from `peerDependencies` to `dependencies`

**Checklist**:

- [x] Documentation
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,13 +76,13 @@ change the state of the checkbox.
 With NPM:
 
 ```sh
-npm install @testing-library/user-event @testing-library/dom --save-dev
+npm install @testing-library/user-event --save-dev
 ```
 
 With Yarn:
 
 ```sh
-yarn add @testing-library/user-event @testing-library/dom --dev
+yarn add @testing-library/user-event --dev
 ```
 
 Now simply import it in your tests:
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "validate": "kcd-scripts validate"
   },
   "dependencies": {
-    "@babel/runtime": "^7.10.2"
+    "@babel/runtime": "^7.10.2",
+    "@testing-library/dom": ">=7.21.4"
   },
   "devDependencies": {
     "@testing-library/dom": "^7.21.4",
@@ -46,9 +47,6 @@
     "is-ci": "^2.0.0",
     "jest-serializer-ansi": "^1.0.3",
     "kcd-scripts": "^6.2.3"
-  },
-  "peerDependencies": {
-    "@testing-library/dom": ">=7.21.4"
   },
   "eslintConfig": {
     "extends": "./node_modules/kcd-scripts/eslint.js",


Gpx:
Hi @pustovalov, just to clarify, do you only get the warning, or is the library also not working?
The reason why we move to a peer dependency is that RTL is already dependent on testing-library/dom and we risked to import two different versions at the same time.

Since the long-goal for this library is to be included in RTL, if we only have a warning I think we can live with that for now. If on the other hand there's an error we need to look into it.

pustovalov:
Hey @Gpx,

library is working because I have `@testing-library/dom` from `@testing-library/react`

kentcdodds:
The problem is that if you install `@testing-library/react` and `@testing-library/user-event` with this change, it's possible you'd get two different versions of `@testing-library/dom` which is not optimal.

The only issue here is that you have to manually install `@testing-library/react` or `@testing-library/dom` which I don't think is that big of an issue. Additionally, in v7 of the npm CLI, npm will automatically install peerDependencies for you, so not even this won't be a problem in the near future.

Thanks for taking the time (seriously), but we won't be making this change for that reason.

pustovalov:
@kentcdodds WDYT about to move `@testing-library/dom` to peerDependencies for `@testing-library/react`?

kentcdodds:
No, we won't do that. The fact is that it's rare to install `@testing-library/dom` by itself. And it's also rare to install `@testing-library/user-event` without a framework-specific library. So if we moved to `peerDeps` for the framework specific libs, then we'd just increase the package.json listings for everyone for no good reason.

dietergeerts:
> No, we won't do that. The fact is that it's rare to install @testing-library/dom by itself. And it's also rare to install @testing-library/user-event without a framework-specific library. So if we moved to peerDeps for the framework specific libs, then we'd just increase the package.json listings for everyone for no good reason.

There is a good reason though. When you work with pnpm (rush combo in our case), you have problems here, because you get an error on the peer-dependency that is not solvable, and if we add the peerdependency ourselves, we can end up with multiple versions of it. Please see https://github.com/testing-library/user-event/issues/551#issuecomment-930963323

ph-fritsche:
> So the `@testing-library/react` lib here has an actual dependency on `@testing-library/dom`, which makes it unavailable to any other package due to our usage of `pnpm/rush combo`.

@dietergeerts 
`@testing-library/react` not only uses but also configures `@testing-library/dom`.
If we require `@testing-library/dom` the user might get different versions of it for `@testing-library/react` and `@testing-library/user-event`. This is without being warned about it.
He will then run into issues - primarily a bunch of "not wrapped in `act`" warnings - that are non-obvious for most users.

If your environment isolates indirect dependencies, you would have to care for this. So us listing a direct dependency won't solve your problem, but might create issues for people with common environments.

I could think of two solutions (without knowing if your environment allows for this):
1. In post-install link `@testing-library/user-event/node_modules/@testing-library/dom` to `@testing-library/react/node_modules/@testing-library/dom`
2. In pre-install (or as maintenance ci job) fetch the `@testing-library/react` package and copy its dependency version into your direct dependencies so that it always matches.

PR #443, opened 2020-09-05
fix(tab): respect the correct sequence using shift+tab with radio group

### LINKED ISSUE #441: cannot tab back to previuos element when focus is on radiogroup
Since 11.3.2 it is not possible to tab back to the previous element when focus is on a radiogroup.
The focus is then instead switched to the last accessible element.

I wrote a little test scenario where the problem is shown:
https://codesandbox.io/s/distracted-beaver-zwbh2

marcosvega91:
fix #441 

**What**:

Fix a bug when radio group has no checked element and shift+tab is pressed 

**Why**:

This was an issue 

**How**:

Remove lines that reverse the array of elements and check the correct sequence in the loop

**Checklist**:

- [ ] Documentation
- [X] Tests
- [ ] Typings
- [X] Ready to be merged

### DIFF
--- a/src/tab.js
+++ b/src/tab.js
@@ -46,30 +46,58 @@ function tab({shift = false, focusTrap} = {}) {
     })
     .map(({el}) => el)
 
+  const checkedRadio = {}
+  let radioSubgroupContainsFocusedElement = false
+  let currentTakenElementInSubgroup
   // keep only the checked or first element in each radio group
   const prunedElements = []
   for (const el of orderedElements) {
     if (el.type === 'radio' && el.name) {
-      const replacedIndex = prunedElements.findIndex(
-        ({name}) => name === el.name,
-      )
-      if (replacedIndex === -1) {
-        prunedElements.push(el)
-      } else if (el.checked) {
-        prunedElements.splice(replacedIndex, 1)
-        prunedElements.push(el)
-      } else if (
-        shift &&
-        !prunedElements[replacedIndex].checked &&
-        el.ownerDocument.activeElement !== prunedElements[replacedIndex]
+      if (
+        currentTakenElementInSubgroup &&
+        currentTakenElementInSubgroup.name !== el.name
       ) {
-        prunedElements.splice(replacedIndex, 1)
-        prunedElements.push(el)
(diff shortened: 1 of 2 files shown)


src/tab.js
```diff
@@ -46,33 +46,34 @@ function tab({shift = false, focusTrap} = {}) {
     })
     .map(({el}) => el)
 
-  if (shift) orderedElements.reverse()
-
   // keep only the checked or first element in each radio group
```
ph-fritsche, line 49: This premise might be flawed.

```html
<button>Foo</button>
<label><input type='radio' name='myRadio'/>Radio1</label>
<label><input type='radio' name='myRadio'/>Radio2</label>
<button>Bar</button>
<label><input type='radio' name='myRadio'/>Radio3</label>
<button>Baz</button>
```

(Tested in Chrome:)
Until one of the radio buttons is checked, the focus jumps `Foo -> Radio1 -> Bar -> Radio3 -> Baz` and with shift `Baz -> Radio3 ->Bar -> Radio2 -> Foo`

While you cannot focus an unchecked radio button with keyboard/mouse, when another radio of that group is checked, focus can still be set by a script.

```js
document.querySelector('button').addEventListener('click', () => document.querySelector('input').focus())
```

If you check `Radio3`, then click `Foo`, and then press `tab` the focus will be on `Bar`. Focused element should never be pruned, even if another radio button is checked.

marcosvega91: I got you point. Maybe we should make more check on this. For example if the current element in the loop is the one with the check and the previous has the focus we should not delete that element.
We need to consider also a case where there are more radios with different names

marcosvega91: hi @ph-fritsche I have reimplemented that part. There are some checks to make this works. Now the implementation is little tricky, maybe it can be made in a simpler way. Could you please take a look ? Thank you

ph-fritsche: see https://github.com/marcosvega91/user-event/pull/1

marcosvega91: much clearer. Do you want to push your changes here? For me you can eventually close this PR and open a new one

PR #444, opened 2020-09-07
fix(tab): tab order on radio and tabindex=-1

### LINKED ISSUE #441: cannot tab back to previuos element when focus is on radiogroup

ph-fritsche:
fix #441 
replaces #443 

**What**:

`tab()` did not produce the same tab order as the browser does.

**How**:

Reimplemented how the tab sequence is determined.
* Never dismiss activeElement for tab sequence
* Keep DOM order (ignore tabindex) if activeElement has `tabindex="-1"`
* Reimplemented which radio elements to prune from tab order:
   * For radio groups keep only the active radio
   * If there is no active radio, keep only the checked radio

**Checklist**:
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/src/__tests__/tab.js
+++ b/src/__tests__/tab.js
@@ -432,6 +432,71 @@ test('should respect radio groups', () => {
   expect(firstRight).toHaveFocus()
 })
 
+it('should respect the correct sequence when elements in the same radiogroup are not consecutive', () => {
+  setup(`
+    <div>
+      <input
+        data-testid="element"
+        type="radio"
+        name="first"
+        value="radio_left"
+      />
+      <button  data-testid="element">Right Button</button>
+      <input
+        data-testid="element"
+        type="radio"
+        name="first"
+        value="radio_right"
+      />
+    </div>
+  `)
+
+  const [leftRadio, centralButton, rightRadio] = document.querySelectorAll(
+    '[data-testid="element"]',
+  )
+  userEvent.tab()
+  expect(leftRadio).toHaveFocus()
+
+  userEvent.tab()
+  expect(centralButton).toHaveFocus()
+
+  userEvent.tab()
+  expect(rightRadio).toHaveFocus()
+})
+
+it('should respect the correct sequence when non focusable radio has the focus', () => {
+  setup(`
+    <div>
+      <input
+        data-testid="element"
+        type="radio"
+        name="first"
+        value="radio_left"
+      />
+      <input
+        data-testid="element"
+        type="radio"
+        name="first"
+        value="radio_right"
+      />
+      <button  data-testid="element">Right Button</button>
+    </div>
+  `)
+
+  const [leftRadio, rightRadio, rightButton] = document.querySelectorAll(
+    '[data-testid="element"]',
+  )
+
+  userEvent.click(rightRadio)
+
+  expect(rightRadio).toBeChecked()
+
+  focus(leftRadio)
+
+  userEvent.tab()
+  expect(rightButton).toHaveFocus()
+})
+
 it('should respect the correct sequence when radio group is beetwen other elements', () => {
   setup(`
     <div>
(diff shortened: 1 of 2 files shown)


marcosvega91 (changes requested):
Hi @ph-fritsche thanks a lot for this 😄 

Do you wanna work to add some tests to make the coverage 100% ?

marcosvega91 (approved):
thanks a lot for this 🥳  . For me is ready

src/__tests__/tab.js
```diff
@@ -244,6 +244,26 @@ test('should suport a mix of elements with/without tab index', () => {
...
 })
 
+test('ignore tabindex when active element has tabindex="-1"', () => {
+  setup(`
+    <input tabindex='1'/>
+    <input tabindex='0'/>
+    <input tabindex='-1'/>
+    <input tabindex='2'/>
+  `)
+  const [inputA, inputB, inputC, inputD] = document.body.lastChild.children
```
kentcdodds, line 254: Rather than using `document.body.lastChild`, could we get that element from `const {element} = setup('...html...')`?

src/__tests__/tab.js
```diff
@@ -388,51 +408,103 @@ test('should keep focus on the document if there are no enabled, focusable eleme
...
+  const [
+    inputA,
+    radioA,
+    radioB,
+    inputB,
+    radioC,
+    radioD,
+    radioE,
+    inputC,
+  ] = document.body.lastChild.children
```
kentcdodds, line 431: Here as well, let's not rely on `document.body.lastChild` anywhere.

src/__tests__/tab.js
```diff
@@ -388,51 +408,103 @@ test('should keep focus on the document if there are no enabled, focusable eleme
...
+  `)
+  const [
+    inputA,
+    ,
+    inputB,
+    radioB,
+    inputC,
+    ,
+    inputD,
+  ] = document.body.lastChild.children
```
kentcdodds, line 478: Here as well.

src/__tests__/tab.js
```diff
@@ -388,51 +408,103 @@ test('should keep focus on the document if there are no enabled, focusable eleme
...
+  setup(`
+    <input/>
+    <input type="radio" name="radio" checked/>
+    <input/>
+    <input type="radio" name="radio"/>
+    <input/>
+  `)
 
-  userEvent.tab({shift: true})
+  const [, , , radioB, inputC] = document.body.lastChild.children
```
kentcdodds, line 501: Here too.

kentcdodds (commented):
Looks solid. Just one slight comment.

src/__tests__/tab.js
kentcdodds: Noticed the recent push. Don't miss this one :)

ph-fritsche: done.

While we're at it:
I think the extra `<div>` in most tests to comply with the implementation of `setup()` is no improvement.

What do you think about rewriting that `setup()` function as:
```js
function setup(ui, {eventHandlers} = {}) {
  const div = document.createElement('div')
  div.innerHTML = ui.trim()
  document.body.append(div)

  return {
    element: div.firstChild,
    elements: div.children,
    ...addListeners(div, {eventHandlers}),
  }
}
```

kentcdodds: Sure! Let's do that in another PR though. I'll merge this as-is 👍

kentcdodds: Looks like this line hasn't been updated yet.

kentcdodds (approved):
Super!

kentcdodds:
@all-contributors please add @ph-fritsche for code and tests

PR #447, opened 2020-09-08
refactor: test setup

ph-fritsche:
**What**:

The test setup function was written with testing a single element in mind.
Tests working on a sequence of elements needed to either wrap their sequence in an extra `<div>` or access their elements per
```js
const { element} = setup(/*...*/);
element.parentElement.children
```
Both don't improve readability.

The event helpers were set up on the first element.

**How**:

Include an extra property `elements` for the sequence of elements.
If there is more than one element, set up the event helpers on the container.

**Checklist**:

- [x] Tests
- [ ] Ready to be merged

What do you think? Ready or could we do better? Should we refactor more tests first?

### DIFF
--- a/src/__tests__/helpers/utils.js
+++ b/src/__tests__/helpers/utils.js
@@ -17,9 +17,14 @@ function setup(ui, {eventHandlers} = {}) {
   div.innerHTML = ui.trim()
   document.body.append(div)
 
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
Perfect. Thank you!

PR #451, opened 2020-09-14
Also fire pointer/mouseEnter/-Leave on parents

mischnic:
**What**:

pointerEnter/mouseEnter and pointerLeave/mouseLeave don't bubble but are called for every parent (with `e.target` being set to the parent).

https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseenter_event

**Why**:

React normalizes this behaviour but it's not compatible with the native DOM/Preact.

**How**:

For hover/unhover, fire these events for all parent elements as well.

**Checklist**:

- [x] Documentation N/A
- [x] Tests
- [x] Typings N/A
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/hover.js
+++ b/src/__tests__/hover.js
@@ -1,5 +1,5 @@
 import userEvent from '../'
-import {setup} from './helpers/utils'
+import {addEventListener, setup} from './helpers/utils'
 
 test('hover', () => {
   const {element, getEventSnapshot} = setup('<button />')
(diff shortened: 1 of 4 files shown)


src/__tests__/hover.js
```diff
@@ -38,3 +38,87 @@ test('no events fired on labels that contain disabled controls', () => {
...
+      'pointerleave',
+      'pointerout',
+    ]) {
+      addEventListener(el, event, () => {
+        calls.push(`${el.tagName}: ${event}`)
+      })
+    }
+  }
+  const div = wrapper.firstChild
+  const button = div.firstChild
```
kentcdodds, line 65: Could we add a comment here explaining why `setup` isn't being used like it is for the rest of the tests?

src/__tests__/hover.js
```diff
@@ -38,3 +38,87 @@ test('no events fired on labels that contain disabled controls', () => {
...
+
+  addListeners(div)
+  addListeners(button)
```
kentcdodds, line 110: Here as well?

kentcdodds (commented):
I think I'm good with these changes. Just one thing.

kentcdodds:
Also, thank you :)

src/__tests__/hover.js
mischnic: Done, I did change it out for `setup` though and added an explanation why I didn't use `getEventSnapshot`.

kentcdodds (approved):
Perfect! Thanks :)

kentcdodds:
@all-contributors please add @mischnic for code and tests

PR #463, opened 2020-10-08
Ft issue template update

cwinters8:
**What**:
This is just an addition to the issue template for those, such as myself, whose issues may be fixed by upgrading their version of user-event.

**Why**:
To help people resolve their issues or at least verify their issue hasn't already been fixed in the most recent version of user-event. I discussed this change with @nickmccurdy and he thought it might be a good idea too.

**How**:
Added a comment to ISSUE_TEMPLATE.md right below where the user provides their current version.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Typings N/A
- [x] Ready to be merged
      

Happy to make changes to the wording on this. I did a little digging to see if I could programmatically grab the current release version but I think it would probably end up adding more complexity than it's worth because we'd also need a build system to generate this file.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -12,6 +12,13 @@ learn how: http://kcd.im/pull-request
 -->
 
 - `@testing-library/user-event` version:
+<!--
+If your version doesn't match this project's current release found at
+https://github.com/testing-library/user-event/releases, please try upgrading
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thanks!

kentcdodds:
@all-contributors please add @cwinters8 for docs

nickmccurdy:
Thanks for doing this. What other libraries do we want to document this for?

PR #467, opened 2020-10-14
fix: exclude hidden type inputs from focusable selector

### LINKED ISSUE #466: userEvent.tab tries to focus <input type="hidden" />, preventing active element from changing
- `@testing-library/user-event` version: 12.1.7

- Testing Framework and version: jest@26.5.2
  
- DOM Environment: jsdom@16.4.0
  

Relevant code or config

```javascript
test('repro', () => {
    const input = document.createElement('input');
    input.type = 'hidden';
    document.body.appendChild(input);

    const div = document.createElement('div');
    div.tabIndex = 0;
    document.body.appendChild(div);

    expect(document.activeElement).toBe(document.body);

    // This attempts to focus <input type="hidden /> so active element doesn't change
    userEvent.tab();

    // Error since active element is still document.body
    expect(document.activeElement).toBe(div);
});
```

What you did: `jest`

What happened:
```js
    expect(received).toBe(expected) // Object.is equality

    expect(received).toBe(expected) // Object.is equality

    - Expected  - 0
    + Received  + 5

    + <body>
    +   <input
    +     type="hidden"
    +   />
        <div
          tabindex="0"
        />
    + </body>
```

Problem description: Hidden-type input elements do not receive tab focus in common browsers, but `userEvent.tab` fails to exclude them from its tab-able element list which produces incorrect behavior.

Suggested solution: Change `input:not([disabled])` to `input:not([type=hidden]):not([disabled])` in `FOCUSABLE_SELECTOR` in `src/utils.js`

lazytype:
fix #466 

**What**: `tab` attempts to focus hidden type inputs

**Why**: Hidden-type input elements do not receive tab focus in common browsers, but `tab` fails to exclude them from its tab-able element list which produces incorrect behavior.

**How**: Update `FOCUSABLE_SELECTOR` to exclude hidden inputs.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/tab.js
+++ b/src/__tests__/tab.js
@@ -288,14 +288,13 @@ test('should not tab to <a> with no href', () => {
 })
 
 test('should not tab to <input> with type="hidden"', () => {
-  setup(`
-    <div>
-      <input data-testid="element" tabIndex="0" type="checkbox" />
-      <input type="hidden">
-      <input data-testid="element" type="text"  />
-    </div>`)
-
-  const [checkbox, text] = document.querySelectorAll('[data-testid="element"]')
+  const {
+    elements: [checkbox, , text],
+  } = setup(`
+    <input tabIndex="0" type="checkbox" />
+    <input type="hidden" />
+    <input type="text" />
+  `)
 
   userEvent.tab()
 
--- a/src/__tests__/tab.js
+++ b/src/__tests__/tab.js
@@ -221,7 +221,7 @@ test('should respect tab index order, then DOM order', () => {
   expect(checkbox).toHaveFocus()
 })
 
-test('should suport a mix of elements with/without tab index', () => {
+test('should support a mix of elements with/without tab index', () => {
   setup(`
     <div>
       <input data-testid="element" tabIndex="0" type="checkbox" />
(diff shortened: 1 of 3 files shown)


src/__tests__/tab.js
```diff
@@ -287,6 +287,25 @@ test('should not tab to <a> with no href', () => {
...
 
+test('should not tab to <input> with type="hidden"', () => {
+  setup(`
+    <div>
+      <input data-testid="element" tabIndex="0" type="checkbox" />
+      <input type="hidden">
+      <input data-testid="element" type="text"  />
+    </div>`)
+
+  const [checkbox, text] = document.querySelectorAll('[data-testid="element"]')
```
ph-fritsche, line 298: We can keep the code leaner here:
```js
const {elements: [checkbox,, text]} = setup(`
  <input type="checkbox" />
  <input type="hidden" />
  <input type="text"  />
`)
```

ph-fritsche (changes requested):
Thank you for the report and quick fix. :)

src/utils.js
```diff
@@ -219,7 +219,7 @@ function setSelectionRangeIfNecessary(
 }
 
 const FOCUSABLE_SELECTOR = [
-  'input:not([disabled])',
+  'input:not([type=hidden]):not([disabled])',
```
kentcdodds, line 222: Actually, the hidden attribute could be applied to any element. I think we need to add this to all of the selectors here.

kentcdodds (commented):
Thanks! One thing.

src/utils.js
ph-fritsche: I just checked in Chrome: `<div tabIndex="0" type="hidden">` has no effect.
I don't think it is specified to have any meaning on other elements.

kentcdodds: Oh, duh. I was thinking about the `hidden` attribute itself (like `<button hidden="true">`).

But yeah, the `hidden` type is what we're dealing with here and that works fine. Thanks!

I wonder whether we should/need to fix how we handle the hidden attribute though 🤔

I'll merge this as-is. Thanks!

kentcdodds: @lazytype, I mean, I'm happy to merge after you sort out @ph-fritsche's comment https://github.com/testing-library/user-event/pull/467/files#r504487490

kentcdodds (approved):
Solid, thanks!

kentcdodds:
@all-contributors please add @lazytype for code and tests

Issue #469, opened 2020-10-20
Cannot select accessible listbox options


luistak:
**Envs**

- `@testing-library/user-event` version: ^12.1.7

- Testing Framework and version: Jest@24.9.0

- DOM Environment: jsdom 16

**Relevant code or config**

https://codesandbox.io/s/silent-brook-ksldl?file=/src/App.js

**What you did:**

Tried to use `selectOptions` in with a custom select using [`Downshift`](https://github.com/downshift-js) `useSelect` hook

**What happened:**

It did not work as expected selecting the right options

**Reproduction repository:**

https://codesandbox.io/s/silent-brook-ksldl?file=/src/App.js

**Problem description:**

When trying to use the `selectOptions` API using a [non-native select](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html), the current selector does not match accessible compatible roles

**Suggested solution:**

Change [this line](https://github.com/testing-library/user-event/blob/master/src/select-options.js#L13) to

```js
Array.from(select.querySelectorAll("option, [role='option']"));
```

luistak:
The #470  PR was incomplete when selecting a non-native accessible `listbox` we can't handle like a normal option

There small differences:
- Option values need to be compared with `innerHTML` instead of `value`
- When selecting the option we need to change the `aria-selected` property instead of `selected`
- We need to click instead of firing an input event


I'm going to open a new pr with those fixes, can you please re-open this issue @kentcdodds?

PR #473, opened 2020-10-21
Added listbox options support in selectOptions

luistak:
**What**:

Added a few behaviors when selecting a `listbox` option to `selectOptions` 

**Why**:

After adding the `role="option"` selector it was missing those behaviors to work as expected

**How**:

Added those behaviors listed in [this comment](https://github.com/testing-library/user-event/issues/469#issuecomment-713528494) according to [w3](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html)

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Typings N/A
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/select-options.js
+++ b/src/__tests__/select-options.js
@@ -46,28 +46,28 @@ test('fires correct events on listBox select', () => {
     ul - pointerup
     ul - mouseup: Left (0)
     ul - click: Left (0)
-    option[value="2"][selected=false] - pointerover
+    li[value=0] - pointerover
     ul - pointerenter
-    option[value="2"][selected=false] - mouseover: Left (0)
+    li[value=0] - mouseover: Left (0)
     ul - mouseenter: Left (0)
-    option[value="2"][selected=false] - pointermove
-    option[value="2"][selected=false] - mousemove: Left (0)
-    option[value="2"][selected=false] - pointerover
+    li[value=0] - pointermove
+    li[value=0] - mousemove: Left (0)
+    li[value=0] - pointerover
     ul - pointerenter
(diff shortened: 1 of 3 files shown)


src/__tests__/select-options.js
```diff
@@ -29,6 +29,53 @@ test('fires correct events', () => {
...
+    ul - mouseover: Left (0)
+    ul - mouseenter: Left (0)
+    ul - pointermove
+    ul - mousemove: Left (0)
+    ul - pointerdown
+    ul - mousedown: Left (0)
+    ul - pointerup
+    ul - mouseup: Left (0)
+    ul - click: Left (0)
+    li[value=0] - pointerover
```
kentcdodds, line 49: Should this be firing on the `li[value=0]` element? I would expect these to fire on `li[value=2]` element since that's the option that's being selection 🤔

luistak: I will check it, because I think the `li` element has no value just `innerHTML` so this snapshot could be improved

kentcdodds: Ah yes, I wondered whether it's the utility function acting up here. If you could try to get that fixed that would be great (or you could add an ID to each of the lis).

luistak: I tried both
- Added ID to each `li`
- Added `[role="listbox"]` and  `[role="option"]` behaviors to `getElementValue` and `getElementDisplayName` utils

What do you think?

kentcdodds (approved):
Solid!

kentcdodds:
@all-contributors please add @luistak for tests

kentcdodds:

PR #495, opened 2020-11-14
Fix documented require usage (fix #485)

### LINKED ISSUE #485: Using recommended require method in README resulting in 'is not a function' errors
I'm quite new to JavaScript/React programming so please forgive me if this issue is redundant or misguided. It may be that my environment (with webpack) is rewriting/transpiling imports in a way that introduces this issue, but I thought I would log it in case it was something bigger. If not relevant to all users please feel free to close it.

- `@testing-library/user-event` version: `"@testing-library/user-event": "^12.2.0"`

- Testing Framework and version: `"mocha": "^8.1.3"`
  
- DOM Environment: `npx mocha -r @babel/register -r @babel/polyfill` on the CLI with `react` and `@testing-library/react` `require`d in the actual test. `jsdom` isn't required at the point that the issue appears.
  

### Relevant code or config

```javascript
var userEvent = require('@testing-library/user-event')
console.log(userEvent);
```
...vs...
```javascript
import userEvent from '@testing-library/user-event'
console.log(userEvent);
```

### What you did:

I used `require` instead of `import` and the behaviour differed, resulting in an `...is not a function` error when trying to use one of the methods.

### What happened:

When using the `require` method described in [README](https://github.com/testing-library/user-event#installation) I ended up with a different object assigned to `userEvent`, so when I tried to use one of the methods I got `TypeError: userEvent.clear is not a function`, because that function would actually now live at `userEvent.default.type`.

You can see the two objects output by `require`/`import` respectively below:

### `require`

```javascript
{
  default: {
    click: [Function: click],
    dblClick: [Function: dblClick],
    type: [AsyncFunction: type],
    clear: [Function: clear],
    tab: [Function: tab],
    hover: [Function: hover],
    unhover: [Function: unhover],
    upload: [Function: upload],
    selectOptions: [Function: bound selectOptionsBase],
    deselectOptions: [Function: bound selectOptionsBase],
    paste: [Function: paste]
  }
}
```

### `import`

```javascript
{
  click: [Function: click],
  dblClick: [Function: dblClick],
  type: [AsyncFunction: type],
  clear: [Function: clear],
  tab: [Function: tab],
  hover: [Function: hover],
  unhover: [Function: unhover],
  upload: [Function: upload],
  selectOptions: [Function: bound selectOptionsBase],
  deselectOptions: [Function: bound selectOptionsBase],
  paste: [Function: paste]
}
```
### Reproduction repository

See above.

### Problem description

The behaviour of the `require` method differs from that of the `import` method recommended in the README.

### Suggested solution

I'm not clear on a suggested solution, but my workaround was to use:

`const { default: userEvent } = require('@testing-library/user-event');` to import the module.

nickmccurdy:
**What**: Encouraging the use of the `default` export from User Event's CJS build

**Why**: If you don't do this, you'll get errors about properties missing from the top level build (like #485)

**How**: Destructuring (supported since Node 6, and Node 10 is the oldest supported version)

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Typings
- [ ] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ import userEvent from '@testing-library/user-event'
 
 // or
 
-var userEvent = require('@testing-library/user-event')
+const { default: userEvent } = require('@testing-library/user-event')
 ```
 
 ## API


nickmccurdy:
Do we need to update the CJS types for this or are they already correct?

nickmccurdy:
If you're using TS with CJS, `userEvent` will be incorrectly typed as `any`. Please use this workaround for now:
```ts
import userEventModule = require('@testing-library/user-event')
const userEvent = userEventModule.default
```

nickmccurdy:
@all-contributors add @nickmccurdy for docs

Issue #497, opened 2020-11-18
userEvent.click triggers warning that it must be wrapped in act in beforeEach


amandapouget:
- `@testing-library/user-event` version: 12.2.2

- Testing Framework and version: Jest 26.6.3
  
- DOM Environment: whatever default ships with Jest 26.6.3
  

Relevant code or config

```javascript
    beforeEach(() => {
        userEvent.click(button.getByTestId('foobar'));
    });
```

What you did:
Ran the test

What happened:
```javascript
Warning: An update to Foobar inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
    
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
    
    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
        in Foobar
```

Reproduction repository:

Problem description:
We don't have this problem when clicking the same button inside the test itself. But in beforeEach, it flips out and demands that we wrap in act and await. Then the warning goes away.

Suggested solution:
Either fix it or document it in the description for userEvent.click.

kentcdodds:
Hi @amandapouget,

> We don't have this problem when clicking the same button inside the test itself. But in beforeEach, it flips out and demands that we wrap in act and await.

Here's a blog post to explain what that warning is all about: https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning

And here's another blog post to suggest you change the structure of your test to avoid issues like this: https://kentcdodds.com/blog/avoid-nesting-when-youre-testing

This may help as well: https://kentcdodds.com/blog/write-fewer-longer-tests

The recommendation is to restructure your test and not do stuff like this in those setup/teardown hooks (like beforeEach and afterEach). Good luck!

amandapouget:
Can you update the documentation for userEvent.click to say either:
a) do not use in beforeEach, or
b) you must wrap in act when using in beforeEach

Right now the docs say that the _only_ time you need to wrap it is when you are using delay, and this example shows that is simply not true.

kentcdodds:
You're more than welcome to make a PR for recommendation `a`.

> Right now the docs say that the only time you need to wrap it is when you are using delay, and this example shows that is simply not true.

This is incorrect. `userEvent.click` is already wrapped in act. The reason the warning is going away fro you isn't because of `act` but because of `async/await`. If you read the blog posts I provided hopefully they will explain why.

[...]

PR #498, opened 2020-11-18
Doc need for `await` in before/after blocks to avoid warnings

amandapouget:
**What**:

**Why**: See (this issue)[https://github.com/testing-library/user-event/issues/497

**How**: Updated README at the top-level because I suspect that applies to all of the methods not just click.

**Checklist**:

- [X] Documentation
- [X] Tests
- [ ] Typings N/A
- [ ] Ready to be merged N/A
      

Feel free to clean up or clarify this language.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,11 @@ const { default: userEvent } = require('@testing-library/user-event')
 
 ## API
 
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thanks for the original PR. I made a few edits to make this more correct.

kentcdodds:
@all-contributors please add @amandapouget for docs

Issue #484, opened 2020-11-06
userEvent.type not working with <input type="time">


kellengreen:
- `@testing-library/user-event` version: `12.2.0`
- Testing Framework and version: `jest:26.6.0`
  
- DOM Environment: `@testing-library/jest-dom:5.11.5`

Relevant code or config

```javascript
import React from 'react'
import { render, fireEvent } from '@testing-library/react'
import userEvent from '@testing-library/user-event'

function InputNumber() {
  return <input type="time" role="textbox" />
}

describe('Test <InputNumber>', () => {
  it('Update with fireEvent', () => {
    const { getByRole } = render(<InputNumber />)
    const input = getByRole('textbox')
    fireEvent.change(input, { target: { value: '01:01' } })
    expect(input.value).toBe('01:01')
  })

  it('Update with userEvent', () => {
    const { getByRole } = render(<InputNumber />)
    const input = getByRole('textbox')
    userEvent.type(input, '0101AM')
    expect(input.value).toBe('01:01') // <- FAILS
  })
})
```

What you did: Called `userEvent.type` on `<input type="time">`.

What happened: `input.value` remains `""`.

Problem description: It would seem that keystrokes are not making their way to the input field. I've tested this under both `en-US` and `en-GB` locales, the result is the same.

kentcdodds:
@kellengreen, thanks for reporting this. Would you be interested in getting this supported?

gndelia:
If @kellengreen does not take it, I'd be interested in fixing it  👀

kellengreen:
Hey all, thank you for your quick replies.

If this is indeed an unsupported use case (and not a bug) then yes I would love to see this resolved. In my current role I wont be able to provide much assistance so @gndelia please feel free to take the lead on this if you're interested.

kentcdodds:
@gndelia, go for it 👍

gndelia:
ok I've been doing some testing yesterday and I think I've found the root cause. I'm leaving some insights in case anyone wants to add their thoughts or tell me if I'm in the wrong track 😅 

How it looks 

![image](https://user-images.githubusercontent.com/352474/98742214-cc01dd80-238c-11eb-882a-cc31de6ccfac.png)

(I'm going to be honest, I've never used the HTML native timer lol)

[...]

PR #502, opened 2020-11-25
feat(type): add time input support

gndelia:
**What**: Adding support for `<input type="time" />`. Check #484 

**Why**: That type of input does not work. It does with `fireEvent`, though

**How**: Adding the functionality for it. I had to interpret the typed characters to hours and minutes, adding the `:` between both.

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Typings N/A
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -1066,3 +1066,274 @@ test('navigation key: {arrowleft} and {arrowright} moves the cursor', () => {
     input[value="abc"] - keyup: c (99)
   `)
 })
+
+test('can type into an input with type `time`', () => {
+  const {element, getEventSnapshot} = setup('<input type="time" />')
+  userEvent.type(element, '0105')
+  expect(getEventSnapshot()).toMatchInlineSnapshot(`
+  Events fired on: input[value="01:05"]
+  
+  input[value=""] - pointerover
+  input[value=""] - pointerenter
+  input[value=""] - mouseover: Left (0)
+  input[value=""] - mouseenter: Left (0)
+  input[value=""] - pointermove
+  input[value=""] - mousemove: Left (0)
+  input[value=""] - pointerdown
+  input[value=""] - mousedown: Left (0)
+  input[value=""] - focus
+  input[value=""] - focusin
+  input[value=""] - pointerup
+  input[value=""] - mouseup: Left (0)
+  input[value=""] - click: Left (0)
+  input[value=""] - keydown: 0 (48)
+  input[value=""] - keypress: 0 (48)
+  input[value=""] - keyup: 0 (48)
+  input[value=""] - keydown: 1 (49)
+  input[value=""] - keypress: 1 (49)
+  input[value=""] - keyup: 1 (49)
+  input[value=""] - keydown: 0 (48)
+  input[value=""] - keypress: 0 (48)
+  input[value="01:00"] - input
+    "{CURSOR}" -> "{CURSOR}01:00"
+  input[value="01:00"] - change
+  input[value="01:00"] - keyup: 0 (48)
+  input[value="01:00"] - keydown: 5 (53)
+  input[value="01:00"] - keypress: 5 (53)
+  input[value="01:05"] - input
+    "{CURSOR}01:00" -> "{CURSOR}01:05"
+  input[value="01:05"] - change
+  input[value="01:05"] - keyup: 5 (53)
+  `)
+  expect(element.value).toBe('01:05')
+})
+
+test('can type more a number higher than 60 minutes into an input `time` and they are converted into 59 minutes', () => {
+  const {element, getEventSnapshot} = setup('<input type="time" />')
+  userEvent.type(element, '2390')
+
+  expect(getEventSnapshot()).toMatchInlineSnapshot(`
+    Events fired on: input[value="23:59"]
+    
+    input[value=""] - pointerover
+    input[value=""] - pointerenter
+    input[value=""] - mouseover: Left (0)
+    input[value=""] - mouseenter: Left (0)
+    input[value=""] - pointermove
+    input[value=""] - mousemove: Left (0)
+    input[value=""] - pointerdown
+    input[value=""] - mousedown: Left (0)
+    input[value=""] - focus
+    input[value=""] - focusin
+    input[value=""] - pointerup
+    input[value=""] - mouseup: Left (0)
+    input[value=""] - click: Left (0)
+    input[value=""] - keydown: 2 (50)
+    input[value=""] - keypress: 2 (50)
+    input[value=""] - keyup: 2 (50)
+    input[value=""] - keydown: 3 (51)
+    input[value=""] - keypress: 3 (51)
+    input[value=""] - keyup: 3 (51)
+    input[value=""] - keydown: 9 (57)
+    input[value=""] - keypress: 9 (57)
+    input[value="23:09"] - input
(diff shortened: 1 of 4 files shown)


src/__tests__/type.js
```diff
@@ -1066,3 +1066,274 @@ test('navigation key: {arrowleft} and {arrowright} moves the cursor', () => {
...
+    "{CURSOR}01:00" -> "{CURSOR}01:05"
+  input[value="01:05"] - change
+  input[value="01:05"] - keyup: 5 (53)
+  `)
+  expect(element.value).toBe('01:05')
+})
+
+test('can type more a number higher than 60 minutes into an input `time` and they are converted into 59 minutes', () => {
+  const {element, getEventSnapshot} = setup('<input type="time" />')
+  userEvent.type(element, '2390')
```
kentcdodds, line 1113: If I type `2390` in the `time` input here: https://codesandbox.io/s/user-event-playground-eo909?file=/index.html

That doesn't end up firing a change event because I need to type in `a` (for AM) or `p` (for PM). It's possible that this is a user-agent thing, in which case I think the implementation here is fine. In any case, we'll want to make sure we document how a time is intended to be entered for a given time value.

I'm thinking that rather than having them provide `0105` we should have them pass the value they want the `input.value` to be when the typing is complete. So that would be: `userEvent.type(element, '01:05')`. I think that would be more intuitive. We'd just need to make sure we handle this internally because the end user wouldn't actually bother typing the `:` I think.

kentcdodds (commented):
Thanks for your work on this! Just one thing. 

Also, let's make sure we add docs :)

src/__tests__/type.js
gndelia: > That doesn't end up firing a change event because I need to type in a (for AM) or p (for PM). It's possible that this is a user-agent thing, in which case I think the implementation here is fine. In any case, we'll want to make sure we document how a time is intended to be entered for a given time value.

In my case trying with chrome I don't have to type `a` or `p`. I guess you're right, it depends on the browser or user-agent, or something else.

> I'm thinking that rather than having them provide 0105 we should have them pass the value they want the input.value to be when the typing is complete. So that would be: userEvent.type(element, '01:05'). I think that would be more intuitive. We'd just need to make sure we handle this internally because the end user wouldn't actually bother typing the : I think.

_technically_ speaking, all the non-digit characters are ignored, as that's what happens when using the `<input type="time" />`. So I could rewrite the tests to use `:` and they will work with the current implementation (and the docs could be with that). What do you think?

if that's enough as the request, I could rewrite the tests and the docs (yet to be added) to show the examples using  `:`. That's probably better in terms of DX, but as you say the user would never type `:`. Working without `:` is an extra in that sense

If you think it is better to change the implementation to just split hours and minutes by `:`, then I can do it; in this scenario, the user would _always_ have to type `:` or it wouldn't work (with the current approach, both works). and I'd have to validate both separatedly

(I also went without `:` probably biased by the example in the ticket, which was not using it).

kentcdodds: That all sounds great. Thank you. I think having the `:` in the docs would be preferable/sensible. And having at least one test that uses `:` as well as at least one that does not would be good.

Thanks for all of this work!

gndelia: changes applied in https://github.com/testing-library/user-event/pull/502/commits/8333075b4da7bebb856904a748f0c87992d82940

- updated docs to have an example of using `<input type="time />`. My main doubt is that there are no roles associated to this type of input, so the only way to query it was with a `data-testid`. Should I use another option?

 - I updated all tests to use `01:05`, and also added one without the `:` to verify it works in both scenarios

src/type.js
```diff
@@ -376,6 +385,19 @@ function typeCharacter(
...
+  timeNewEntry,
+) {
+  if (
+    isValidInputTimevalue(currentElement(), timeNewEntry) &&
+    prevValue !== timeNewEntry
+  ) {
+    fireEvent.change(currentElement(), {target: {value: timeNewEntry}})
+  }
+}
+
```
kentcdodds, line 400: I think we don't need this to be a function. Could you inline it please?

src/utils.js
```diff
@@ -269,6 +277,47 @@ function isValidDateValue(element, value) {
...
+    }
+    return build(onlyDigitsValue, index)
+  }
+  if (value.length === 2) {
+    return value
+  }
+  return build(onlyDigitsValue, 2)
+}
+
+function isValidInputTimevalue(element, timeValue) {
```
kentcdodds, line 313:
```suggestion
function isValidInputTimeValue(element, timeValue) {
```

src/utils.js
```diff
@@ -280,6 +329,8 @@ export {
   setSelectionRangeIfNecessary,
   eventWrapper,
   isValidDateValue,
+  isValidInputTimevalue,
```
kentcdodds, line 332:
```suggestion
  isValidInputTimeValue,
```

src/type.js
```diff
@@ -13,6 +13,8 @@ import {
   getSelectionRange,
   getValue,
   isContentEditable,
+  isValidInputTimevalue,
```
kentcdodds, line 16:
```suggestion
  isValidInputTimeValue,
```

src/type.js
```diff
@@ -324,6 +326,11 @@ function typeCharacter(
         newEntry = textToBeTyped
       }
 
+      const timeNewEntry = buildTimeValue(textToBeTyped)
+      if (isValidInputTimevalue(currentElement(), timeNewEntry)) {
```
kentcdodds, line 330:
```suggestion
      if (isValidInputTimeValue(currentElement(), timeNewEntry)) {
```

kentcdodds (commented):
This is super. Just a few small things and then we can get this merged. Thanks!

src/type.js
gndelia: I tried but there was a rule in the linter when running `kcd validate` that was stating there was too much branching complexity (I cannot recall the name right now)

gndelia: this 👇 
> error  Function 'typeCharacter' has a complexity of 16. Maximum allowed is 14

kentcdodds: Oh, lol, bummer.... We should probably do a bigger refactor of that function or re-evaluate whether the rule is catching an actual problem here... For now, let's leave this as-is.

src/type.js
gndelia: Applied here https://github.com/testing-library/user-event/pull/502/commits/32b5dd248184898055a1f5ee2ec5ace134add290 (same for the other comments)

src/__tests__/type.js
gndelia: to clarify, I tried with labels in testing-playground but it did not work either - that's why I went for `data-testid`

kentcdodds: How about this? https://testing-playground.com/gist/7b5d922642a0a1814209c81db6867b59/1b0329837c8afa1d5e972e219ed65ba041f69c82

gndelia: I swear I tried it and I was getting an error 🤔  but 🤷 
updated in https://github.com/testing-library/user-event/pull/502/commits/774a0303a7ca3b9e6a0fa3e08c701febc8f7db99

kentcdodds (approved):
Well done 👏

kentcdodds:
@all-contributors please add @gndelia for code, tests, and docs

kentcdodds:

PR #507, opened 2020-11-30
feat: add capslock to userEvent.type

Sonic12040:
**What**: Adding CapsLock to the modifier key support so that tests using CapsLock can be implemented.

**Why**: When typing and using the CapsLock modifier, we need to be able to write tests that implement the CapsLock key.

**How**: Added a test to the type-modifiers file, then implemented the suggested changes from Kent C Dodds office hours to the type file.

**Checklist**:

- [x] Documentation
- [x] Tests
- [N/A] Typings
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,15 +89,14 @@ import userEvent from '@testing-library/user-event'
 
 // or
 
-const { default: userEvent } = require('@testing-library/user-event')
+const {default: userEvent} = require('@testing-library/user-event')
 ```
 
 ## API
 
 Note: All userEvent methods are synchronous with one exception: when `delay`
-with `userEvent.type` as described below). We also discourage using
-`userEvent` inside `before/after` blocks at all, for important reasons
-described in
+with `userEvent.type` as described below). We also discourage using `userEvent`
+inside `before/after` blocks at all, for important reasons described in
 ["Avoid Nesting When You're Testing"](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing).
 
 ### `click(element, eventInit, options)`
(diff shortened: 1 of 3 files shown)


src/__tests__/type-modifiers.js
```diff
@@ -353,6 +353,37 @@ test('{shift}a{/shift}', () => {
...
+    input[value=""] - pointermove
+    input[value=""] - mousemove: Left (0)
+    input[value=""] - pointerdown
+    input[value=""] - mousedown: Left (0)
+    input[value=""] - focus
+    input[value=""] - focusin
+    input[value=""] - pointerup
+    input[value=""] - mouseup: Left (0)
+    input[value=""] - click: Left (0)
+    input[value=""] - keydown: CapsLock (20)
```
kentcdodds, line 377: Ah, this one is tricky. Basically for caps lock, it should keydown *and* keyup here, but preserve the modifier state. Hmmm...

src/__tests__/type-modifiers.js
```diff
@@ -353,6 +353,37 @@ test('{shift}a{/shift}', () => {
...
+    input[value=""] - keydown: a (97)
+    input[value=""] - keypress: a (97)
+    input[value="a"] - input
+      "{CURSOR}" -> "a{CURSOR}"
+    input[value="a"] - keyup: a (97)
+    input[value="a"] - keyup: CapsLock (20)
```
kentcdodds, line 383: And then this should keydown and keyup and disable the modifier state. I'm afraid this one isn't quite as simple as I thought at first.

src/__tests__/type-modifiers.js
Sonic12040: So, do you think a PR will also be needed to fireEvent for the case of CapsLock?

kentcdodds: No, `fireEvent` is all about dispatching individual events in complete isolation. It is not concerned about state. This will only need to happen here in user-event. It's possible we won't be able to re-use the existing `createModifierCallbackEntries` for this one.

Sonic12040: Scratch that. Updated with a switch/case. I'm not sure if it's initially beneficial to write it that way, but I didn't know if other modifier keys might also require the behavior at some point.

Sonic12040: Saw the comment after. I can write a separate function for it if preferred?

kentcdodds: @Sonic12040, could you try removing the custom logic and inlining it for the caps lock. I think that will result in simpler code in this case. I'm worried that this is a hasty abstraction ([AHA Programming](https://kcd.im/aha)) and I think the small amount of duplication will be better.

src/type.js
```diff
@@ -42,6 +42,12 @@ const modifierCallbackMap = {
     keyCode: 93,
     modifierProperty: 'metaKey',
   }),
+  ...createCapsLockModifierCallbackEntries({
```
kentcdodds, line 45: Rather than using a function, maybe it would make more sense to just inline this one with a comment explaining why it can't use `createModifierCallbackEntries`?

kentcdodds (commented):
Thanks for your patience on this! I think we're almost there.

src/__tests__/type-modifiers.js
Sonic12040: Updated with a duplicate function that performs the CapsLock behavior.

src/type.js
Sonic12040: Done. Sorry about the delay; I missed this comment at first.

README.md
```diff
@@ -189,20 +188,21 @@ to `await`!
...
+| `{shift}`      | Shift      | `shiftKey`         | Does **not** capitalize following characters.                                                                                                                       |
+| `{ctrl}`       | Control    | `ctrlKey`          |                                                                                                                                                                     |
+| `{alt}`        | Alt        | `altKey`           |                                                                                                                                                                     |
+| `{meta}`       | OS         | `metaKey`          |                                                                                                                                                                     |
+| `{capslock}`   | CapsLock   | `modifierCapsLock` | Fires both keydown and keyup when used.                                                                                                                             |
```
kentcdodds, line 205:
```suggestion
| `{capslock}`   | CapsLock   | `modifierCapsLock` | Fires both keydown and keyup when used (simulates a user clicking their "Caps Lock" button to enable caps lock).                                                                                                                             |
```

kentcdodds (approved):
I love it. Thank you for taking the time to make this happen!

kentcdodds:
@all-contributors please add @Sonic12040 for code, tests, and docs

kentcdodds:

PR #511, opened 2020-12-02
Support arrowup and arrowdown events

### LINKED ISSUE #503: Add support for arrowdown and arrowup events
- Testing-library/user-event` version: 12.2.2
- Testing Framework and version: jest/26.0.15
- DOM Environment: react-dom/17.0.0

What you did:

I have a module I'm migrating from Enzyme over to testing-library, the module is a React hook that for the most part manages keyboard controls to follow accessibility guidelines for dropdown menus. Part of our tests simulate `arrowdown` and `arrowup` events while focused on list items to ensure the correct resulting focused element is obtained. We ran into this issue while trying to replicate that behavior using `userEvent`.

What happened:

`userEvent.type` only supports `ArrowLeft` and `ArrowRight` keys.

Problem description:

While trying to implement tests to replace Enzyme with testing-library we were attempting to use `userEvent` instead of `fireEvent` where ever possible, as suggested by the documentation. However, we are currently unable to replicate tests that used `ArrowDown` and `ArrowUp` as that's just not one of the special characters supported by `userEvent.type`.

Suggested solution:

Add `ArrowDown` and `ArrowUp` as supported special keys for `userEvent.type`. Also, while reading over the documentation and through some issues, the documentation for `userEvent.type` seems off the mark for the variety of purposes it is used for now:

> Writes text inside an <input> or a <textarea>.

This should probably be modified to indicate that it's used with a variety of types of elements now, or a new method could be added to `userEvent` meant for these situations were you aren't necessarily "typing".

corymharper:
**What**:

These changes add support for the arrow up and down keys to `userEvent.type`.

**Why**:

Arrow up and down keys are used for various accessibility reasons and are part of regular user actions.

**How**:

I added two functions, `handleArrowDown` and `handleArrowUp`, and tied those into the existing special character map.

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Typings N/A
- [x] Ready to be merged

Unless something unexpected comes up, this is probably ready for merging immediately.

Closes #503

### DIFF
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -1066,3 +1066,59 @@ test('navigation key: {arrowleft} and {arrowright} moves the cursor', () => {
     input[value="abc"] - keyup: c (99)
   `)
 })
+
+test('{arrowdown} fires keyup/keydown events', () => {
+  const {element} = setup('<input />')
+  const handleKeyDown = jest.fn()
+  const handleKeyUp = jest.fn()
+
+  addListeners(element, {
+    eventHandlers: {
+      keyDown: handleKeyDown,
+      keyUp: handleKeyUp,
+    },
+  })
+
+  userEvent.type(element, '{arrowdown}')
+
+  expect(handleKeyDown).toHaveBeenCalledWith(
+    expect.objectContaining({
+      key: 'ArrowDown',
+      keyCode: 40,
+    }),
+  )
+  expect(handleKeyUp).toHaveBeenCalledWith(
+    expect.objectContaining({
+      key: 'ArrowDown',
+      keyCode: 40,
+    }),
+  )
+})
+
+test('{arrowup} fires keyup/keydown events', () => {
+  const {element} = setup('<input />')
+  const handleKeyDown = jest.fn()
+  const handleKeyUp = jest.fn()
+
+  addListeners(element, {
+    eventHandlers: {
+      keyDown: handleKeyDown,
+      keyUp: handleKeyUp,
+    },
+  })
+
+  userEvent.type(element, '{arrowup}')
+
+  expect(handleKeyDown).toHaveBeenCalledWith(
+    expect.objectContaining({
+      key: 'ArrowUp',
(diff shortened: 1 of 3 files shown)


src/__tests__/type.js
```diff
@@ -1066,3 +1066,59 @@ test('navigation key: {arrowleft} and {arrowright} moves the cursor', () => {
...
+  const {element} = setup('<input />')
+  const handleKeyDown = jest.fn()
+  const handleKeyUp = jest.fn()
+
+  addListeners(element, {
+    eventHandlers: {
+      keyDown: handleKeyDown,
+      keyUp: handleKeyUp,
+    },
+  })
```
kentcdodds, line 1080: Any reason we're using the `addListeners` util here instead of the `getEventSnapshot` as above? Gives us a better idea of all events fired.

I think this would be better for both tests.

kentcdodds (commented):
This looks pretty solid to me 👍 Just one note about the tests.

src/__tests__/type.js
corymharper: Will do! I've made this change for both tests.

kentcdodds (approved):
Brilliant 👏

kentcdodds:
Oh rats, there are merge conflicts. Could you address those?

corymharper:
> Oh rats, there are merge conflicts. Could you address those?

Done!

kentcdodds (approved):
Perfect! Thank you :clap:

Shipow:
It's really the perfect timing for this, thanks!!!

wilsonmedeiros:
Hello Everyone, I'm handling a test on input type number, where I expect increase and decrease value when typing on keyboard arrow up and down respectively. The Issue is that nothing happens, the input value remains the same. Someone could look into my code and see what is happening. I believe that userEvent and input are not communicating with each other in this particular event

Code: https://codesandbox.io/s/late-http-ff27ln?file=/src/App.test.tsx

PR #510, opened 2020-12-02
feat(specialChars): Create and expose `specialChars` map

vasilii-kovalev:
**What**: Create and expose `specialChars` map

**Why**:
* It can help to avoid typos providing the constant values;
* Instead of creating the set manually in each project again and again, it could be more convenient to use the official, always up-to-date one.

**How**:
* Create a constant `specialCharMap`
* Link it with `specialCharCallbackMap`
* Export the constant from `src/type.js` and create a named export of the constant called `specialChars` in `src/index.js`
* Add an appropriate enum in `typings/index.d.ts` called `specialChars`
* Add `specialChars` section to `README.md`

**Checklist**:

- [X] Documentation
- [ ] Tests
- [X] Typings
- [X] Ready to be merged
      

P.S.: I've defined enum because `const` is not allowed in `index.d.ts`. If `src/type.js` could be written in TypeScript, it could be possible to use a [const assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions) syntax instead, so the `specialCharMap` constant could be a single source of truth.

PR #526, opened 2020-12-20
docs: Remove duplicated content on special chars

DaleSeo:
**What**:

The content on special chars that can be used in `type` method is already covered [above](https://github.com/testing-library/user-event#special-characters).

**Checklist**:

- [X] Documentation
- [ ] Tests N/A
- [ ] Typings N/A
- [ ] Ready to be merged N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -521,47 +521,6 @@ test('should paste text in input', () => {
 You can use the `eventInit` if what you're pasting should have `clipboardData`
 (like `files`).
 
-### `specialChars`
-
-A handful set of special characters used in [type](#typeelement-text-options)
-method.
-
-| Key        | Character      |
-| ---------- | -------------- |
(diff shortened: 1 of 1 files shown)


vasilii-kovalev:
This section was added in #510 with intention to:
* Highlight what special characters can be used, because not all of the characters from [this section](https://github.com/testing-library/user-event#special-characters) can be are added there (they have different implementations)
* Show usage case

It would be great not to just delete the "duplicated" section, but merge these two somehow. See [this comment](https://github.com/testing-library/user-event/pull/510#issuecomment-737415791).

ph-fritsche:
I'll close this as won't merge for the moment. If one comes up with a satisfying merge of those two sections, I'll gladly reopen. :)

Issue #529, opened 2020-12-22
upload() doesn't convert file to an array with input multiple


tozaicevas:
Could someone explain to me the logic behind this line of code: https://github.com/testing-library/user-event/blob/master/src/upload.js#L17 ?

Consider this code snippet:

```
  function handleChooseImage(event: ChangeEvent<HTMLInputElement>) {
    const selectedFiles = event.target.files

    if (!selectedFiles?.length) return
   
    ...
  }

      <input
        data-testid="add-photos-input"
        ref={inputRef}
        type="file"
        onChange={handleChooseImage}
        accept="image/*"
        multiple
        hidden
      />
```

My goal is to write following lines of test code
```
      const imageFile = new File([], 'imgname', { type: 'image/png' })
      const imageInput = screen.getByTestId('add-photos-input')

      userEvent.upload(imageInput, imageFile)
```

This test code result in event handler being called with an object, where `length` is `undefined` and `item()` returns `undefined`:

`selectedFiles:  { length: undefined, item: [Function: item] }` 

Is that expected behaviour?

ph-fritsche:
The function expects you to provide `File` for single file inputs and `File[]` for multiple file inputs.
```js
userEvent.upload(multipleFileInput, [imageFile])
```
I think this could be improved by just converting it.

ph-fritsche:
resolved in [12.6.2](https://github.com/testing-library/user-event/releases/tag/v12.6.2)

PR #530, opened 2020-12-22
fix(upload): accept File or File[] #529

ph-fritsche:
**What**:

```js
const f = new File(/*...*/)

// these two will be equivalent
userEvent.upload(element, [f])
userEvent.upload(element, f)
```

**Why**:

Previous implementation required the user to provide `File` for single file inputs and `File[]` for multiple file inputs.

**How**:

Convert to an array if it is not.
Truncate to length 1 if input is not multiple.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Typings N/A
- [x] Ready to be merged

### DIFF
--- a/src/upload.js
+++ b/src/upload.js
@@ -6,16 +6,14 @@ import {focus} from './focus'
 function upload(element, fileOrFiles, init) {
   if (element.disabled) return
 
-  let files
-  let input = element
-
   click(element, init)
-  if (element.tagName === 'LABEL') {
-    files = element.control.multiple ? fileOrFiles : [fileOrFiles]
-    input = element.control
-  } else {
-    files = element.multiple ? fileOrFiles : [fileOrFiles]
-  }
+
+  const input = element.tagName === 'LABEL' ? element.control : element
+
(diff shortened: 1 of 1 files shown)


kentcdodds:
I'm not sure I understand the motivation for this change.

> Previous implementation required the user to provide File for single file inputs and File[] for multiple file inputs.

I think that's sensible and expected. I'm what situation would I want to pass an array for a single or a single file for a multiple?

ph-fritsche:
> I think that's sensible and expected. I'm what situation would I want to pass an array for a single or a single file for a multiple?

As #529 demonstrated people get confused why their file input won't accept the file upload.

And I like the thought of being as agnostic as possible about implementation details.
If someone has a file input that changes to `multiple` a test where someone tries to upload a file to it should not fail.
If a file input changes away from `multiple` a test trying to e.g. drag&drop one file should still pass and one where he drags two files should fail because the second file gets lost not because of syntactic errors.

kentcdodds (approved):
That makes sense. Thanks!

PR #534, opened 2020-12-28
Support 'Enter' key press on links

benadamstyles:
**What**: in `handleEnter`, there is currently a check for whether the element is clickable, and if it is, it fires a click event on that element. This PR includes anchor tags (with truthy `href` values) in that check.

**Why**: anchor tags are incorrectly missed out of this logic – browsers fire click events on anchor tags when they are focused and the user presses the `Enter` key.

**How**: included anchor tags in the `isClickable` util function.

**Checklist**:

- [x] ~~Documentation~~ N/A
- [x] Tests
- [x] ~~Typings~~ N/A
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -1121,43 +1121,43 @@ test('can type into an input with type `time`', () => {
   const {element, getEventSnapshot} = setup('<input type="time" />')
   userEvent.type(element, '01:05')
   expect(getEventSnapshot()).toMatchInlineSnapshot(`
-  Events fired on: input[value="01:05"]
-  
-  input[value=""] - pointerover
-  input[value=""] - pointerenter
-  input[value=""] - mouseover: Left (0)
-  input[value=""] - mouseenter: Left (0)
-  input[value=""] - pointermove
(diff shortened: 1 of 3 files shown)


ph-fritsche:
`isClickable` is used by `handleSpace` too, but `space` does not trigger a `click` on `Anchor` elements in the browser.
This would introduce a bug there.

I suggest to edit this instead of the util:
https://github.com/testing-library/user-event/blob/287e5b8d0a5c537297b5a338cdd60a476049c375/src/type.js#L582
```js
if (isClickable(currentElement()) || isInstanceofElement(currentElement(), 'HTMLAnchorElement')) {
```

(Maybe slap some comment on `isClickable` (or even change the name) that it only covers clickable *inputs*.)

benadamstyles:
@ph-fritsche Thanks for the feedback! I think I've addressed everything.

ph-fritsche (changes requested):

ph-fritsche (approved):

ph-fritsche:
@all-contributors add @benadamstyles for code, tests

PR #536, opened 2020-12-30
Add type support for {home} and {end}

curiosity26:
**What**: I've added support for typing the Home and End keys

**Why**: Per aria-role spec for listbox, Home and End are necessary inputs that need to be handled and tested. Not only that, it's common in textbox input, so why not, right? 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role

**How**: Since Home and End are navigation keys, I added to the existing `navigationKeys` function to support `Home` and `End` and ensured cursor functionality worked as intended.

**Checklist**:

- [X] Documentation
- [X] Tests
- [X] Typings
- [X] Ready to be merged


src/keys/navigation-key.js
```diff
@@ -3,6 +3,12 @@ import {fireEvent} from '@testing-library/dom'
 import {setSelectionRangeIfNecessary} from '../utils'
 
 const keys = {
+  Home: {
+    keyCode: 35,
+  },
+  End: {
+    keyCode: 36,
+  },
```
luistak, line 11: Following [keycode.info](https://keycode.info/)'s table, those keys are inverted is it right?
image

curiosity26:
Omg. Dyslexia strikes again. Let me fix that. Good catch.

src/keys/navigation-key.js
curiosity26: @luistak I've corrected the blunder.

ph-fritsche (approved):

luistak (approved):

audiolion:
Who makes the call to merge? This has been hanging out in approved for a while.

kentcdodds:
I'm no longer an active maintainer. Any maintainer is welcome to merge a PR if you feel like it's a good direction. If you run into issues, let me know.

kentcdodds:
@ph-fritsche can you not merge this yourself? We should give you merge access if you don't already have it.

kentcdodds:
@ph-fritsche can you not merge this yourself? We should give you merge access if you don't already have it.

kentcdodds:
Just gave you merge access 👍

nickmccurdy:
@all-contibutors add @curiousity26 for code

nickmccurdy:
@all-contibutors add @curiosity26 for code

nickmccurdy:
@all-contributors add @curiosity26 for code

Issue #542, opened 2021-01-11
`selectOptions` fails on options with `role` attribute


comatory:
- `@testing-library/user-event` version:

`12.6.0`

- Testing Framework and version:

- `jest@26.6.0`

- DOM Environment:

- not sure, I'm using `react-scripts@3.4.3` in CodeSandbox

[Relevant code or config](https://codesandbox.io/s/user-event-changle-handler-with-role-p78f5?file=/src/Select.spec.js:190-586)

```javascript
// Select.js

import React from "react";

const Select = ({ value, options, role, onChange }) => {
  return (
    <select
      value={value}
      onChange={onChange}
      role={role === null ? undefined : "listbox"}
      data-testid="select"
    >
      {options.map(({ label, value }) => (
        <option
          key={`option-${value}`}
          value={value}
          role={role === null ? undefined : "option"}
          data-testid={`option-${value}`}
        >
          {label}
        </option>
      ))}
    </select>
  );
};

export default Select;

// Select.spec.js
  it("should call onChange callback", () => {
    const onChange = jest.fn();
    const options = [
      { label: "Foo", value: "1" },
      { label: "Bar", value: "2" }
    ];
    const screen = render(
      <Select value="1" options={options} onChange={onChange} />
    );

    userEvent.selectOptions(screen.getByTestId("select"), "2");

    expect(onChange).toHaveBeenCalledTimes(1);
  });


```

What you did:

I used `role` property on my react component. My native select component has role `listbox`. I created a test for identifying if `onChange` handler is called.

What happened:

When using `userEvent.selectOptions` my `onChange` handler is not called. However when `role` attribute is removed, the test passes.

Reproduction repository:

See this sandbox: https://codesandbox.io/s/user-event-changle-handler-with-role-p78f5?file=/src/Select.spec.js:190-586

Problem description:

`change` event does not seem to be triggered when the DOM element has `role` property defined.

Suggested solution:

comatory:
Clarification: Actually it seems to be a problem not with `role` present on `select` DOM node, but actually `option` DOM node.

ph-fritsche:
resolved in [12.6.1](https://github.com/testing-library/user-event/releases/tag/v12.6.1)

PR #543, opened 2021-01-11
fix: selectOptions

ph-fritsche:
**What**:

These changes fix bugs on selectOptions.

**Why**:

1. `selectOptions` failed to trigger correct events on options with `role` attribute. - see #542
2. `selectOptions` missed pointer events that happen when selecting an option of regular `<select>`.
3. `selectOptions` made the false assumption the browser would change `aria-selected` attributes when clicking an element with `role="option"`.

See https://codesandbox.io/s/user-event-changle-handler-with-role-forked-d5ubs?file=/src/LoggingSelects.js for tracking events that actually happen in the browser.

**How**:

1. Decide which event pattern to follow based on element type of the `listbox` instead of `role` attribute of the `option`.
2. Add the pointer events observed in the browser.
3. Move event handling into utility component and fix the EventSnapshot.

**Checklist**:
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/helpers/utils.js
+++ b/src/__tests__/helpers/utils.js
@@ -74,6 +74,17 @@ function setupListbox() {
   document.body.append(wrapper)
   const listbox = wrapper.querySelector('[role="listbox"]')
   const options = Array.from(wrapper.querySelectorAll('[role="option"]'))
+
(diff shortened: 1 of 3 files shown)


kentcdodds (approved):
This looks good to me! Thanks 👏

PR #550, opened 2021-01-29
Do not assume global HTMLSelectElement

nathanforce:
**What**: Do not assume that `HTMLSelectElement` is defined globally.

**Why**: Running Node without a shimmed browser environment has no browser globals.

**How**: Following patterns elsewhere in the code, lookup HTMLSelectElement from the `document`

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests
- [ ] Typings N/A
- [x] Ready to be merged
      

I've not added any tests here. I think there may be value in a suite of tests that verifies the library functions in a non-dom environment but didn't want to bloat the PR.

### DIFF
--- a/src/select-options.js
+++ b/src/select-options.js
@@ -36,7 +36,7 @@ function selectOptionsBase(newValue, select, values, init) {
 
   if (select.disabled || !selectedOptions.length) return
 
-  if (select instanceof HTMLSelectElement) {
+  if (select instanceof select.ownerDocument.defaultView.HTMLSelectElement) {
     if (select.multiple) {
       for (const option of selectedOptions) {
         // events fired for multiple select are weird. Can't use hover...


ph-fritsche:
`document.defaultView` might be `null`.

When it yields a `window` I guess this usually has the `HTML` elements als properties, but it is not required to do so per specs.
So I guess there could also be environments not implementing `window.HTMLSelectElement`.

nathanforce:
I suppose you're right.  It feels like that is starting to push the limits of assumptions made by this library and by the testing-library ecosystem.  My understanding is that testing library expects to run against a DOM or a DOM-like thing, but not necessarily run _in_ that environment, so assuming the global is not desirable.  

Maybe all usage of this "lookup from defaultView" pattern should be updated to safely guard against null.  Here is another usage: https://github.com/testing-library/user-event/blob/0d30a790f81c1efdbdace4f54c55c62bcddda4b6/src/utils.js#L259-L260

All said, it feels much less likely that somebody would be using the testing-library ecosystem against a DOM that does not implement the spec compared with using it in a pure Node environment.  Outside of Jest, most test runners I've seen do not pollute the global with DOM globals.

ph-fritsche:
> but not necessarily run _in_ that environment, so assuming the global is not desirable.

Agreed.

> a DOM that does not implement the spec compared with using it in a pure Node environment

My point being that when you have a pure Node environment you use some DOM implementation.
That DOM implementation might go beyond the specs but it does not violate them when it does not provide the HTMLElement constructors as properties of `window`.

> Maybe all usage of this "lookup from defaultView" pattern should be updated to safely guard against null.

I think so.

I think this whole "lookup from defaultView" is flawed because it assumes that the element constructor is on the `window` 
So I'm not sure if it is any safer than assuming the global.

> Outside of Jest, most test runners I've seen do not pollute the global with DOM globals.

The clean solution might be to let the user plug in a callback to verify the interface.
Defaulting to a helper which uses fallbacks:
1) `element.ownerDocument.defaultView`
2) `global`
3) `element.tagName` ?

So the vast majority of users is fine and those using exotic environments can plug in a workaround.

Could be implemented in [@testing-libary/dom](https://github.com/testing-library/dom-testing-library).

nathanforce:
@ph-fritsche How would you suggest we proceed? Without this safety check I cannot use the latest release of this package.

Overall I agree there should be a system wide solution for ensuring the library is runnable against all dom-like environments.

>I think this whole "lookup from defaultView" is flawed because it assumes that the element constructor is on the window
**So I'm not sure if it is any safer than assuming the global.**

In theory you may be right, but in practice I think that a very strong majority of users will be using JSDOM, so my changes here are safer than what exists now, but agree there is room for improvement.  

As this change conforms with existing solutions within the codebase I propose we proceed in order to avoid runtime errors and then take the time to solve the larger issue.

ph-fritsche:
@nathanforce I added a helper that should solve your problem in #552.
If https://github.com/testing-library/dom-testing-library/pull/885 gets merged, we can remove the code from this library once the peerDependency is bumped and resolve any problems that might come up with different environments there.

ph-fritsche:
@nathanforce Your problem is fixed in [v12.6.3](https://github.com/testing-library/user-event/releases/tag/v12.6.3) :tada:

PR #552, opened 2021-01-31
fix: check element type in different environments

ph-fritsche:
Add a helper to verify the element type in different environments.

**What**:

Add a helper to utils.
This can be removed from this library and replaced by a shared helper in `@testing-library/dom` once it is bumped to a version that includes https://github.com/testing-library/dom-testing-library/pull/885.

**Why**:

See #550 

**How**:

The helper tries to find the element constructor on the window.

As the window is neither required to be associated with the observed document nor is it required to provide any element constructors, the helper provides a fallback for other environments than Jest+JSDOM.

**Checklist**:

- N/A Documentation
- [x] Tests
- N/A Typings
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/utils.js
+++ b/src/__tests__/utils.js
@@ -1,0 +1,73 @@
+import {isInstanceOfElement} from '../utils'
+import {setup} from './helpers/utils'
+
+// isInstanceOfElement can be removed once the peerDependency for @testing-library/dom is bumped to a version that includes https://github.com/testing-library/dom-testing-library/pull/885
+describe('check element type per isInstanceOfElement', () => {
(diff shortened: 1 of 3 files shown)


kentcdodds:
I think I'd prefer to merge the one in DOM, and then just bump the peerDep here, but either way this looks fine to me 👍 thanks!

ph-fritsche:
@kentcdodds https://github.com/testing-library/dom-testing-library/pull/885 is closed so I guess we should merge this one.

kentcdodds:
Sounds good to me 👍

kentcdodds (approved):
Looks good to me 👍

PR #556, opened 2021-02-11
Don't allow tabbing to elements inside hidden parents

benadamstyles:
**What**: Don't allow tabbing to elements inside hidden parents.

**Why**: I _think_ this correctly matches browser behaviour.

**How**: Add a `!el.closest('[hidden]')` check in the focusable elements filter.

**Checklist**:

- [x] Documentation N/A
- [x] Tests
- [x] Typings N/A
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/tab.js
+++ b/src/__tests__/tab.js
@@ -17,7 +17,7 @@ test('fires events when tabbing between two elements', () => {
   userEvent.tab()
   expect(getEventSnapshot()).toMatchInlineSnapshot(`
     Events fired on: div
-    
+
     input#a[value=""] - keydown: Tab (9)
     input#a[value=""] - focusout
     input#b[value=""] - focusin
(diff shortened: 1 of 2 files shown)


ph-fritsche:
Thanks for the report. Yes, we should exclude hidden elements. :)

But while elements with `hidden` attribute should not contain elements that are relevant for the user, they are still displayed and available if a stylesheet overwrites `display`.

```html
<input/>
<button hidden style="display: inline-block">click</button>
<input />
```

So we need to traverse the ancestors and verify if they are visible.

I've opened https://github.com/benadamstyles/user-event/pull/1 with a solution.
But the bug in jsdom causing `window.getComputedStyle()` to ignore `hidden` attributes has to be resolved first.

ph-fritsche:
Replaced by #579

PR #558, opened 2021-02-18
fix: do not allow to upload files with invalid formats

LauraBeatris:
**What**: Don't allow to upload files with invalid formats.

**Why**: Files with formats that aren't accepted aren't uploaded in an input file, in fact, it's not even allowed to select them, so the tests should also follow the same behavior.

**How**: Verify if the element has an `accept` attribute, if it does, verify if the file/files match the MIME types specified. 

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Typings
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/upload.js
+++ b/src/__tests__/upload.js
@@ -183,7 +183,7 @@ test('should upload multiple files with accepted format', () => {
   const {element} = setup(`
     <input 
       type="file" 
-      accept="image/*,audio/*,text/csv" multiple 
+      accept="video/*,audio/*,.png" multiple 
     />
   `)
 
--- a/src/__tests__/upload.js
+++ b/src/__tests__/upload.js
@@ -206,7 +206,6 @@ test('should not upload multiple files with unaccepted formats', () => {
     new File(['hello'], 'hello.txt', {type: 'text/plain'}),
     new File(['there'], 'there.pdf', {type: 'application/pdf'}),
     new File(['there'], 'there.png', {type: 'image/png'}),
-    new File(['there'], 'there.mp4', {type: 'video/mp4'}),
   ]
   const {element} = setup(`
     <input id="files" type="file" accept="video/*" multiple />
@@ -214,5 +213,5 @@ test('should not upload multiple files with unaccepted formats', () => {
 
   userEvent.upload(element, files)
 
-  expect(element.files).toHaveLength(1)
+  expect(element.files).toHaveLength(0)
 })
--- a/src/__tests__/upload.js
+++ b/src/__tests__/upload.js
@@ -164,31 +164,55 @@ test('should call onChange/input bubbling up the event when a file is selected',
   expect(onInputForm).toHaveBeenCalledTimes(1)
 })
 
-test('should not upload file with invalid unaccepted format', () => {
+test('should upload file with accepted format', () => {
   const file = new File(['hello'], 'hello.png', {type: 'image/png'})
-
-  const {element} = setup('<input type="file" accept="image/jpg" />')
+  const {element} = setup('<input type="file" accept="image/png" />')
 
   userEvent.upload(element, file)
 
-  expect(element.files[0]).toBeUndefined()
-  expect(element.files.item(0)).toBeNull()
-  expect(element.files).toHaveLength(0)
+  expect(element.files).toHaveLength(1)
 })
 
-test('should not upload multiple files with invalid unaccepted formats', () => {
+test('should upload multiple files with accepted format', () => {
   const files = [
-    new File(['hello'], 'hello.txt', {type: 'text/plain'}),
-    new File(['there'], 'there.pdf', {type: 'application/pdf'}),
+    new File(['hello'], 'hello.png', {type: 'image/png'}),
+    new File(['there'], 'there.jpg', {type: 'audio/mp3'}),
+    new File(['there'], 'there.csv', {type: 'text/csv'}),
+    new File(['there'], 'there.jpg', {type: 'video/mp4'}),
   ]
-
   const {element} = setup(`
-    <input id="files" type="file" accept="image/jpeg,image/png" multiple />
+    <input 
+      type="file" 
+      accept="image/*,audio/*,text/csv" multiple 
+    />
   `)
 
   userEvent.upload(element, files)
 
-  expect(element.files[0]).toBeUndefined()
-  expect(element.files.item(0)).toBeNull()
+  expect(element.files).toHaveLength(3)
+})
+
+test('should not upload file with unaccepted format', () => {
+  const file = new File(['hello'], 'hello.png', {type: 'image/png'})
+  const {element} = setup('<input type="file" accept="image/jpg" />')
+
+  userEvent.upload(element, file)
+
   expect(element.files).toHaveLength(0)
+})
+
+test('should not upload multiple files with unaccepted formats', () => {
+  const files = [
+    new File(['hello'], 'hello.txt', {type: 'text/plain'}),
+    new File(['there'], 'there.pdf', {type: 'application/pdf'}),
+    new File(['there'], 'there.png', {type: 'image/png'}),
+    new File(['there'], 'there.mp4', {type: 'video/mp4'}),
+  ]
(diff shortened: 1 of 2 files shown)


LauraBeatris:
Still need to add a test case for the following MIME types: `image/*,video/*`

LauraBeatris:
> Still need to add a test case for the following MIME types: `image/*,video/*` , or in general, the ones that accept all types of extensions

Handled at a7c715a9f07800b6ac6b29fbe386a79f988b8e27

I've decided to use regex to validate cases where multiple images, videos, and audio formats are supplied, like "flags"

LauraBeatris:
Still need to improve in the validation logic for MIME types - After thinking a little bit more, it would be better to use another dependency to validate the file type. It would return code and make it easier to maintain.

ph-fritsche:
First of all: Thanks for this pull request :)

As the user is not able to select files that don't match the `accept` field, I'm not sure if any changes are necessary - if it is just up to the author of the test to be sensible about it.
userEvent methods are `"simulate this action by the user"` and we don't need to simulate actions that a user can not perform because the browser already prevents them from doing so.
But maybe it makes the life of the test author easier when reusing some fixtures and the necessary changes are small.
@nickmccurdy What do you think?

The changes in `upload` are too spread out over the function.
The change should be just one line at [line 16](https://github.com/testing-library/user-event/blob/ad427a6d8802e453bd66cdf72a4d29bfa786247a/src/upload.js#L16), applying the `filter` before the `slice`.

I don't think validating the types here is complex enough to command outsourcing it in another dependency.
Correct me if I'm wrong, but the only allowed variants with wildcard are:
- `type/*` which can be easily translated to ``new RegExp(`^${type}\/`) ``
- `*` (which is equivalent to not providing an `accept` attribute)

LauraBeatris:
@ph-fritsche 

> userEvent methods are "simulate this action by the user" and we don't need to simulate actions that a user can not perform because the browser already prevents them from doing so.

I really agree with that but, there are some situations worth discussing. For instance, if a button is disabled, the user won't be able to click on it, so the browser is already preventing the action - but `upload` also checks for the `disable` attribute.

Another situation is, the user may have 4 files and just one of them is not allowed to be selected, so the action will be performed anyway, in that way, I think that it's better if `upload` follows the same behavior. 

Let's also say that an engineer removes the `accept` attribute of an input component, or removes an specific file type - The tests would break.

LauraBeatris:
> I don't think validating the types here is complex enough to command outsourcing it in another dependency.
> Correct me if I'm wrong, but the only allowed variants with wildcard are:
> 
> type/* which can be easily translated to new RegExp(`^${type}\\/`)
> * (which is equivalent to not providing an accept attribute)

I think that `*` is not a valid value 🤔 , MDN just mentions `audio/*`,  `image/*` and `video/*`

ph-fritsche:
Let's follow the (living standard) spec: https://html.spec.whatwg.org/multipage/input.html#file-upload-state-(type=file)

* anything starting with a dot `.` - true if the [`File.name`](https://developer.mozilla.org/en-US/docs/Web/API/File/name) ends on it
* `audio/*`, `image/*`, `video/*` - true if the [`File.type`](https://developer.mozilla.org/en-US/docs/Web/API/File/type) starts with it (excluding the `*`)
* anything else - true if the `File.type` is equal

jeysal:
Could throwing or `console.error`ing (not sure if there's precedent for that in testing lib) be a way to get that benefit Laura mentioned (tests breaking when removing an accept attribute) while also not causing potentially hard to understand behavior where the tester may not know why some files are just being lost?

ph-fritsche:
@LauraBeatris I opened a PR with suggested changes here: https://github.com/LauraBeatris/user-event/pull/1

ph-fritsche:
> Could throwing or `console.error`ing (not sure if there's precedent for that in testing lib) be a way to get that benefit Laura mentioned (tests breaking when removing an accept attribute) while also not causing potentially hard to understand behavior where the tester may not know why some files are just being lost?

Removing an accept attribute would not break the test. Adding one would, but would also be the expected behavior of the test if you exclude a file per accept that you tried to upload.

jeysal:
> Removing an accept attribute would not break the test. Adding one would, but would also be the expected behavior of the test if you exclude a file per accept that you tried to upload.

Sorry, I worded that incorrectly - _removing a type from the list of accepted types_.
And yes, you would want the test to break if that type is being uploaded in it, that's the point.

Wdyt about throw/console.error? Might save some people from getting stuck trying to figure out why their uploads are silently ignored

ph-fritsche:
> And yes, you would want the test to break if that type is being uploaded in it, that's the point.

The test would break if it tests for the uploaded files to be applied in that input/change event.

> Wdyt about throw/console.error? Might save some people from getting stuck trying to figure out why their uploads are silently ignored

This would also mean that the user could not use `userEvent.upload` to test the `accept` attribute.
So there would be exactly the same situation as before this PR.

LauraBeatris:
I'll fix the coverage soon

src/__tests__/upload.js
```diff
@@ -163,3 +163,55 @@ test('should call onChange/input bubbling up the event when a file is selected',
   expect(onInputInput).toHaveBeenCalledTimes(1)
   expect(onInputForm).toHaveBeenCalledTimes(1)
 })
+
+test('should upload file with accepted format', () => {
```
ph-fritsche, line 167: Minor style issue: The tests in this repo follow the `imperative` style. So just `upload` instead of `should upload`.

LauraBeatris: But the other test cases that were already this file are using `should` 🤔

LauraBeatris: It's not just in this file btw, there are a lot of other tests mixing imperative style with declarative style - Should we fix that in a separate PR?

ph-fritsche: Just checked it: Actually nothing mentioned in the guideline. Guess it's just that everybody used that `imperative` style around here recently. I guess everybody just brings their style and each suite reflects who contributed most/first to it.

LauraBeatris: I've actually have sent a PR to fix that but I messed up and created a branch with the changes of this PR 🤕 

I'll submit another one with isolated changes and we can discuss it there

LauraBeatris:
@ph-fritsche About the test style, I think it's better to create an issue for that so we can discuss it there & then raise a PR

ph-fritsche (approved):
Looks good to me. I'll give others a few hours to weigh in on it before merging.

Thanks a lot for your contribution :)

ph-fritsche:
@all-contributors add @LauraBeatris code, tests

pleunv:
Do keep in mind that, even though an `accept` attribute might be set, users can easily circumvent this on both macOS & Windows by simply disabling the filter on the file upload dialog and selecting any file they want. Consequently, we have some tests in place that verify that the client-side validation captures this scenario correctly and still disallows uploading the file. These tests are now unfortunately broken as the upload is blocked as of this PR.

Is there any way to disable this new `accept` behavior?

ph-fritsche:
@pleunv Thanks for raising this. We should at least add an option here.

I'm at odds with myself if that option should be opt-in to test if an `accept` attribute is set correctly (as the user agent does not apply any filter here) or opt-out to test client-side filters like yours (as the user has to actively take a non-default path to get there and it depends on the file picker if he can get there)

@LauraBeatris Please weigh in on this.

pleunv:
We only use it in a handful of places so opt-in or opt-out doesn't matter much for us. That said, personally I'd be inclined to say it should be opt-in since `accept` cannot be enforced and tends to behave in ways people are not always aware about (i.e. mime-types). When it comes to testing & validation we treat it as not being there at all, since that's the behavior you get when a user simply bypasses the filter.

LauraBeatris:
@pleunv Thanks for reporting this. I agree with you, it should be opt-in and we should add a flag for this validation.

LauraBeatris:
> @pleunv Thanks for reporting this. I agree with you, it should be opt-in and we should add a flag for this validation.

And in that case, I think that it should start as `false`

PR #559, opened 2021-02-19
style(tests): apply imperative styles to all tests to be consistent 

LauraBeatris:
**What**:

**Why**: As mentioned by @ph-fritsche in https://github.com/testing-library/user-event/pull/558, this repo follows the imperative style. There are files using a mix of `declarative` and `imperative` test descriptions, we should be consistent and use `imperative` style in all tests.

**How**: Replace `declarative` style with `imperative` style 

**Checklist**: 

- [ ] Documentation
- [ ] Tests
- [ ] Typings
- [x] Ready to be merged


LauraBeatris:
Is there a way that we can avoid contributors to use declarative style? For instance, some kind of lint rule of leaving it explicit in the docs

ph-fritsche:
Just checked it: Actually nothing mentioned in the guideline. Guess it's just that everybody used that imperative style around here recently. I guess everybody just brings their style and each suite reflects who contributed most/first to it.

But I like the changes. I think the reports are better to read. And it's consistent with the commit messages where we squash and apply imperative style.
So a fix and the test it brings with it should sound at least about the same.

ph-fritsche:
> Is there a way that we can avoid contributors to use declarative style? For instance, some kind of lint rule of leaving it explicit in the docs

On a second thought:
Ruling out declarative style might need a lot of discussion.
So `it('clicks some object')` will have to stay.
Getting rid of superfluous `it('should click some object')` on the other hand might be something everybody could get behind and that is easier to enforce in linting.

PR #562, opened 2021-02-23
feat(upload): accept filter option

ph-fritsche:
**What**:

Add `applyAccept` option to `userEvent.upload` that defaults to `false`.

**Why**:

Filtering the files for the the input causes problems for those testing their own filters.
https://github.com/testing-library/user-event/pull/558#issuecomment-783397519

Those who want to test their accept attributes can opt-in per:
```js
  userEvent.upload(element, files, undefined, { applyAccept: true })
```

**How**:

Add an options parameter to the API.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Typings
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,11 +271,14 @@ test('types into the input', () => {
 })
 ```
 
-### `upload(element, file, [{ clickInit, changeInit }])`
+### `upload(element, file, [{ clickInit, changeInit }], [options])`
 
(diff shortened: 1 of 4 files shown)


ph-fritsche:
@LauraBeatris @pleunv

LauraBeatris:
It's all good from my side regarding the filter option

pleunv:
:shipit:

Issue #566, opened 2021-02-27
click handler should not fire blur when the click target inside currently focused element


just-boris:
- `@testing-library/user-event` version: 7.29.6 (latest at the time of writing)

- Testing Framework and version: n/a
- DOM Environment: Tested in Chrome/Safari/Firefox

### Steps to reproduce
1. Open this CodeSandbox: https://rvh87.csb.app/
2. Click on "click with user-event" button
3. You will see "blur" logs in the console
4. Click on "test button" directly
5. There are no "blur" logs in the console

### Code Example:

Codesanbox source: https://codesandbox.io/s/rvh87

```html
<button class="to-test">
  <span>test button</span>
</button>
```

```js
userEvent.click(screen.getByText("test button"));
```

Suggested solution:

The library should check if the click target is inside already focused container and should not blur it.

ph-fritsche:
Please share the link to the codesandbox as you see it in your address bar / per click on "Share"->"Copy sandbox link" not just the browser output.

The reproduction in the sandbox serves as an easily accessible reference point for everybody running into this issue or working on it.

just-boris:
Updated the original comment

ph-fritsche:
The problem seems to be that `userEvent` tries to move the focus to the clicked element (here the `<span>`).
Since that is not focusable the focus is moved to the `document.body`.
https://github.com/testing-library/user-event/blob/7a5c51e7f89c4ab0592174eff591b9e9275ff81c/src/click.js#L68

I think the solution is to check if one of the parents is focusable and move the focus there.

ph-fritsche:
@all-contributors add @just-boris bug

PR #567, opened 2021-02-28
fix(click): focus closest focusable (#566)

ph-fritsche:
**What**:

Focus closest focusable ancestor of the clicked element.

**Why**:

See #566 

**How**:

Only `blur()` if the clicked element does not have a focusable ancestor.

**Checklist**:

- n/a Documentation
- [x] Tests
- n/a Typings
- [x] Ready to be merged

### DIFF
--- a/src/click.js
+++ b/src/click.js
@@ -61,10 +61,7 @@ function clickElement(element, init, {clickCount}) {
       element,
       getMouseEventOptions('mousedown', init, clickCount),
     )
-    if (
-      continueDefaultHandling &&
-      element !== element.ownerDocument.activeElement
-    ) {
+    if (continueDefaultHandling) {
       const closestFocusable = findClosest(element, isFocusable)
       if (previousElement && !closestFocusable) {
         blur(previousElement, init)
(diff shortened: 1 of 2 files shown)


ph-fritsche:
@just-boris

src/click.js
```diff
@@ -65,10 +65,11 @@ function clickElement(element, init, {clickCount}) {
       continueDefaultHandling &&
       element !== element.ownerDocument.activeElement
     ) {
-      if (previousElement && !isFocusable(element)) {
+      const closestFocusable = findClosest(element, isFocusable)
+      if (previousElement && !closestFocusable) {
```
just-boris, line 69: What will happen when `previousElement` and `closestFocusable` is pointing to the same element?

ph-fritsche: `focus` is safe to call
https://github.com/testing-library/user-event/blob/7a5c51e7f89c4ab0592174eff591b9e9275ff81c/src/focus.js#L6-L7

ph-fritsche: I removed the check above. It is redundant and gives the impression we'd have to guard the call to `focus`.

just-boris: With this implementation it makes sense, thank you for the explanation!

just-boris (approved):

just-boris (commented):
I also downloaded the patch and tested inside our project. The fix resolves our original issue

PR #569, opened 2021-03-02
fix(typings): add home and end to specialChars enum

jKatt:
**What**: Added home and end to the specialChars enum in typings

**Why**: In PR #536 home and end typings were not added, so `specialChars.home` could not be used with typescipt

**How**: added the missing typings

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Typings
- [x] Ready to be merged


ph-fritsche:
Thanks a lot for taking the time to contribute to this library :)

@allcontributors add @jKatt code

Issue #574, opened 2021-03-04
Upload with empty filelist no longer fires event (breaks simulation of "user closes file dialog")


schoeneu:
- `@testing-library/user-event` version: 12.8.1

- Testing Framework and version: jest 26
  
- DOM Environment: jsdom 16.4.0
  

**Sample code that breaks:**

```javascript
    user.upload(fileChooser, new File([validInputFile], 'test.txt'));
    await waitFor(() =>
      expect(logicThatTriggersOnChange).toBeCalled(),
    );

    user.upload(fileChooser, []);
// this no longer succeeds.
    await waitFor(() =>
      expect(logicThatTriggersOnChange).toBeCalled(),
    );
```

**Problem description:**

12.8.1 no longer fires an event when triggering `upload` with an empty file list. This means I can't simulate the user interaction "user opens file dialog, then aborts without selecting a file" anymore.

Probably caused by https://github.com/testing-library/user-event/pull/562/files#diff-0d02b9dbce9a4eb9b8e939cd4606e84f9348a9a289903a87636ecc6ad25ad413R22 in #562 

The PR in question explicitly tests this behaviour and wants no event to be fired: https://github.com/testing-library/user-event/pull/562/files#diff-96272a98d39b469bbb095ed9a8be60a76ff04515be29afad4cdc45814d640d1eR195

... but my tests on both Firefox and Chrome show that they do fire an event (and go back to "no file selected") when a file was selected previously and then the upload dialog is opened again and aborted afterwards.

**Suggested solution:**

Reintroduce firing an "empty" event.

ph-fritsche:
Thanks a lot for this report. Looks like our investigation on this matter was not thorough enough.

The browser does not fire an input event when the file dialog is closed at first.
But looks like us fixing one bug introduced another here:

It seems it does not matter if a user closes the file dialog, but if his selection changes the value of the file input.
When selecting the same file(s) - regardless of the selection being multiple, single or none - no input event is fired.
Aborting the selection clears an already existing FilesList and triggers an `input` event.

https://codesandbox.io/s/busy-fermat-4ok6l?file=/src/App.test.js

PR #575, opened 2021-03-05
fix(upload): fire input event when files change (#574)

ph-fritsche:
**What**:

Do not fire an input event if the user selects the same files again.
Do fire an input event if the user selects no files when files have been selected before.

**Why**:

See #574 

**How**:

Compare the `files` provided by the user with `input.files`.

**Checklist**:

- N/A Documentation
- [x] Tests
- N/A Typings
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/upload.js
+++ b/src/__tests__/upload.js
@@ -192,10 +192,30 @@ test.each([
   },
 )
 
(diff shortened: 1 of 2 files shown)


ph-fritsche:
@schoeneu

Issue #521, opened 2020-12-14
Rework userEvent.type


ph-fritsche:
We've got multiple open issues regarding the implementation of `userEvent.type`.
And while a few people offered help on this, I've got the impression we're a bit lost regarding what the code should do and look like in general as well as how and where each smaller problem should be addressed.
I propose we collect those issues, discuss what we want to achieve and refactor and comment the code so that we get more confidence working on this.

- [ ] Typing on contenteditable #442 
- [x] Referencing characters by `keyEvent.code` #456 
- [x] Handling activeElement changing to another element #460 / #487 
- [ ] Preserving input over value sanitization #479 
- [x] Typing on non-input elements #481 , #506 
- [ ] Selecting on non-text input elements #501, #515 
- [ ] Default behavior on clickable elements #554 
- [x] Prevent default behavior and missing events on navigation keys #578 

Should I have overlooked something, please mention it and I'll add it to the list.

What do people expect `userEvent.type` to do? What should it do?
While `userEvent.click` is very clear and intuitive to use as `userEvent click this element!`,
I guess `userEvent.type` started the same way `userEvent trigger the key events to add to the value of this element!`,
but I think most people expect it to be more `userEvent press those buttons on the keyboard!`.

I think it should resolve around `document.getSelection()` and `userEvent.type(null, "a{del}{shift}b{tab}c{/shift}[Semicolon]d{selectall}")` should be a valid expression to test handling of hot keys and similar.

One problem that comes to my mind is how to handle the keypress events as they are triggered multiple times as long as the button is pressed. Maybe a configurable random timeout between key events? So that even an expression for `Press [A] and then press [B] before releasing [A]. Then release [B].` would make sense.

sethreidnz:
This makes a lot of sense to me, I am keen to try to pick up some part of this as I mentioned yesterday but digging into the code, it is very much written with inputs in mind and it is a little unclear the best way to improve this... This looks like a great summary!

ph-fritsche:
@kentcdodds Could you weigh in on this before we start any work here?
> What do people expect `userEvent.type` to do? What should it do?
> [...] I think most people expect it to be more `userEvent press those buttons on the keyboard!`.
> 
> I think it should resolve around `document.getSelection()` and `userEvent.type(null, "a{del}{shift}b{tab}c{/shift}[Semicolon]d{selectall}")` should be a valid expression to test handling of hot keys and similar.

kentcdodds:
[...]

PR #581, opened 2021-03-09
Rework type

ph-fritsche:
**What**:

This PR will include a complete rewrite of `userEvent.type` to tackle the issues tracked in #521 

There are a lot of changes and some of the fixes along the way might break code that considered these bugs features. So a major version update should happen.

With the changes `userEvent.type` will actually return no Promise if there is no `delay`.

The newly introduces `userEvent.keyboard` skips all convenience of `userEvent.type` and just applies the key events on the active element.
It allows to apply events and continue with a keyboard state.

https://github.com/testing-library/user-event/blob/b8a78c8f9dc9d19b0ef57ffdeef8e4af8d62f3a8/src/__tests__/keyboard/index.ts#L111-L137

**Why**:

`userEvent.type` currently is a patchwork of specialized implementations for edge cases.
We've got a bunch of open issues that nobody wants to tackle because it is unclear where exactly a change should be implemented and if there are more things to consider.
The code lacks structure and documentation.

The conversion to Typescript highlighted a lot of hidden debt inside the code regarding assumptions about return values passed around without further checks.

**How**:

- [x] First Step: Split `utils` and `type` into multiple files and convert to Typescript to improve confidence in upcoming changes
- [x] Refactor the `typeImplementation` to provide a shared internal API to extend for special characters, meta characters, etc
- [x] Convert handler implementations to new plugins
- [x] Add tests for uncovered lines
- [x] Introduce new API
- [x] Add documentation

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Typings
- [x] Ready to be merged


ph-fritsche:
@adamayres The rewrite is complete. All previous tests (apart from a) a bug and b) an unlogic edge case) pass.
Your bug should be fixed by the new implementation too.

eps1lon:
I'm not familiar with `user-event` and don't intend to use. I wouldn't have the time to review this change.

README.md
```diff
@@ -271,6 +268,76 @@ test('types into the input', () => {
...
+  ```
+  This does not keep any key pressed. So `Shift` will be lifted before pressing
+  `f`.
+- Per
+  [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)
+  ```js
+  userEvent.keyboard('[ShiftLeft][KeyF][KeyO][KeyO]') // translates to: Shift, f, o, o
+  ```
+- Per legacy `userEvent.type` modifier/specialChar The modifiers like `{shift}`
+  (note the lowercase) will automatically be kept pressed, just like before. You
```
kentcdodds, line 305: Interesting that there's inconsistency in whether it's upper or lower case. Any chance we could make it just not matter and it can be cased however folks want?

README.md
```diff
@@ -271,6 +268,76 @@ test('types into the input', () => {
...
+  userEvent.keyboard('{Shift}{f}{o}{o}') // translates to: Shift, f, o, o
+  ```
+  This does not keep any key pressed. So `Shift` will be lifted before pressing
+  `f`.
+- Per
+  [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)
+  ```js
+  userEvent.keyboard('[ShiftLeft][KeyF][KeyO][KeyO]') // translates to: Shift, f, o, o
+  ```
+- Per legacy `userEvent.type` modifier/specialChar The modifiers like `{shift}`
```
kentcdodds, line 304: Calling it "legacy" makes it sound like we're recommending *against* using `userEvent.type`. If I understand these changes correctly, that's not necessarily the case. Could do:

```suggestion
- Per `userEvent.type`-like modifier/specialChar modifiers like `{shift}`
```

src/__tests__/keyboard/getNextKeyDef.ts
```diff
@@ -0,0 +1,138 @@
+import {getNextKeyDef} from 'keyboard/getNextKeyDef'
```
kentcdodds, line 1: Just an idea, not necessarily a recommendation. I think this could be a good place for `jest-in-case` as lots of these tests look a lot like each other.

src/__tests__/keyboard/index.ts
```diff
@@ -0,0 +1,137 @@
...
+    body - keyup: o (111)
+    body - keydown: o (111)
+    body - keypress: o (111)
+    body - keyup: o (111)
+  `)
+})
+
+it('type with focus', () => {
+  const {element} = setup('<input/>')
+  const {getEventSnapshot} = addListeners(document.body)
```
kentcdodds, line 28: Curious why this isn't:

```suggestion
  const {element, getEventSnapshot} = setup('<input/>')
```

src/__tests__/type.js
```diff
@@ -792,9 +792,7 @@ test('typing an invalid input value', () => {
   const {element} = setup('<input type="number" />')
   userEvent.type(element, '3-3')
 
-  // TODO: fix this bug
-  // THIS IS A BUG! It should be expect(element.value).toBe('')
-  expect(element).toHaveValue(-3)
+  expect(element).toHaveValue(null)
```
kentcdodds, line 795: HOORAY! A TODO that's actually been addressed. It's a miracle 😆

src/keyboard/getEventProps.ts
```diff
@@ -0,0 +1,27 @@
...
+export function getKeyEventProps(keyDef: keyboardKey, state: keyboardState) {
+  return {
+    key: keyDef.key,
+    code: keyDef.code,
+    altKey: state.modifiers.alt,
+    ctrlKey: state.modifiers.ctrl,
+    metaKey: state.modifiers.meta,
+    shiftKey: state.modifiers.shift,
+
+    // deprecated
```
kentcdodds, line 12: If we use the `@deprecated` annotation then vscode/TypeScript will offer some suggestions :)

```suggestion
    /** @deprecated use code instead */
```

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAcwKZQNKoJ4FEBuqYUACgE5wAOAzgBQCUiA3gLABQiiZ6IZSrHTogDWOAFyIA5KOwARVMAB0MyQBp2QxBDgATVBOk55S7XrUahAQwA2mcVOpRLUVIoC2umMBioy1RTZQ5oKc0GTWWNgGjs6uHjpePn6KYdbBmm7olpHRTi7unt6+-plO6ULUABZedlEOeXGFSf5VNcEWnAD0AFTdiAACepTcELE6iCDUqFq60zBgjqiW492dHSI4AMKzBtWSFgC+7Eds7GenbNoLUFyo1IgAvCjokQREpBQ0DOfc-jLbeiAA

image

src/type/typeImplementation.ts
```diff
@@ -0,0 +1,89 @@
+// TODO: wrap in asyncWrapper
```
kentcdodds, line 1: I think this is accomplished in `src/type/index.ts` right?

```suggestion
```

src/type/typeImplementation.ts
```diff
@@ -0,0 +1,89 @@
...
+      initialSelectionEnd ?? value.length,
+    )
+  }
+
+  // previous implementation did wait before the first character
+  // this probably can be removed
+  if (delay > 0) {
+    await wait(delay)
+  }
+
```
kentcdodds, line 76: Agreed, if we're breaking things anyway we may as well remove this which doesn't have much reason to exist.

```suggestion
```

kentcdodds (approved):
Looking great! Fantastic work 👏👏👏

A few general comments:

1. I typically avoid camelCasing filenames due to inconsistencies in operating systems with filename casing. I use kebab-case instead
2. There are a lot of tiny files. I've done things that way in the past and I've found that things are simpler when you only split things out when necessary. I'm mostly thinking about the utils/misc stuff.

In any case, you're the primary maintainer at this point, so you can think of my comments/questions as ideas and suggestions and you're more than welcome to ignore any and all of them :)

With that said, I'll give you my green stamp of approval because I think what you have here is awesome. Thank you for all you do!

README.md
```diff
@@ -271,6 +268,76 @@ test('types into the input', () => {
...
+Keystrokes can be described:
+
+- Per printable character
+  ```js
+  userEvent.keyboard('foo') // translates to: f, o, o
+  ```
+  The brackets `{` and `[` are used as special character and can be referenced
+  by doubling them.
+  ```js
+  userEvent.keyboard('{{a[[') // translates to: {, a, [
```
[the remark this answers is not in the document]
ph-fritsche: I'm not sure what I prefer. Escaping in strings is always nasty. I just went forward interpreting the string as when encountering `{` or `[` treat the following string in a special way. A little bit like some template languages like Twig.

README.md
```diff
@@ -271,6 +268,76 @@ test('types into the input', () => {
...
+- Per
+  [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)
+  ```js
+  userEvent.keyboard('[ShiftLeft][KeyF][KeyO][KeyO]') // translates to: Shift, f, o, o
+  ```
+- Per legacy `userEvent.type` modifier/specialChar The modifiers like `{shift}`
+  (note the lowercase) will automatically be kept pressed, just like before. You
+  can cancel this behavior by adding a `/` to the end of the descriptor.
+  ```js
+  userEvent.keyboard('{shift}{ctrl/}a{shift}') // translates to: Shift(down), Control(down+up), a, Shift(up)
```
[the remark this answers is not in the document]
ph-fritsche: Good catch. Thanks

README.md
ph-fritsche: Sadly not. As the previous behavior was different for different keys.
The new implementation is consistent for all keys and adds an exception for the old modifiers.

README.md
ph-fritsche: No, but I would recommend against the old modifiers as they behave differently for different keys.

src/type/typeImplementation.ts
ph-fritsche: I was not sure what this TODO was referring to. Might have been accomplished a long time ago.

src/__tests__/keyboard/index.ts
ph-fritsche: Because `keyboard` does not move the focus.
If the focus is not on the `input`, there are no events on the input.
The eventSnapshot on `document.body` tells the full story for these tests and provides a meaningful error message both for wrong events and wrong event targets.

src/__tests__/keyboard/getNextKeyDef.ts
ph-fritsche: I implemented it. Not sure if I like the look of the test file. But it did get rid of some repetition. :)

ph-fritsche:
> 2. There are a lot of tiny files. I've done things that way in the past and I've found that things are simpler when you only split things out when necessary. I'm mostly thinking about the utils/misc stuff.

Yes, that might have been overkill. But I went crazy jumping through code when figuring out what interacts / is used where and what to reuse/rewrite/remove. So I just went ahead and split everything up before doing any changes.
I think when restructuring code it's a lot easier to deal with a bunch of files than multiple functions in different positions in a shared file with others you don't want to touch at that moment. But it might be just personal preference.

README.md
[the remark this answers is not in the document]
kentcdodds: Sounds fine to me. I agree escaping in strings is really annoying and using `\` as the mechanism just makes it more challenging. This is probably fine +

README.md
kentcdodds: Makes sense 👍

src/__tests__/keyboard/index.ts
kentcdodds: Ah, yeah, that makes sense 👍

kentcdodds (approved):
Thanks for answering my questions and for all your hard work on this project. It's fantastic 👍 

You've got a big green ✅ from me!

PR #588, opened 2021-03-17
fix(keyboard): parse modifiers as characters

### LINKED ISSUE #587: bug: modifiers as printable characters
When typing a `:` or a `/`, it is also escaped/removed from the input value.
See: https://codesandbox.io/s/user-event-type-url-sc8by?file=/src/__tests__/App.test.js

Escaping them with the provided regex helps with the `:` but not with the `/`

_Originally posted by @SiebeVE in https://github.com/testing-library/user-event/issues/586#issuecomment-800990846_

ph-fritsche:
**What**:

Fix parsing modifiers as printable characters.

**Why**:

Closes #587 

**Checklist**:

- N/A Documentation
- [x] Tests
- N/A Typings
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/keyboard/getNextKeyDef.ts
+++ b/src/__tests__/keyboard/getNextKeyDef.ts
@@ -19,6 +19,16 @@ cases(
           key,
           code,
         }) as keyboardKey,
+        consumedLength: text.length,
+      }),
+    )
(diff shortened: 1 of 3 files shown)


SiebeVE:
I didn't know we got to travel at light speed yet! That was so fast I couldn't even say supercalifragilisticexpialidocious!

PR #595, opened 2021-03-18
Support empty string for maxlength

Aprillion:
**What**:

An input's maxlength can be an empty string, which should NOT limit typing.

**Why**:

Browsers work that way.

**How**:

~~Use `||` instead of `??` operator to convert "" into -1 instead of 0.~~

ph-fritsche  introduced a new function `getSanitizedMaxLength` to handle multiple edge cases

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] Typings N/A
- [x] Ready to be merged


src/utils/edit/calculateNewValue.ts
```diff
@@ -16,7 +16,10 @@ export function calculateNewValue(
 
   // can't use .maxLength property because of a jsdom bug:
   // https://github.com/jsdom/jsdom/issues/2927
-  const maxLength = Number(element.getAttribute('maxlength') ?? -1)
+  //
+  // '' is a valid value in DOM with the same meaning as -1, not 0 => use ||
+  // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+  const maxLength = Number(element.getAttribute('maxlength') || -1)
```
ph-fritsche, line 22: Looks like `maxLength` is only used at line 78ff. We should move it there.

Also [MDN states](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength):
> This must be an integer value 0 or higher. If no maxlength is specified, or an invalid value is specified, the input or textarea has no maximum length.

Let's sanitize it for other cases than `""` too.

ph-fritsche (changes requested):
Opened a PR with code suggestions at https://github.com/Aprillion/user-event/pull/1

src/utils/edit/calculateNewValue.ts
Aprillion: good point, now merged

ph-fritsche (approved):

Issue #583, opened 2021-03-16
typing fails if input matches selection


ph-fritsche:
I have found what I believe to be a related issue to this one, which is documented in this CodeSandbox: https://codesandbox.io/s/userevent-unit-test-ncmgu?file=/src/App.test.js

Setup of the tested component
1. contains a text input with an initial value of 1 string character (ex: `'1'`)
2. executes the `HTMLInputElement.select()` method on input focus so every new value completely overwrites the previous one

Test execution:
1. Render the tested component with input
2. Get the input by role
3. Change the input value via `userEvent.type(...` (ex: `'11123'`)
4. Notice the received input value is `'23'` and not `'11123'`

_Originally posted by @psullivan6 in https://github.com/testing-library/user-event/issues/521#issuecomment-796888618_

ph-fritsche:
This is probably caused by `fireInputEventIfNeeded`. The key events fire as expected, the `input` event is missing.

fergusmcdonald:
Hi @ph-fritsche 👋  

What is reasoning behind not firing input events if `newValue` is the same as `prevValue` in `fireInputEventIfNeeded`?  https://github.com/testing-library/user-event/blob/master/src/keyboard/shared/fireInputEventIfNeeded.ts#L27-L32

I can see an input event is fired when replicating the issue in the sandbox.

https://user-images.githubusercontent.com/3115675/111896397-8efec500-89d6-11eb-8235-53a99aad78b1.mp4

ph-fritsche:
@fergusmcdonald Thanks for taking the time to look into this. :smiley: 

> What is reasoning behind not firing input events if `newValue` is the same as `prevValue` in `fireInputEventIfNeeded`? https://github.com/testing-library/user-event/blob/master/src/keyboard/shared/fireInputEventIfNeeded.ts#L27-L32
> 
> Is this a generalization which is mostly true for how the browser decides to fire `input` events? 

This is one of the parts of which I'm not sure yet, why they were implemented the way they are.

> Removing the rule will result in input events being fired in many scenarios where they shouldn't (see snapshots of commit referenced below).

When rewriting the implementation for `userEvent.type` I tried to maintain the behavior that was established before, unless it was already determined to be a bug caused by the previous implementation of `userEvent.type`.
One of the reasons for the rewrite was that it should allow us to isolate concerns like the one tackled by `fireInputEventIfNeeded` and verify if the expectations laid out by the existing tests are correct.

[...]

PR #607, opened 2021-03-22
fix(keyboard): fix fireInputEventIfNeeded (#583)

fergusmcdonald:
**What**: #583 

When a text input with a single character value is selected, typing text
which starts with the same character results in the first character of
the typed text not to be typed.

This occurs as fireInputEventIfNeeded does not fire an input event due
to the new value matching the previous value of the input. The next
character is then typed and the selected text is then overwritten.

This fix changes the fireInputEventIfNeeded condition from no event
being fired when the previous and new values match to an event being
fired in this scenario when the element value has a selection.

**Why**:

To support the scenario in #583 to match expected events.

**How**: 

The sandbox shows that the scenario outlined in the issue and test 
should fire an input event which leads to the selection being 
replaced so that the remainder of the text can be typed. This 
change adds that input event to be fired and does not impact any
of the existing tests or snapshots.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Typings
- [x] Ready to be merged


fergusmcdonald:
@ph-fritsche - This draft is an attempt at fixing where typing the same character as highlighted into a text field was not firing an input event. This had resulted in the selection persisting and the first character to be typed to be overwritten by the second. 

I'm not sure what confidence levels the existing test passing and snapshots not changing gives for a change to made. A review would be greatly appreciated. 🙏

ph-fritsche:
@fergusmcdonald First of all: Thanks a lot for taking the time to contribute. :heart: 

I would like to avoid introducing more workarounds for edge cases when we might be able to achieve the desired results by removing/refactoring code.

There are six references of `fireInputEventIfNeeded`. I'm not sure they really have a common requirement when calling this function. We might be able to move (some of) the conditions directly into the handlers that require them.

Instead of checking `isClickableInput` inside `fireInputEventIfNeeded`, we should only apply the editing behavior described for `keypress` when the element is editable. Same for `isReadonly`.
This might be achieved by a `isEditable` check inside `behaviorPlugin.matches`.

The comparison with the previous value obviously follows a flawed premise. We should remove it.
We need to verify if the `date` and `time` input require this or a similar check. If they do, we should add it there.
We need to add new checks in the behavior for `[Backspace]` and `[Delete]` if the cursor is at position `0` respectively `length`.

This reduces `fireInputEventIfNeeded` to `fireInputEvent` and it only applies `target.value` or `target.textContent` depending on `isContentEditable`.

Issue #512, opened 2020-12-02
Paste eventInit Type should be ClipboardEventInit


dvargas92495:
- `@testing-library/user-event` version: 12.3.0

- Testing Framework and version: Not Relevant
- DOM Environment: Not Relevant

Relevant code or config
https://github.com/testing-library/user-event/blob/master/typings/index.d.ts#L65

What you did:

When using user-event in typescript, I wanted to pass in `clipboardData` into the `eventInit` arg of `paste`, as directed by the `README`: https://github.com/testing-library/user-event/blob/master/README.md#pasteelement-text-eventinit-options

What happened:

I get a type error indicating that `clipboardData` is not found on `MouseEventInit`

Reproduction repository:

I'd be happy to put a repro repo, but the issue and fix seem straightforward.

Problem description:

The `eventInit` arg in `paste` is of type `MouseEventInit`.

Suggested solution:

The `eventInit` arg in `paste` should be type `ClipboardEventInit`. (https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.clipboardeventinit.html)

nickmccurdy:
How did you create `clipboardData`? Are you using a raw JS object? JSDOM doesn't implement `ClipboardData` (see https://github.com/jsdom/jsdom/issues/1568).

I'm tracking support for custom implementations of `ClipboardData` and `DataTransfer` in https://github.com/testing-library/user-event/issues/355#issuecomment-642941030. Would either of these fix your issue or are you looking for something else?

dvargas92495:
I just need the type fixed. I'm actually using the library in a hacky fashion in the browser haha

https://github.com/dvargas92495/roam-js-extensions/blob/master/src/components/WikiData.tsx#L82

ph-fritsche:
A PR correcting typings would be very much welcome :)

fergusmcdonald:
@ph-fritsche - I've attempted to create a type fix PR in #611

PR #611, opened 2021-03-23
fix(paste): correct the paste event type (#512)

fergusmcdonald:
**What**: Fix the event init type of paste by changing from MouseEventInit to ClipboardEventInit. This aligns with the type definition of fireEvent for pasting: https://github.com/testing-library/dom-testing-library/blob/master/src/event-map.js#L11-L14

**Why**: Types should be correct.

**How**: N/A

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Typings
- [x] Ready to be merged

### DIFF
--- a/src/paste.ts
+++ b/src/paste.ts
@@ -14,7 +14,7 @@ interface pasteOptions {
 function paste(
   element: HTMLInputElement | HTMLTextAreaElement,
   text: string,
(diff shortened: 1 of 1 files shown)


ph-fritsche:
@allcontributors add @fergusmcdonald code

PR #622, opened 2021-03-24
fix: apply changes before input event

### LINKED ISSUE #515: Incorrect event.target.selectionStart value when using left arrow and backspace keys
- `@testing-library/user-event` version: 12.4.0
- Testing Framework and version: Jest (using `react-scripts`, version: 3.4.3)
- DOM Environment: `@testing-library/jest-dom` (5.11.6)

**Relevant code or config**
I'm using `useState` here just to be able to display the `event.target.selectionStart` value and demonstate the issue.
```javascript
// component
export default function App() {
  const [selectionStart, setSelectionStart] = useState(0);

  const onChangeHandler = (event) => {
    setSelectionStart(event.target.selectionStart);
  };

  return (
    <div className="App">
      <input onChange={onChangeHandler} data-testid="input" />
      <code>
        event.target.selectionStart:{" "}
        <span data-testid="selectionStartValue">{selectionStart}</span>
      </code>
    </div>
  );
}

```
```javascript
// test
userEvent.type(input, "123{arrowleft}{arrowleft}{backspace}");

const selectionStartValue = screen.getByTestId("selectionStartValue");
expect(selectionStartValue).toHaveTextContent("0"); // it fails - the value is 2 instead
```

**What you did:**
I discovered it when writing a Jest test for a React component which is an input reacting to `onChange` and relying heavily on the `event.target.selectionStart` value passed in this event. I used `userEvent.type` in the test.

**What happened:**
I noticed that when the cursor is not at the end of the text in the input and backspace is pressed, the value of `event.target.selectionStart` provided by `userEvent.type` is different from the value I get in my browser.

In order to see a more tangible example, please see the CodeSandbox project I created and linked below and follow these steps:
1. Type "abc" in the input
2. Hit left arrow button on your keyboard twice
3. Hit backspace
4. Check the value of `event.target.selectionStart`

Result in the browser: 0
Result in the test (App.test.js): 2

**Reproduction repository:** 
https://codesandbox.io/s/user-event-and-backspace-selectionstart-nx6ij?file=/src/App.test.js

**Problem description:**
I think the problem is related to `fireInputEventIfNeeded`. The value of `event.target.selectionStart` in `fireEvent` is incorrect (at least in my case) but it's correct in the `setSelectionRange` called at the end of the function (line 319). Shouldn't it be called before the event is fired?
https://github.com/testing-library/user-event/blob/e9dd0a6d0413a3b6184a419b2daa77088290498a/src/type.js#L295-L324

**Suggested solution:**
As in the problem description. I'm happy to make a pull request if you confirm my reasoning is correct.

ph-fritsche:
**What**:

Apply new value and selection range before `input` event so that `input` handlers can use it.
Get correct selection range for `[ArrowLeft]` and `[ArrowRight]` when active element does not support selectionRange.

**Why**:

Closes #515 

**Checklist**:

- N/A Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/clear.js
+++ b/src/__tests__/clear.js
@@ -24,7 +24,6 @@ test('clears text', () => {
     input[value="hello"] - select
     input[value="hello"] - keydown: Delete (46)
     input[value=""] - input
-      "{SELECTION}hello{/SELECTION}" -> "{CURSOR}"
     input[value=""] - keyup: Delete (46)
   `)
 })
--- a/src/__tests__/keyboard/index.ts
+++ b/src/__tests__/keyboard/index.ts
@@ -130,7 +130,6 @@ it('continue typing with state', () => {
     input[value=""] - keydown: F (70) {shift}
     input[value=""] - keypress: F (70) {shift}
     input[value="F"] - input
-      "{CURSOR}" -> "F{CURSOR}"
     input[value="F"] - keyup: F (70) {shift}
     input[value="F"] - keyup: Shift (16)
   `)
--- a/src/__tests__/keyboard/plugin/functional.ts
+++ b/src/__tests__/keyboard/plugin/functional.ts
@@ -54,12 +54,10 @@ test('backspace to valid value', () => {
     input[value=""] - keydown: 5 (53)
     input[value=""] - keypress: 5 (53)
     input[value="5"] - input
-      "{CURSOR}" -> "{CURSOR}5"
     input[value="5"] - keyup: 5 (53)
     input[value="5"] - keydown: e (101)
     input[value="5"] - keypress: e (101)
     input[value=""] - input
-      "{CURSOR}5" -> "{CURSOR}"
     input[value=""] - keyup: e (101)
     input[value=""] - keydown: - (45)
     input[value=""] - keypress: - (45)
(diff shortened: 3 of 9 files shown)


stefcameron:
@ph-fritsche Hey! 👋 Thanks for your work on this awesome library! I'm trying to figure out why tests are breaking now, either after this or #619, though this one, based on the code changes (which I don't fully understand) feels like it's the more likely culprit.

So far, I've been using the React Testing Library, and using `user-event` to type into an `<input type="text">` and expecting it to cause the input's change event to fire, and a state update to my component. But now, nothing happens. There's no change event triggered.

Here's a simplification, which is breaking:
```javascript
const handler = jest.fn();
const { container } = render(<div><input type="text" onChange={handler} /></div>);
const inputEl = container.querySelector('input');
expect(inputEl).toHaveValue('');
userEvent.type(inputEl, 'red');
expect(inputEl).toHaveValue('red');
expect(handler).toHaveBeenCalled();               // <- fails, but should succeed
```

The above tests all pass under v13.0.10...

stefcameron:
To be clear, my test is breaking under 13.0.13, so it's some patch 11, 12, or 13, that's causing the issue.

ph-fritsche:
@stefcameron Yes, this fix caused #624 which is fixed in #626 . Update to the latest version and you should be good :rocket: :smiley:

stefcameron:
@ph-fritsche Nice! Glad it should be fixed already. I'll check this out in 13.0.15!

stefcameron:
Fixed, thank you!

PR #631, opened 2021-03-28
Prevent element to be clicked if pointer events are disabled for the element

### LINKED ISSUE #630: Clicking on elements having css property set as pointer-events to none still invoke onClick
- `@testing-library/user-event` version: 13.0.16

- Testing Framework and version: @testing-library/react": ^11.1.0
  
- DOM Environment: jest
  

Relevant code or config

``` 
function CustomButton({callback}) {
  return (
    <button style={{width: 100, height: 60, cursor:'pointer', pointerEvents: 'none'}} type="button" onClick={callback} data- 
 testid='react-link'>
      Learn React
    </button>
  );
}

export default CustomButton

// test case
const callback = jest.fn();

beforeEach(() => jest.clearAllMocks());
test("renders learn react link", () => {
  render(<CustomButton callback={callback} />);

  userEvent.click(screen.getByTestId(/react-link/));
  expect(callback).toHaveBeenCalledTimes(0); // should not invoke callback function and respects pointer-events
});

 FAIL  src/button.test.js
  ● renders learn react link

    expect(jest.fn()).toHaveBeenCalledTimes(expected)

    Expected number of calls: 0
    Received number of calls: 1

      10 |
      11 |   userEvent.click(screen.getByTestId(/react-link/));
    > 12 |   expect(callback).toHaveBeenCalledTimes(0); // should not invoke callback function and respects pointer-events
         |                    ^
      13 | });
```

Problem description: It should have passed and should not propagate events to click the element as on button element `pointer-events` are disabled that should prevent clicking on button or any other element.

Suggested solution: I went through the repo and prepared a solution to stop propagating the respective events in case the element is not clickable if `pointer-events` are disabled for the element. I am happy to create a PR for the same.

```
Inside click() method,
if(!isElementClickable(element)) return;

// Utility fn to check if element is clickable

import {getWindowFromNode} from '@testing-library/dom/dist/helpers'

export function isElementClickable(element: Element): boolean {
  const window = getWindowFromNode(element)

  for (
    let el: Element | null = element;
    el?.ownerDocument;
    el = el.parentElement
  ) {
    const pointerEvents = window.getComputedStyle(el).pointerEvents
    if (pointerEvents === 'none') {
      return false
    }
  }

  return true
}
```

MohitPopli:
**What**: Before firing click event on element, added check to validate if element is clickable.

**Why**: Though we trust browser to do it's job perfectly, but still if element have pointer-events disabled then it should not propagate any events or callbacks.

**How**: Created a utility function to check for `pointer-events` css value, if it's set to `none` then we should not click the element and return.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] Typings N/A
- [x] Ready to be merged
      

Closes #630


ph-fritsche (changes requested):
Thanks a lot for taking the time to contribute :heart: 

Could you rename the `isElementClickable` to `hasPointerEvents` and add it to `dblClick`, `hover` and `unhover` too?

We should also exclude the pointerEvents in `selectOptionsBase`.

MohitPopli:
> Thanks a lot for taking the time to contribute ❤️
> 
> Could you rename the `isElementClickable` to `hasPointerEvents` and add it to `dblClick`, `hover` and `unhover` too?
> 
> We should also exclude the pointerEvents in `selectOptionsBase`.

Sure. I'll update PR with your suggestions

src/select-options.ts
```diff
@@ -10,6 +10,7 @@ function selectOptionsBase(
   values: HTMLElement | HTMLElement[] | string[] | string,
   init?: MouseEventInit,
 ) {
+  if (!hasPointerEvents(select)) return
```
ph-fritsche, line 13: `selectOptions()` on `<select/>` is possible without pointer events. We don't want to break backward compatibility here.
Please apply it only to the pointer events.

ph-fritsche (changes requested):

src/select-options.ts
MohitPopli: Alright. If I understand correctly then we don't want to prevent events to propagate here. Rather I should revert changes from `select-options.ts` file. Am I right?

ph-fritsche: I think for now the solution is to skip the pointer events in [line 49ff](https://github.com/testing-library/user-event/blob/c2186493bb329b6e50fe4256df7921377e3a309a/src/select-options.ts#L49-L92) if the clicked element (select/option) should not have pointer events but keep the call to `selectOption()`

src/utils/misc/hasPointerEvents.ts
```diff
@@ -0,0 +1,18 @@
...
+export function hasPointerEvents(element: Element): boolean {
+  const window = getWindowFromNode(element)
+
+  for (
+    let el: Element | null = element;
+    el?.ownerDocument;
+    el = el.parentElement
+  ) {
+    const pointerEvents = window.getComputedStyle(el).pointerEvents
+    if (pointerEvents === 'none') {
```
ph-fritsche, line 12: > however, pointer events may target its descendant elements if those descendants have pointer-events set to some other value

```suggestion
    if (pointerEvents === 'auto') {
      return true
    } else if (pointerEvents === 'none') {
```

ph-fritsche (changes requested):

src/utils/misc/hasPointerEvents.ts
```diff
@@ -0,0 +1,18 @@
...
+      return false
+    }
```
ph-fritsche, line 14: > however, pointer events may target its descendant elements if those descendants have pointer-events set to some other value

```suggestion
    if (pointerEvents) {
       return pointerEvents === 'none' ? false : true
    }
```

ph-fritsche (changes requested):

MohitPopli:
I'm not sure if I understood use case here,
https://github.com/testing-library/user-event/pull/631/files#diff-a681e050ffe760739876a36e05c9cfb1bb1ce0b6dae27c54152e16b9f6fac2f4R12

As once I merged this suggestion, my existing test cases started failing that I added to validate that we should return and should not fire any other events. 

Please advise @ph-fritsche

src/select-options.ts
MohitPopli: I think this is what we wanted here to maintain behaviour,
https://github.com/testing-library/user-event/pull/631/files#diff-1031ae1312b9a49cb4a49f43fdd91f4d13537c7f6f31a7e3021d4ef7797a237aR48

src/utils/misc/hasPointerEvents.ts
```diff
@@ -0,0 +1,18 @@
...
+export function hasPointerEvents(element: Element): boolean {
+  const window = getWindowFromNode(element)
+
+  for (
+    let el: Element | null = element;
+    el?.ownerDocument;
+    el = el.parentElement
+  ) {
+    const pointerEvents = window.getComputedStyle(el).pointerEvents
+    if (pointerEvents && ['inherit', 'unset'].includes(pointerEvents)) {
```
ph-fritsche, line 12: Oops, dropped the `!` when adding the second exception.
```suggestion
    if (pointerEvents && !['inherit', 'unset'].includes(pointerEvents)) {
```

Those two values of `pointerEvents` don't change the behavior inherited from an ancestor. So like no value we continue traversing ancestors.
If the value is `none` we return `false`. If the value is any other value, it overrides any value of its ancestors and we return `true`.

Probably need to add a test case for this.
```html
<div style="pointer-events: none">
  <input style="pointer-events: initial"/>
  <input style="pointer-events: inherit"/>
</div>
```
The first `<input>` is clickable. The second one is not.

ph-fritsche (changes requested):

src/select-options.ts
ph-fritsche: See [line63](https://github.com/testing-library/user-event/blob/a72cc3f64c2b9c20808d3d5e2f7e64870d07cc34/src/select-options.ts#L63) and [line70](https://github.com/testing-library/user-event/blob/a72cc3f64c2b9c20808d3d5e2f7e64870d07cc34/src/select-options.ts#L70) the call to `selectOption` .
Those calls should not depend on pointer events.

The issue for another day might be to add keyboard events doing this selections. But for now I think it's sensible to just skip all pointer events and only select the options the user wanted to select.

ph-fritsche:
Have a look at https://github.com/MohitPopli/user-event/pull/1

ph-fritsche (approved):
Looks good to me now. Thanks a lot for your work on this :smiley: 
I will have another look later and merge it if nothing leaps out at me then. :rocket:

ph-fritsche:
@all-contributors add @MohitPopli bug, code, test

rbung:
Hello, 

I use [reactstrap](https://reactstrap.github.io) and test my components with RTL and userEvent and since this feature, I encounter an error when I want to test my dropdown. I made a codesandbox to reproduce my error : https://codesandbox.io/s/dropdown-userevent-mo7h1?file=/src/MyDropdown.tsx

It looks like the click is not triggered on my menu item. To make it work, I found a workaround by setting the pointer-events style to `auto`. But maybe we should consider the element to have pointerEvents if this property is not set ?

MohitPopli:
@rbung I tried to triage the issue on my system and everything seems to work fine with me without having to set `pointer-events` property explicitly on any of the control.

I have used below queries used to click on dropdown and then menu item,

`userEvent.click(screen.getByRole('button', {
    name: /dropdown/i
  })) // clicks on dropdown button`

`userEvent.click(screen.getByRole('menuitem', {
    name: /click me/i,
    hidden: false
  })) // clicks on menu item`

Playground link: https://testing-playground.com/gist/1e9acc0b446eba5d32d611b0475997b4/bd9fec04e960bf99a17138efda8327c03003b149

Let me know if you have any questions.

ph-fritsche:
@rbung Your problem is described in https://github.com/testing-library/user-event/issues/639#issuecomment-813379774
Your `menuitem` button renders inside a parent with `pointer-events: none`. The component removes this style property later.

Update `user-event` to [`v13.1.3`](https://github.com/testing-library/user-event/releases/tag/v13.1.3) and you can work around per:
```js
await waitFor(() => userEvent.click(screen.getByRole("menuitem")))
```

PR #642, opened 2021-04-06
remove unnecessary references to global document object

loreanvictor:
**What**:

Removing unnecessary references to global `document` object. There seems to be some care taken across the code to avoid such references when not necessary ([example](https://github.com/testing-library/user-event/blob/7143ae30353e0cc7f8f12e405763eaf2257ef03e/src/click.ts#L18)), but some cases seem to be missed ([example](https://github.com/testing-library/user-event/blob/7143ae30353e0cc7f8f12e405763eaf2257ef03e/src/click.ts#L112)).

**Why**:

Well I ran into an issue using this library, because generally I avoid registering a global document object in my tests. However it was particularly easy to fix since generally the code is careful not to reference global document object, and only a few places were mistakenly making references.

Note that in some situations referencing global document object is necessary ([example](https://github.com/testing-library/user-event/blob/master/src/tab.ts)) if additional parameters (from which reference document can be inferred) are not provided.

**How**:

I could only find such references in code related to [`click()`](https://github.com/testing-library/user-event/blob/master/src/click.ts) and [`tab()`](https://github.com/testing-library/user-event/blob/master/src/tab.ts).

In case of `click()`, there is a reference element whose owner document can be used instead of the global object. So I just switched to that.

In case of `tab()`, if a focus trap is not provided, then reference to global document object is the only option. However, when focus trap is provided, it can be used to infer the document object. Additionally, in some places checks where being made whether given focus trap is the document object itself or not. I wrote a simple utility function for that, which checks the node type and truthiness of the given node instead of comparing it to global document object.

**Checklist**:

- [N/A] Documentation
- [x] Tests
  **Note**: I did not add additional tests for when a global document object is missing, I just checked if all existing tests are passing
- [x] Typings
  **Note**: I didn't do anything manually here, the pre-commit hook did.
- [x] Ready to be merged
  **Note**: there are notes on other checklist items.


src/tab.ts
```diff
@@ -14,14 +15,14 @@ function getNextElement(
   elements: Element[],
   focusTrap?: Document | Element,
```
ph-fritsche, line 16:
```suggestion
  focusTrap: Document | Element,
```

This should never be undefined and if it is we'll have an error in `tab()` anyway.

src/utils/misc/isDocument.ts
```diff
@@ -0,0 +1,3 @@
+export function isDocument(el?: Document | Element): el is Document {
+  return !!el && el.nodeType === el.DOCUMENT_NODE
```
ph-fritsche, line 2:
```suggestion
export function isDocument(el: Document | Element): el is Document {
  return el.nodeType === el.DOCUMENT_NODE
```

ph-fritsche (changes requested):
Thanks a lot for taking the time to contribute :heart: 

Let's remove one more reference to `document` and the redundant `undefined` cases.

While we're at it - could you refactor `getNextElement` and remove the `else` clauses? Just one `if` and the default case without nesting.

ph-fritsche (approved):

loreanvictor:
btw any timeframes for when this will get released?

ph-fritsche:
@loreanvictor Wanted to circle back to this earlier. Sry about the delay.

ph-fritsche:
@all-contributors add @loreanvictor code

Issue #639, opened 2021-04-01
Regression: Click event suddenly stopped working


sbaechler:
- `@testing-library/user-event` version:  13.1.1

- Testing Framework and version: Jest 26.6.3. 
  
- DOM Environment: JSDOM 16.4.0, node 12.19.0
  

Relevant code or config

This code opens a modal that contains some buttons and clicks on them. The modal is provided by Reactstrap (8.9.0).

```javascript
    it('can discard and apply a filter selection', () => {
        render(
            <TestRootContainer initialState={mockStore}>
                <BookingsFilter />
            </TestRootContainer>,
        );

        // open filter popover
        user.click(
            screen.getByRole('button', {
                name: 'cmb-web.filter.filter-by',
            }),
        );

        // expect filter apply button
        screen.getByRole('button', { name: 'cmb-web.filter.apply' });
        // expect no filters set
        screen.getByText('cmb-ui-core.active-filter-items.no-active-filters');
        // expect no reset button
        expect(screen.queryByRole('button', { name: 'cmb-web.filter.reset' })).toBe(null);

        // select a byType filter
        user.click(screen.getByRole('button', { name: 'cmb-ui-core.account-detail-filter.credit' }));
       //^^^^^^^^^^^^^
       //Here the code execution just stops. The onclick handler on the button is never called.

        // The test failed here because the button does not appear.
        user.click(screen.getByRole('button', { name: 'cmb-web.filter.reset' }));


        // Expect popover to have closed
        expect(screen.queryByRole('button', { name: 'cmb-web.filter.apply' })).toBe(null);

```

What you did:
This test failed after an upgrade of user-events from 12.6.3 to 13.1.1. It worked again when I downgraded back to 12.6.3.

What happened:
The test failed at the marked spot. Adding async / await did not help. Also using the bound `getByRole` did not help.

When run with the debugger the context never switched back to the React component after the click on the button. The
flow went straight to the next line in the test.

The test then failed in the next line since the button click handler had not been called the UI never updated.

Reproduction repository:

I'll try to find time to create a minimal repo that can reproduce the issue. But I hope that maybe you have an idea what could be the reason for this regression.

Problem description:
The modal component does register some event listeners on document that close it when the user clicked outside of it.
Could this be the reason?

If it helps, I can go through every release in between and try to see where the regression happened.

bulutfatih:
I have the same. It doesn't work with >=13.1.0
Downgraded to 13.0.16 for now.

[...]

PR #647, opened 2021-04-12
fix(click): throw an error when trying to click with pointer-events set to none

vicrep:
**What**:

As discussed here: https://github.com/testing-library/user-event/issues/639#issuecomment-816924740 , the new behaviour which was recently added to prevent clicks from being emitted when pointer-events are set to `none` had the side-effect of silently no-oping when this was the case, making it:
- non-trivial to figure out why your next assertion failed, given the click 'executed' in the PoV of the test
- hard to work around libraries which set these pointer events while animating

I've added this behaviour to both the click and double-click.

**Why**:

This change follows the paradigm that testing-library provides for selectors, where `getBy` will throw an error if the element doesn't exist. Similarly, trying to click something which is not clickable should also throw an error, as this is an action which cannot be performed by a real user. Some other testing frameworks, such as Cypress, follow similar behaviour of failing when trying to click non-clickable elements.

This also makes dealing with animating elements which might set their pointer events to `none` while animating way more straightforward, using the `waitFor` utility:

```ts
await waitFor(() => userEvent.click(screen.getByText("OK")))
```

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Typings
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ constructor documentation for more options.
 
 Note that `click` will trigger hover events before clicking. To disable this,
 set the `skipHover` option to `true`. Also note that trying to click an element
-with `user-events` being set to `"none"` (i.e. unclickable) will throw an error.
+with `pointer-events` being set to `"none"` (i.e. unclickable) will throw an error.
 
 ### `dblClick(element, eventInit, options)`
 
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ constructor documentation for more options.
 
 Note that `click` will trigger hover events before clicking. To disable this,
 set the `skipHover` option to `true`. Also note that trying to click an element
-with `user-events` being set to `"none"` (i.e. unclickable) will throw an eror.
+with `user-events` being set to `"none"` (i.e. unclickable) will throw an error.
 
 ### `dblClick(element, eventInit, options)`
 
--- a/README.md
+++ b/README.md
@@ -135,7 +135,8 @@ See the
 constructor documentation for more options.
 
 Note that `click` will trigger hover events before clicking. To disable this,
-set the `skipHover` option to `true`.
+set the `skipHover` option to `true`. Also note that trying to click an element
+with `user-events` being set to `"none"` (i.e. unclickable) will throw an eror.
 
 ### `dblClick(element, eventInit, options)`
 
@@ -256,11 +257,8 @@ test('types into the input', () => {
   render(
     <>
       <label for="time">Enter a time</label>
-      <input
-        type="time"
-        id="time"
-      />
-    </>
+      <input type="time" id="time" />
+    </>,
   )
   const input = screen.getByLabelText(/enter a time/i)
   userEvent.type(input, '13:58')
--- a/src/click.ts
+++ b/src/click.ts
@@ -120,7 +120,7 @@ function click(
 ) {
   if (!hasPointerEvents(element)) {
     throw new Error(
-      'unable to click element as it or a parent has pointer-events set to "none".',
+      'unable to click element as it has or inherits pointer-events set to "none".',
     )
   }
   if (!skipHover) hover(element, init)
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -136,7 +136,7 @@ constructor documentation for more options.
 
 Note that `click` will trigger hover events before clicking. To disable this,
 set the `skipHover` option to `true`. Also note that trying to click an element
-with `user-events` being set to `"none"` (i.e. unclickable) will throw an eror.
+with `user-events` being set to `"none"` (i.e. unclickable) will throw an error.
```
ph-fritsche, line 139:
```suggestion
with `pointer-events` being set to `"none"` (i.e. unclickable) will throw an error.
```

Thanks for spotting the typo.
While we're at it, let's fix the property name to.

ph-fritsche (changes requested):

README.md
```diff
@@ -135,7 +135,8 @@ See the
 constructor documentation for more options.
 
 Note that `click` will trigger hover events before clicking. To disable this,
-set the `skipHover` option to `true`.
+set the `skipHover` option to `true`. Also note that trying to click an element
+with `user-events` being set to `"none"` (i.e. unclickable) will throw an error.
```
ph-fritsche, line 139:
```suggestion
with `pointer-events` being set to `"none"` (i.e. unclickable) will throw an error.
```

src/click.ts
```diff
@@ -118,7 +118,11 @@ function click(
   init?: MouseEventInit,
   {skipHover = false, clickCount = 0}: clickOptions = {},
 ) {
-  if (!hasPointerEvents(element)) return
+  if (!hasPointerEvents(element)) {
+    throw new Error(
+      'unable to click element as it or a parent has pointer-events set to "none".',
```
ph-fritsche, line 123:
```suggestion
      'unable to click element as it has or inherits pointer-events set to "none".',
```

src/click.ts
```diff
@@ -143,7 +147,11 @@ function fireClick(element: Element, mouseEventOptions: MouseEventInit) {
 }
 
 function dblClick(element: Element, init?: MouseEventInit) {
-  if (!hasPointerEvents(element)) return
+  if (!hasPointerEvents(element)) {
+    throw new Error(
+      'unable to double-click element as it or a parent has pointer-events set to "none".',
```
ph-fritsche, line 152:
```suggestion
      'unable to double-click element as it has or inherits pointer-events set to "none".',
```

ph-fritsche (changes requested):
Thanks for your contribution. :heart:

ph-fritsche:
We probably should throw the same error on `userEvent.hover` and `userEvent.unhover`

vicrep:
@ph-fritsche I actually considered doing it for hover too, but wasn't sure if out of scope for this PR. Given that hovers usually provide side-effects and aren't usually explicit actions, I'm not sure if we should error when hovering over something that's disabled. 

FWIW, I find that Cypress has very sensible assertions they make on their commands to simulate user behaviour (such as, [failing when trying to click a non-clickable element](https://docs.cypress.io/api/commands/click#Assertions)) and this is the inspiration I used for this PR. Interestingly, `hover` is [basically the only user 'event' they don't provide a command for](https://docs.cypress.io/api/commands/hover), due to how non-deterministic its behaviour can be depending on implementation. This further added to my uncertainty about adding this to the PR.

What do you think? I'd be happy to add it to this PR, or to propose it as a future (or separate) improvement.

README.md
vicrep: Good catch 🙈

ph-fritsche:
@vicrep I think users calling `userEvent.hover` expect a `mouseover` to be dispatched. So the same logic as for the `click` applies. We don't have layout in jsdom so we could not determine the elements the pointer events would fall through to.

vicrep:
@ph-fritsche makes sense, I've added the same behaviour to hover/un-hover 👍 

Please let me know if there's anything else you need from me to merge this :)

ph-fritsche (approved):
Looks good to me :)

ph-fritsche:
@all-contributors add @vicrep code

PR #654, opened 2021-04-16
Fix to some tests with [Space], Adds checkbox,radio to clickable inputs

### LINKED ISSUE #653: How should I test a Radio with onChange is called when Space is pressed?
By default, an onChange event will fire when a radio button is focused and the user presses [Enter] or [Space]. Example: https://codepen.io/bozdoz/pen/zYNLGWe?editors=1001

Here's what I expect should work with the new keyboard method:

```javascript
it('activates radio with spacebar', () => {
    const onChange = jest.fn();

    render(<input type="radio" onChange={onChange} />);

    userEvent.tab();

    userEvent.keyboard('[Space]');

    expect(onChange).toHaveBeenCalled();
});
```

I expect that test should pass, but it is not passing (because onChange has not been called).  Any idea where I'm going wrong here, or suggestions?

bozdoz:
**What**:

Fixes issues with pressing Space on a focused checkbox/radio input.  Fixes tests for it.

**Checklist**:

- [x] Tests

Closes #653

### DIFF
--- a/src/__tests__/keyboard/plugin/functional.ts
+++ b/src/__tests__/keyboard/plugin/functional.ts
@@ -138,13 +138,32 @@ test('trigger click event on [Space] keyup on HTMLButtonElement', () => {
 })
 
 test('trigger click event on [Space] keyup on HTMLInputElement type=button', () => {
-  const {element, getEventSnapshot, getEvents} = setup(`<button/>`)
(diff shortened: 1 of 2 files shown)


ph-fritsche:
@all-contributors add @bozdoz bug, test

PR #663, opened 2021-04-23
chore: continue testing on Node 15

nickmccurdy:
It's still supported until June 1

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -17,7 +17,7 @@ jobs:
   main:
     strategy:
       matrix:
-        node: [10, 12, 14, 16]
+        node: [10, 12, 14, 15, 16]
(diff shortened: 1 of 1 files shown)


nickmccurdy:
I would merge, but I'm not sure how we'd want to update the branch protection rules

ph-fritsche:
I think running it without requiring it is fine. But we could as well drop it or add the check requirement.
It's just about a month and this repo doesn't do anything that can be expected to work in one node version but not the other.
It's more or less just a smoke test for our config and dependencies and I don't expect any of them to support node16 and node14 but break in node15.

nickmccurdy:
Personally I'd still rather have that smoke test even though it's unlikely to fail, because Node 15 is still expected to be supported for now.

PR #675, opened 2021-05-05
Chore/rename to main branch

itsik-avidan:
**What**:
Changed the refs from the 'master' branch into 'main' 
resolves https://github.com/testing-library/user-event/issues/674

**Why**:
To be more compliment with updated web standards 

**How**:
updated mainly the contribution guide and the validation file, kept 'master' branch available to be backward compatible as an intermediate phase 

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -5,6 +5,7 @@ on:
       [
         '+([0-9])?(.{+([0-9]),x}).x',
         'main',
+        'master',
         'next',
         'next-major',
         'beta',
@@ -46,7 +47,7 @@ jobs:
     runs-on: ubuntu-latest
     if:
       ${{ github.repository == 'testing-library/user-event' &&
-      contains('refs/heads/main,refs/heads/beta,refs/heads/next,refs/heads/alpha',
+      contains('refs/heads/main,refs/heads/master,refs/heads/beta,refs/heads/next,refs/heads/alpha',
       github.ref) && github.event_name == 'push' }}
     steps:
       - name: ⬇️ Checkout repo
@@ -75,6 +76,7 @@ jobs:
             [
               '+([0-9])?(.{+([0-9]),x}).x',
               'main',
+              'master',
               'next',
               'next-major',
               {name: 'beta', prerelease: true},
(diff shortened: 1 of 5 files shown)


ph-fritsche:
I'm not sure about this. Basically the `default` branch can be named anything. So changing it doesn't really hurt.
On the other hand: Why change it at all as it is arbitrary and meaningless?
If you want to attribute any meaning, imho it should be renamed `latest` as it is our release branch to the `latest` channel.

nickmccurdy:
> So changing it doesn't really hurt.

We're already using `main` in most of our projects, so not renaming user-event's branch could cause confusion.

> On the other hand: Why change it at all as it is arbitrary and meaningless?

The term `master` has potentially racist and offensive connotations.

> If you want to attribute any meaning, imho it should be renamed `latest` as it is our release branch to the `latest` channel.

While I personally like keeping default branches stable, commits in default branches are not necessarily published to npm (IIRC our chore commits aren't actually published until a commit after them needs to be published). Additionally, the git community (including GitHub's new repositories and Testing Library as a whole) is already moving to `main`, so I think it's best to be consistent. We'll keep the `latest` tag on npm because npm is not a version control system, so that's not directly related.

nickmccurdy (approved):

ph-fritsche:
> The term `master` has a history in racism and is potentially offensive.

I'm aware of this. I just don't agree.
Everything might be connoted and be offensive in a different context. But it's a different context.

> are not necessarily published to npm (IIRC our chore commits aren't actually published until a commit after them needs to be published).

You're absolutely correct. But different areas on the same channel aren't something unheard of.
If you have a latest-branch, a latest-testing and a latest-build that doesn't necessarily mean your latest-package will at all time refer to the same commit. One could fail or in the case of `chore` commits decide that it is not necessary to relay it further.

> We're already using `main` in most of our projects, so not renaming user-event's branch could cause confusion. [...]
> Additionally, [...] Testing Library as a whole is already moving to `main`, so I think it's best to be consistent.

:+1: Good enough for me if you guys feel strongly about this.

PR #679, opened 2021-05-11
Fix unexpected behavior with setSelectionRange on inputs in the browser

### LINKED ISSUE #677: `userEvent.type` with selection range behaves unexpectedly when run in a browser
Hello again :wave:   I’m seeing inconsistent behavior with `userEvent.type` being run in the test vs. in the browser outside of the test. For some reason, the typing with selection range example from the docs works in the JS test runner, but doesn't work in the actual browser.

- `@testing-library/user-event` version: 13.1.8
- `@testing-library/react` version: 11.2.5

Relevant code or config
```javascript
import React from “react”;
import { render, screen } from “@testing-library/react”;
import userEvent from “@testing-library/user-event”;

render(
  <div>
    <label htmlFor=“my-input”>Example:</label>
    <input id=“my-input” type=“text” defaultValue=“This is a bad example” />
  </div>
);
const input = screen.getByLabelText(/example/i);
input.setSelectionRange(10, 13);
userEvent.type(input, “good”);
```
What I did:
I took the userEvent.type with selection range example from the docs and ran it in a js file outside a test.

What happened:
The resulting input value was supposed to be `"this is a good example"`, but instead it was `"doothis is a g example"`:

Screen Shot 2021-05-06 at 12 03 53 PM

Reproduction repository:

[Sandbox link](https://codesandbox.io/s/c17hn?file=/src/index.js)

Suggested solution:
Based on stepping through the debugger, I think the issue is related to [these lines](https://github.com/testing-library/user-event/blob/master/src/utils/edit/selectionRange.ts#L108-L113) in `setSelectionRange`. It seems like `selection.addRange(range)` might not be working as expected so then selection range is cleared to zero after `selection.removeAllRanges()`. I'm not sure how to solve this though.

sydneyjodon-wk:
**What**:

`userEvent.type` with selection range does not work as expected in the browser. Closes #677 

**Why**:

When an input or textarea is clicked, the parent div is selected and the `Selection` range should not be updated at the end of `setSelectionRange`.

**How**:

Skip setting `Selection` range for `<input>` and `<textarea>` in `setSelectionRange` utility. (see [comment](https://github.com/testing-library/user-event/issues/677#issuecomment-836280924) on issue)

**Checklist**:

- [x] Documentation
- [x] Tests
  - [x] Tests pass locally
  - [x] Updated [sandbox](https://codesandbox.io/s/rtl-selection-range-example-u63zc?file=/src/index.js) works as expected
- [x] Ready to be merged

### DIFF
--- a/src/utils/edit/selectionRange.ts
+++ b/src/utils/edit/selectionRange.ts
@@ -107,7 +107,13 @@ export function setSelectionRange(
 
   const selection = element.ownerDocument.getSelection()
   // istanbul ignore else
-  if (selection) {
+  if (
+    selection &&
+    // Skip setting the range for <input> and <textarea> because
+    // the parent <div> will be selected.
+    !isElementType(element, 'input') &&
+    !isElementType(element, 'textarea')
+  ) {
     selection.removeAllRanges()
     selection.addRange(range)
   }
--- a/src/utils/edit/selectionRange.ts
+++ b/src/utils/edit/selectionRange.ts
@@ -96,6 +96,11 @@ export function setSelectionRange(
     }
   }
 
+  // Moving the selection inside <input> or <textarea> does not alter the document Selection.
+  if (isElementType(element, 'input') || isElementType(element, 'textarea')) {
+    return
+  }
+
   const range = element.ownerDocument.createRange()
   range.selectNodeContents(element)
 
@@ -107,13 +112,7 @@ export function setSelectionRange(
 
   const selection = element.ownerDocument.getSelection()
   // istanbul ignore else
-  if (
-    selection &&
-    // Skip setting the range for <input> and <textarea> because
-    // the parent <div> will be selected.
-    !isElementType(element, 'input') &&
-    !isElementType(element, 'textarea')
-  ) {
+  if (selection) {
     selection.removeAllRanges()
     selection.addRange(range)
   }


src/utils/edit/selectionRange.ts
```diff
@@ -107,7 +107,13 @@ export function setSelectionRange(
 
   const selection = element.ownerDocument.getSelection()
   // istanbul ignore else
-  if (selection) {
+  if (
+    selection &&
+    // Skip setting the range for <input> and <textarea> because
+    // the parent <div> will be selected.
+    !isElementType(element, 'input') &&
+    !isElementType(element, 'textarea')
```
ph-fritsche, line 115: Could you move the check to above line 99 and just `return` there.
I'd prefer to have the the check and comment stick out.

> the parent `<div>` will be selected.

The selecting of the parent I mentioned on the issue happened on the `click`.
I think the selection is not changed at all. Might be that the selection can be set differently.

Unless you discovered that the typing does move the Range, let's not confuse future devs.
Something like:
> Moving the selection inside `<input>` or `<textarea>` does not alter the document Selection.

ph-fritsche (changes requested):

src/utils/edit/selectionRange.ts
sydneyjodon-wk: Updated in 9efb5b1

ph-fritsche (approved):

ph-fritsche:
@all-contributor add @sydneyjodon-wk code

ph-fritsche:
@all-contributors add @sydneyjodon-wk code

PR #693, opened 2021-07-02
chore: upgrade to Jest 27

nickmccurdy:
**What**: Upgrade to Jest 27

**Why**: Ensuring compatibility with breaking changes in Jest 27

**How**: Upgrading packages

**Checklist**:

- [ ] Documentation (N/A)
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -232,6 +232,7 @@ test('should delay the typing when opts.delay is not 0', async () => {
 
   const text = 'Hello, world!'
   const delay = 10
+  // eslint-disable-next-line testing-library/no-await-sync-events
   await userEvent.type(element, text, {delay})
 
   expect(onInput).toHaveBeenCalledTimes(text.length)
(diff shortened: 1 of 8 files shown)


src/__tests__/type.js
```diff
@@ -232,6 +232,7 @@ test('should delay the typing when opts.delay is not 0', async () => {
 
   const text = 'Hello, world!'
   const delay = 10
+  // eslint-disable-next-line testing-library/no-await-sync-events
   await userEvent.type(element, text, {delay})
```
ph-fritsche, line 236: Is the plugin not picking up the overload anymore ? :disappointed: 

This might be just another reason for us to always return a Promise on `userEvent.type` in the next major.

ph-fritsche (approved):

src/__tests__/type.js
nickmccurdy: It seems like it. I've opened https://github.com/testing-library/eslint-plugin-testing-library/issues/403 so we can discuss it there.

Would you still want to merge this in the meantime or would you rather wait for that to be fixed?

PR #698, opened 2021-07-11
feat: Drop support for node 10

### LINKED ISSUE #681: Drop Node 10 support
Because it's [EOL](https://nodejs.org/en/about/releases).

nickmccurdy:
BREAKING CHANGE: node 10 is no longer supported. It reached its end-of-life on 30.04.2021.

**What**:

Closes #681

**Why**:

Reduces maintenance burden

**How**:

Change `engines` field to require `node>=12`

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,3 +1,4 @@
 {
-  "sandboxes": ["vbcvs"]
+  "sandboxes": ["vbcvs"],
+  "node": "12"
 }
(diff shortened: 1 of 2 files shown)


ph-fritsche:
Would it be reasonable to remove the `engines` restriction and make this a `fix` instead of `major` ?
We don't rely on any of the Node APIs in this package. We even support test environments in the browser.

Issue #638, opened 2021-03-31
add pointerDown, pointerMove and pointerUp events for API


lifeiscontent:
I think it would be beneficial to have pointerDown, pointerMove, and pointerUp user events so you can encapsulate touch events, mouse events, pointer events all under one API.

The reason why I think something like this should exist is it would allow for testing complex UI components like a slider input.

the basic implementation I think could work from a high level is the following:

```js
const input = document.getElementById('input');
// you could use a percent to represent the bounds of the element
// or use a specific number to represent a pixel value
userEvent.pointerDown(input, {x: '50%', y: '50%'}); // starting point of where you "clicked"
userEvent.pointerMove(input, {x: '75%', y: '50%', duration: '1s'}); // where you want to move to over a period of time
userEvent.pointerUp(input); // remove your finger from the screen/mouse/etc
```

@kentcdodds what do you think?

kentcdodds:
What difference would this be from just using fireEvent directly?

lifeiscontent:
It would provide a more ergonomic API for emulating down/move/up events just like all the methods in this API provide.

ph-fritsche:
@lifeiscontent Could you describe the differences between your idea and the behavior of `userEvent.hover()` and `userEvent.click()`?

I don't think that adding more APIs with separate implementations is a good idea.
But the pointer related APIs might benefit from an abstract underlying implementation.
A `pointer` API could provide a solid stateful handling of pointer events that is then used by the convenience APIs like `click`, `dblClick` etc.
Like `keyboard` provides the implementation that is used for `type`.

kentcdodds:
Keep in mind that `user-event` is intended to simulate common things users do. Based on your example, it looks like you're trying to simulate a user pressing down on an element, moving their finger somewhere on that element and then pulling their finger up. I can see how this might be useful in some very niche situations, but most people wouldn't have a use for this. Everything we add to `user-event` adds complexity and therefore every user of `user-event` pays the cost for what's added. We want to make sure the cost everyone pays is worth the benefit to most people. This is just good OSS stewardship.

I think what's being suggested here is outside the scope of `user-event`. You could absolutely build these utilities yourself and even publish them as a package on npm. But I don't think we should add this to `user-event`.

lifeiscontent:
@ph-fritsche Yep, totally.

So if you're testing a range input.

![image](https://user-images.githubusercontent.com/180963/113440928-35b56f00-93a2-11eb-8276-309423214332.png)

[...]

PR #703, opened 2021-07-20
fix(keyboard): dispatch change event on blur

### LINKED ISSUE #682: userEvent.type does not dispatch change events on input with @testing-library/dom
- `@testing-library/user-event` version: 13.1.9
- Testing Framework and version:  `jest` 26.6.3, `@testing-library/dom` 7.31.0
- DOM Environment: `jsdom` 16.5.3

I have some code that listens for `change` events on an `input`. How do I make sure the event is triggered when using `userEvent.type`? I understand that typing alone does not trigger the event - only blurring the field. But in the browser, clicking another element or pressing tab causes this to happen. I am unable to reproduce this interaction with `userEvents`.

```javascript
test("fires change event after typing", () => {
  const listener = jest.fn();

  document.body.innerHTML = `
    <label for="name">Name</label>
    <input type="text" id="name" />
    <button role="button">Save</button>
  `;
  // This would normally live somewhere in application code
  document.querySelector("input").addEventListener("change", listener);

  userEvent.type(screen.getByLabelText("Name"), "Max");
  userEvent.tab();
  userEvent.click(screen.getByRole("button", { name: "Save" }));

  expect(screen.getByLabelText("Name").value).toEqual("Max");
  expect(listener).toHaveBeenCalled(); // fails
});
```

Reproduction sanbox: https://codesandbox.io/s/cocky-swartz-5nnn6?file=/src/type.test.js

ph-fritsche:
**What**:
Closes #682 

**Why**:
The browser dispatches a `change` event when an element with changed value loses focus.
When we alter the element `value`/`textContent`, we dispatch an `input` event but no native `change` event.

**How**:
When setting the `value`/`textContent` an event handler is added to the capture phase of the `blur` event on the `window`.
If the `value`/`textContent` on our element is different than the initial one, we dispatch a `change` event.

**Checklist**:
- [N/A] Documentation
- [x] Tests
- [x] Ready to be merged

**Constraints**:
* We can not determine if an element was changed between gaining focus and our first edit.
  We could (also) set the initial value on the `focus` event, but for this we would either:
  1) add it on the document as side-effect of the `import` (only works if the `document` is global)
  2) add it per additional `setup` API (and/or implicitly when calling one of our other APIs)
* There is no cross-platform solution for adding an event handler to the bottom of the stack.
  Therefore other event handlers on `blur` (that are added on the capture phase on `window`) might be executed before we dispatch the change event.

### DIFF
--- a/src/__tests__/keyboard/shared/fireInputEvent.ts
+++ b/src/__tests__/keyboard/shared/fireInputEvent.ts
@@ -1,0 +1,23 @@
+import {setup} from '__tests__/helpers/utils'
+import userEvent from '../../../'
+
+it('dispatch change event on blur', () => {
+  const {element, getEvents} = setup('<input/>')
+
+  ;(element as HTMLInputElement).focus()
+  userEvent.keyboard('foo')
+  ;(element as HTMLInputElement).blur()
+
+  expect(getEvents('change')).toHaveLength(1)
+})
+
+it('do not dispatch change event if value did not change', () => {
+  const {element, getEvents} = setup('<input/>')
+
+  ;(element as HTMLInputElement).focus()
+  userEvent.keyboard('foo')
+  userEvent.keyboard('{backspace}{backspace}{backspace}')
+  ;(element as HTMLInputElement).blur()
+
+  expect(getEvents('change')).toHaveLength(0)
+})
--- a/src/keyboard/shared/fireInputEvent.ts
+++ b/src/keyboard/shared/fireInputEvent.ts
@@ -70,6 +70,15 @@ function setSelectionRangeAfterInputHandler(
   }
 }
 
+const initial = Symbol('initial input value/textContent')
+const onBlur = Symbol('onBlur')
+declare global {
+  interface Element {
+    [initial]?: string
+    [onBlur]?: EventListener
+  }
+}
+
 /**
  * React tracks the changes on element properties.
  * This workaround tries to alter the DOM element without React noticing,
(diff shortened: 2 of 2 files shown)


ph-fritsche:
@nickmccurdy What do you think about adding a `setup` API?
If we extend the `keyboard` approach of an internal state (`options.keyboardState`) to potential new APIs (pointer https://github.com/testing-library/user-event/issues/638#issuecomment-811925877), we could allow the user to start his test at any state of the input devices (e.g. element is added while `ControlLeft` and left mouse button are pressed).

nickmccurdy:
What would usage look like? And this is unrelated from our internal `setup` function for rendering HTML right?

ph-fritsche:
For the problem at hand - i.e. keeping track of non-userevent changes:
```js
it('test something', () => {
  // ... render empty text field ...
  element.focus()
  element.value = 'foo'
  userEvent.keyboard('bar')
  userEvent.keyboard('{Backspace}{Backspace}{Backspace}')
  element.blur() // this triggers no change event but it should
})
it('test again with setup', () => {
  // ... render empty text field ...
  userEvent.setup({document: element.ownerDocument})
  element.focus()
  element.value = 'foo'
  userEvent.keyboard('bar')
  userEvent.keyboard('{Backspace}{Backspace}{Backspace}')
  element.blur() // does trigger a change event because we tracked all changes since `focus()`
})
```

For `keyboardState` and proposed `pointerState`:
```js
it('act on some dialog that opened on [ctrl]+[mouseLeft]', () => {
  // ... render ...
  userEvent.setup({
    keyboardState: userEvent.createKeyboardState('[ControlLeft>]'),
    pointerState: userEvent.createPointerState('[MouseLeft>]'), // tbd how this API will exactly look like
  })

  // possibly some rudimentary pointer movement
  // will not be accurate in jsdom but could be good enough for many cases
  // e.g. testing that drag&drop handlers work in principle
  userEvent.pointerMove(element, {x: 15, y: 20})

  userEvent.pointer(element, '[/MouseLeft]') // trigger pointerUp, mouseUp, click with ctrl=true
})
```

nickmccurdy:
I may be missing something, but I don't see an advantage to using `setup` over calling `render` with a custom element and some custom function calls after. A more specific pointer API could help for testing edge cases, though I'm not sure how commonly that would be needed in JSDOM.

ph-fritsche:
For our blind spot regarding changes that happen before our API call the users would have to reimplement our workarounds for the change event in their render function.
And they would need to be aware of internals of DOM and possibly some framework (React).
I think the problem and the solution are non-obvious.
Pointing the users towards an easy solution like "if you programmatically alter the value before changing it with userEvent (e.g. inside `onFocus`) please call `userEvent.setup` after creating the document" sounds better to me.

---

For the `keyboardState`: Currently the state ist considered internal and returned from `keyboard` as a black box.
```js
let keyboardState = userEvent.keyboard('[ControlLeft>]')
// ... do something ...
keyboardState = userEvent.keyboard('abc', {keyboardState}) // continue key events with ctrl=true
// ... do something else ...
userEvent.keyboard('[/ControlLeft]', {keyboardState}) // release control key
```
But currently it is not (officially) possible to start off with some state without triggering the extra events that come with it. If that should happen in a setup function or right on the keyboard call or both would be to be determined. Just adding the note that it could be used for this.

PR #706, opened 2021-07-20
test: fix istanbul ignore

ph-fritsche:
**What**:
Fix istanbul ignore.

**Why**:
Previously the `ignore else` had to be writting before `else if` to ignore the consecutive `else`.
Istanbul now seems to expect it right before the `if`.

**How**:
Committed with `--no-verify` as the pre-commit hook lint-fixed the change :shrug: 

**Checklist**:
- [N/A] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/keyboard/shared/fireInputEvent.ts
+++ b/src/keyboard/shared/fireInputEvent.ts
@@ -25,7 +25,7 @@ export function fireInputEvent(
   // apply the changes before firing the input event, so that input handlers can access the altered dom and selection
   if (isContentEditable(element)) {
     applyNative(element, 'textContent', newValue)
-  } /* istanbul ignore else */ else if (
+  } else /* istanbul ignore else */ if (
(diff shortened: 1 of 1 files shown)


nickmccurdy:
Thanks for figuring this out. I guess this was a bug or breaking change in an Istanbul update, do you think it's worth reporting?

ph-fritsche:
For istanbul, I don't think so. There has been no activity on [istanbuljs/istanbuljs](https://github.com/istanbuljs/istanbuljs/pulls?q=is%3Apr+is%3Aclosed) recently. We've bumped `kcd-scripts` from `v7` to `v11` and don't have a lock file that could help pin down what changed. It was most likely a fix a long time ago and we just didn't pick up on it because some dependency was pinned.

But I'll report on `kcd-script` that the lintfix on pre-commit hook changes this to the wrong position.

Issue #479, opened 2020-11-02
userEvent.type not entering a space into inputs with type email


wolfe111:
- `@testing-library/user-event` version: 12.1.8

- Testing Framework and version: Jest version 26.6.0
  
- DOM Environment: jsdom 16.4.0 through jest
  

Relevant code or config

```javascript
      <input
        id='emailId'
        type='email'
        onChange={(e: React.ChangeEvent<HTMLInputElement>): void => {
          console.log(e.target.value)
          setEmailAddress(e.target.value);
          setError(null);
        }}
      />

it('should show an error when an email is typed with a space and submitted', async () => {
  const emailAddress = 'some{space}email@example.com'

  render(<App/>);

  const emailField = screen.getByLabelText('Email');
  await userEvent.type(emailField, emailAddress);
  userEvent.click(screen.getByText('Enter email'));

  await waitFor(() => {
    expect(screen.queryByText('No email errors')).not.toBeInTheDocument()
    expect(screen.queryByText('Email has white space')).toBeInTheDocument()
  })
});
```

What you did:
I wanted to test a email validation message I had in on one of my forms and had an issue inputting a space into an input box with type=email. When running it in the browser, I'm freely allowed to enter a space.

What happened:
userEvent.type would ignore the space being typed in. See console logs when running in repo below. 

Reproduction repository:
https://github.com/wolfe111/user-event-space
(created with create-react-app and updated @testing-library/jest-dom, @testing-library/react and @testing-library/user-event to most recent version)

marcosvega91:
Hi @wolfe111 thanks for reaching us :).

I think that this is a bug. As for the [spec](https://html.spec.whatwg.org/multipage/input.html#e-mail-state-(type=email)) some inputs have `sanitization algorithm`. The sanitization algorithm for the `value` of type `email` will  

- remove new lines from the value 
- remove trailing and leading spaces

`jsdom` is following this algorithm 
https://github.com/jsdom/jsdom/blob/5279cfda5fe4d52f04b2eb6a801c98d81f9b55da/lib/jsdom/living/helpers/form-controls.js#L214-L223

On our side I think that we should keep the value with the space and use on the next iteration of the type function instead of using the value of the input.

BentMylund:
[...]

PR #712, opened 2021-08-06
fix(type): type a space into email and url input (#479)

wolfe111:
**What**:
Closes [#479](https://github.com/testing-library/user-event/issues/479) 

**Why**:
It makes type behave more like it would in the browser for email and url inputs. Currently both input types ignore spaces

**How**:
Added new handlers in character.js for email inputs and url inputs, which keep the value of the space and use on the next iteration.

**Checklist**:

- [N/A] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/keyboard/plugins/character.ts
+++ b/src/keyboard/plugins/character.ts
@@ -4,7 +4,7 @@
 
 import {fireEvent} from '@testing-library/dom'
 import {fireChangeForInputTimeIfValid, fireInputEvent} from '../shared'
-import {behaviorPlugin} from '../types'
+import {behaviorPlugin, keyboardState} from '../types'
 import {
   buildTimeValue,
   calculateNewValue,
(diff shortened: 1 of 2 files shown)


src/keyboard/plugins/character.ts
```diff
@@ -147,12 +147,65 @@ export const keypressBehavior: behaviorPlugin[] = [
...
+  },
+  {
+    matches: (keyDef, element) =>
+      keyDef.key?.length === 1 && isElementType(element, 'input', {type: 'url', readOnly: false}
+      ),
+    handle: (keyDef, element, options, state) => {
+      const oldValue =
+        state.carryValue ?? getValue(element) ?? /* istanbul ignore next */ ''
+
+      const selectionRange = oldValue.endsWith(" ") ? {
```
ph-fritsche, line 187: I think this should check for `state.carryValue` being `undefined`.
Otherwise this adds the implicit and wrong assumption that `Space` at the end of the value is the only possible case for `carryValue`.

src/keyboard/plugins/character.ts
```diff
@@ -147,12 +147,65 @@ export const keypressBehavior: behaviorPlugin[] = [
...
+        selectionStart: null,
+        selectionEnd: null
+      } : undefined
+
+      const {newValue, newSelectionStart} = calculateNewValue(
+        keyDef.key as string,
+        element as HTMLElement,
+        oldValue,
+        selectionRange
```
ph-fritsche, line 196: Am I interpreting this correctly that you set `selectionRange` to force `calculateNewValue` to not use `input.selectionStart` although it exists on this type of element?

Doesn't forcing the `selectionRange` here lead to diverging from other behavior implementations like `Backspace`?
`userEvent.type(input, 'foo {Backspace}bar')` => `fo bar` ?

ph-fritsche (changes requested):
Thanks a lot for taking the time to solve this. :heart: 

Are all possible calculated new values valid as display values? Or does the browser accept some input like `Space` but reject others?

src/keyboard/plugins/character.ts
wolfe111: That's a very good point, I'll add a undefined check in

src/keyboard/plugins/character.ts
wolfe111: Ahhh yeah you are interpreting that correctly, that's my bad!!

I added that to handle the case for the element type being ur. This is due to getSelectionRange using the HTML element to determine the length and the input url element ignores the space.

Open to any suggestions on how you would handle this?

I thought of changing the beginning of getSelectionRange to the below. But I'm not convinced I like it 😂 

```
export function getSelectionRange(
  element: Element,
  value?: string
): {
  selectionStart: number | null
  selectionEnd: number | null
} {
  if (hasSelectionSupport(element)) {
    if (value && value.endsWith(" ") && element.type === 'url') {
      return {
        selectionStart: value.length,
        selectionEnd: value.length,
      }
    }

    return {
      selectionStart: element.selectionStart,
      selectionEnd: element.selectionEnd,
    }
  }
```

wolfe111:
> Thanks a lot for taking the time to solve this. ❤️
> 
> Are all possible calculated new values valid as display values? Or does the browser accept some input like `Space` but reject others?

Both url and email inputs accept any DOMString as their value and then run their own validation after. So I believe it shouldn't reject others. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email#value
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url#value

src/keyboard/plugins/character.ts
ph-fritsche: Remember space at the end is not the only possible display value diverging from value property.
Also I don't like specific workarounds being scattered across codebase.
Maybe this should wait for #714

wolfe111: Yeah, that's a good point. I'll close this pull request and have a look at a later point once #714 is resolved

Issue #618, opened 2021-03-24
Multiple `keydown` events in a single key press


axelboc:
Just wondering if it's possible with the new `userEvent.keyboard()` to trigger multiple consecutive `keydown` events in a single key press. I'm trying to simulate an interaction with a slider where the user keeps pressing an arrow key to move further than they would in a single down/up key press.

I've tried the following, which obviously doesn't work (I only get one `keydown` event) 😄:

```js
userEvent.keyboard('{ArrowLeft>}{ArrowLeft>}{ArrowLeft>}{ArrowLeft>}{ArrowLeft>}{/ArrowLeft}');
```

If that's not currently supported, may I suggest the following syntax?

```js
userEvent.keyboard('{ArrowLeft>5}{/ArrowLeft}');
```

- `@testing-library/user-event` version: 13.0.10
- Testing Framework and version: Jest  via latest CRA (4.0.3)
- DOM Environment: JSDOM 16

ph-fritsche:
Thanks for opening this issue.
Sadly this is not possible yet, but adding syntax for keeping keys pressed and repeating the events is definitely on the roadmap.

The syntax is fine.
If you want to pick up the work on this, a PR is very much welcomed. :smiley: 

The changes should take place in [getNextKeyDef](https://github.com/testing-library/user-event/blob/master/src/keyboard/getNextKeyDef.ts) for analyzing the directive and [keyboardImplementation](https://github.com/testing-library/user-event/blob/master/src/keyboard/keyboardImplementation.ts).

`keyboardState` can be used to carry the repetition.
`{ArrowLeft>}{ArrowLeft}` should still release `ArrowLeft` before it is pressed again.

PR #728, opened 2021-09-22
feat: #618 Trigger multiple keydown events in a single key press

patricklizon:
**What**:

Trigger multiple keydown events in a single key press

**Why**:

As requested in https://github.com/testing-library/user-event/issues/618

**How**:

I followed @ph-fritsche suggestions https://github.com/testing-library/user-event/issues/618#issuecomment-805753235
- added `repeatModifier` detection at `getNextKeyDef`
- added `fireKeyDownTimes` to `keyboardState`
- added `fireOnceOrMore` at `keyboardImplementation`

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/keyboard/keyboardImplementation.ts
+++ b/src/__tests__/keyboard/keyboardImplementation.ts
@@ -16,20 +16,16 @@ test('no character input if `altKey` or `ctrlKey` is pressed', () => {
   expect(eventWasFired('input')).toBe(false)
 })
 
-describe('fire multiple events for', () => {
-  describe('keydown', () => {
-    it('_a', () => {
-      const {element, getEventSnapshot, clearEventCalls} = setup(`<input/>`)
-      const spy = jest.fn()
+describe('fire events', () => {
+  it('_a', () => {
+    const {element, getEventSnapshot, clearEventCalls} = setup(`<input/>`)
 
-      ;(element as HTMLInputElement).focus()
-      element?.addEventListener('keydown', spy)
-      clearEventCalls()
+    ;(element as HTMLInputElement).focus()
+    clearEventCalls()
 
-      userEvent.keyboard('{ArrowLeft>}{ArrowLeft}')
+    userEvent.keyboard('{ArrowLeft>}{ArrowLeft}')
 
-      expect(spy).toBeCalledTimes(2)
-      expect(getEventSnapshot()).toMatchInlineSnapshot(`
+    expect(getEventSnapshot()).toMatchInlineSnapshot(`
 Events fired on: input[value=""]
 
 input[value=""] - keydown: ArrowLeft (37)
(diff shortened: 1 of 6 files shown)


src/keyboard/keyboardImplementation.ts
```diff
@@ -65,6 +76,7 @@ export async function keyboardImplementation(
     if (options.delay > 0) {
       await wait(options.delay)
     }
+
     return keyboardImplementation(text.slice(consumedLength), options, state)
```
ph-fritsche, line 80: I think the implementation should take place here.
Pushing the `keyDef` that is to be repeated on the state. Something like
```js
state.repeatKey = { times: fireKeydownTimes, consumedLength, keyDef }
```
Then these should be used on the next iteration instead of `getNextKeyDef()`.
And when times is counted down to `0` it should be removed and `consumedLength` should be applied.

src/keyboard/keyboardImplementation.ts
```diff
@@ -48,11 +50,20 @@ export async function keyboardImplementation(
         keyDef,
         getCurrentElement,
         options,
-        state,
+        fireKeyDownTimes > (state.fireKeyDownTimes || 1)
+          ? {...state, fireKeyDownTimes}
```
ph-fritsche, line 54: Please don't create a derived `state`.
The state should be a shared truth about the state of the input device(s) across all our events.

ph-fritsche (changes requested):
First of all thanks a lot for taking the time to contribute. :heart: 

The implementation does not properly solve the issue yet.

When a key is pressed in the browser, the browser dispatches a `keydown` event every `n` ms.
(Usually there is a primary delay before the key is considered "pressed" and then there is a secondary delay between each of the `keydown` events until the key is released or another one is either pressed or released - both usually according to some system setting.)

For each of the `keydown` events there might be the default behavior of a `keypress` and depending on the active element an `input` event.

Instead of dispatching each event multiple times the whole implementation needs to loop for `fireKeyDownTimes`.

src/keyboard/keyboardImplementation.ts
patricklizon: Reverted 👍🏻

src/keyboard/keyboardImplementation.ts
patricklizon: Changed 👍🏻

patricklizon:
@ph-fritsche Thank you for explaining how triggering keyboard events work in browser. That's great something I was not aware of. It's great learning opportunity for me, I'll definitely dig deeper into the topic and the source code of library 🙂

src/__tests__/keyboard/keyboardImplementation.ts
```diff
@@ -15,3 +15,197 @@ test('no character input if `altKey` or `ctrlKey` is pressed', () => {
...
+Events fired on: input[value="aa"]
+
+input[value=""] - keydown: a (97)
+input[value=""] - keypress: a (97)
+input[value="a"] - input
+input[value="a"] - keyup: a (97)
+input[value="a"] - keydown: a (97)
+input[value="a"] - keypress: a (97)
+input[value="aa"] - input
+input[value="aa"] - keyup: a (97)
```
ph-fritsche, line 90: See this log produced in the browser:
```
{"type":"keydown","key":"a","code":"KeyA"}
{"type":"keypress","key":"a","code":"KeyA"}
{"type":"input","data":"a","inputType":"insertText","value":"a"}
{"type":"keydown","key":"a","code":"KeyA"}
{"type":"keypress","key":"a","code":"KeyA"}
{"type":"input","data":"a","inputType":"insertText","value":"aa"}
{"type":"keydown","key":"a","code":"KeyA"}
{"type":"keypress","key":"a","code":"KeyA"}
{"type":"input","data":"a","inputType":"insertText","value":"aaa"}
{"type":"keyup","key":"a","code":"KeyA"}
```

You can use [this codesandbox](https://codesandbox.io/s/keyevent-vtdcc?file=/src/App.js) to test out events.

ph-fritsche (changes requested):
I started with a review but achieving this while keeping the code readable required quite some refactoring so I opened a [PR](https://github.com/patricklizon/user-event/pull/1).

src/keyboard/keyboardImplementation.ts
```diff
@@ -53,23 +55,31 @@ export async function keyboardImplementation(
...
+        // don't consume again on the next iteration
+        consumedLength: 0,
+        keyDef,
+        releasePrevious,
+        releaseSelf,
+        repeat: repeat - 1,
+      }
+    } else {
+      delete state.repeatKey
     }
```
ph-fritsche:
```suggestion
```

src/keyboard/keyboardImplementation.ts
```diff
@@ -53,23 +55,31 @@ export async function keyboardImplementation(
         keypress(keyDef, getCurrentElement, options, state)
       }
 
-      if (releaseSelf) {
+      // Release the key only on the last iteration on `state.repeatKey`.
+      if (releaseSelf && repeat <= 1) {
         keyup(keyDef, getCurrentElement, options, state, unpreventedDefault)
       }
     }
   }
```
ph-fritsche, line 64:
```suggestion

    if (repeat > 1) {
      state.repeatKey = {
        // don't consume again on the next iteration
        consumedLength: 0,
        keyDef,
        releasePrevious,
        releaseSelf,
        repeat: repeat - 1,
      }
    } else {
      delete state.repeatKey
    }
```

ph-fritsche (changes requested):
I just noticed that we might leak state into another call if the last key def has a repeat modifier and `keyboardState` is reused on another call.

ph-fritsche (approved):
Looks good to me now.

ph-fritsche:
@patricklizon Could you file a PR for the docs?
https://github.com/testing-library/testing-library-docs/blob/main/docs/ecosystem-user-event.mdx

Issue #729, opened 2021-10-01
Performance regression: hasPointerEvents increases click time


jesperorb:
- `@testing-library/user-event` version: `13.2.1`

- Testing Framework and version: `jest@27.0.2`
  
- DOM Environment: `jest-environment-jsdom@26.6.2`
  

Relevant code or config

```typescript

import userEvent from "@testing-library/user-event";
userEvent.click(await app.findByText("Test"));
```

What you did:

Clicked an element using `@testing-library/user-event`

What happened:

Went slower than expected

Problem description:

I upgraded to a newer version of this library that contains the following PR: https://github.com/testing-library/user-event/pull/631

More specifically it contains the following LOC:
https://github.com/testing-library/user-event/blob/673c2257ef63f69dd46b9fe2d4bbc3146db1426b/src/utils/misc/hasPointerEvents.ts#L11

If I understand correctly it loops over the tree and using `window.getComputedStyle(el)` on every one. In our tests this seems to be quite costly depending on what you are clicking and how much you are rendering.

Below is an earlier version of `user-event` (13.0.0, before the PR had been merged). We do a bunch of clicks and every click clocks in about 10-40ms. The aggregated time for this specific test lands in about 500ms.
![Screenshot 2021-10-01 at 10 02 34](https://user-images.githubusercontent.com/21122051/135588497-5b4974d8-5524-4777-bac1-3af03979b4d1.png)

Below is the version we are using: 13.2.1. As you can see every click has increased and differs a bit of course but around 150-200ms for each click. And it is because of the checking of pointer-events. In my case over a large set of tests it adds minutes.
![Screenshot 2021-10-01 at 10 05 52](https://user-images.githubusercontent.com/21122051/135588757-623a396f-e461-4121-bace-9fceb2634704.png)

Solution description:

I do not know how to solve this. The pointer-events check is a good check to have but is there a more effecient way to check it?

ph-fritsche:
> but is there a more effecient way to check it?

Sadly: no.
The browser issues no pointer events if the element or any of the parents contain a `pointer-events: none` directive.
But there is no API telling if this is the case for some element, so we have to traverse the tree.

As the cost increases with the depth of the tree, the best way to optimize this is to run most tests only on the relevant components and then a few expensive tests on the whole tree.
You could also write a setup function that renders the components either isolated or as part of the whole app. So that you could run the quick version in dev and in CI on pushs and the complete version in CI on PR or special branches.

[...]

PR #731, opened 2021-10-03
feat: add flag to skip pointer-events check

jesperorb:
**What**:

This PR adds a new flag called `skipPointerEventsCheck` which makes it possible to skip checking for `pointer-events: "none"` when using different user events like `click` or `dblclick`.

This PR does **not** add a global config flag as per discussion in: https://github.com/testing-library/user-event/issues/729

**Why**:

As per discussion in: https://github.com/testing-library/user-event/issues/729

Adding the check for `hasPointerEvents` increase the time to execute each click and hover or anything that relies on `pointer-events`. The increased time for each click can be quite drastic depending on how large of a DOM-tree you are rendering in your tests as it checks if any parent has the property set. 

Making it optional can speed up tests. If you for example know that the property is not set in the code your are testing. Or if you want to disable it momentarily when developing and the run it later on in the CI.

**How**:

Adding a common interface for pointer events that has the property `skipPointerEventsCheck` and using this flag anywhere the check `hasPointerEvents` is being used.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged
      

**Feedback and questions**

- What counts as Documentation in the checklist?

### DIFF
--- a/src/__tests__/click.js
+++ b/src/__tests__/click.js
@@ -478,3 +478,9 @@ test('throws when clicking element with pointer-events set to none', () => {
   const {element} = setup(`<div style="pointer-events: none"></div>`)
   expect(() => userEvent.click(element)).toThrowError(/unable to click/i)
 })
+
+test('does not throws when clicking element with pointer-events set to none and skipPointerEventsCheck is set', () => {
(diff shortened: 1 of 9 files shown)


ph-fritsche:
@all-contributors add @jesperorb code

Issue #694, opened 2021-07-05
Support PageUp/PageDown special keys


mammadataei:
Hi. I couldn't find out how I can trigger `pageUp`/`pageDown` keys using the `keyboard` function and these are not in `specialChars` too.

ph-fritsche:
`specialChars` is legacy.
I recommend referencing non-printable keys should [per `key` or `code`](https://github.com/testing-library/user-event#keyboardtext-options).

A PR adding missing keys to the default key map would be welcomed:
https://github.com/testing-library/user-event/blob/402944134d10a33a9837be25ca8b1991f510fc6d/src/keyboard/keyMap.ts#L73-L78

Implementations of default browser behavior for control keys go here:
https://github.com/testing-library/user-event/blob/master/src/keyboard/plugins/control.ts

JohannesFischer:
Happy to have a stab at this. I assume this is to be used in input and textarea elements, is that correct?

JohannesFischer:
Had some time and went ahead adding support for the two keys. Please have a look if you have a minute and let me know if that is what you where going for for.

PR #734, opened 2021-10-06
feat: add support for pageup and pagedown control keys

JohannesFischer:
**What**:
Adding support for `PageUp` and `PageDown` control block keys as requested in issue #694 

**Why**:
To support more possible user inputs.

**How**:
Add `PageDown` and `PageUp` to keymap. The behavior equals the of `Home` and `End` keys.

**Checklist**:
- N/A Documentation
- [x] Tests
- [x] Ready to be merged
      
Keys need to be added to [special characters](https://testing-library.com/docs/ecosystem-user-event/#special-characters) when this gets merged.

### DIFF
--- a/src/__tests__/type.js
+++ b/src/__tests__/type.js
@@ -1024,6 +1024,101 @@ test('navigation key: {home} and {end} moves the cursor', () => {
   `)
 })
 
+test('navigation key: {pageUp} and {pageDown} moves the cursor for <input>', () => {
+  const {element, getEventSnapshot} = setup('<input />')
+  userEvent.type(element, 'c{pageUp}ab{pageDown}d')
+  expect(getEventSnapshot()).toMatchInlineSnapshot(`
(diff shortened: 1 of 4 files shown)


ph-fritsche:
Thanks for contributing to this library❤️

If I recall correctly, `ArrowUp` and `ArrowDown` have not been implemented earlier, because the exact new position of the cursor was hard to determine due to possible line wrapping.
What is the behavior with `PageUp` / `PageDown` on long text content in `teaxtarea`?

ph-fritsche:
I think the cursor movement won't be exactly the same on different platforms in a real browser environment either.

Maybe we should not encourage assuming the cursor position after one of these keys is used.
Even a close-to-real estimate how far the cursor might be moved would depend heavily on the effective width of the content-box which is not reliable in a no-layout environment.
@nickmccurdy What do you think?

The added entries in the `keyMap` are much needed and the changes for `input` look good to me.

PR #774, opened 2021-11-03
feat(keyboard): select all per `{Control}+[KeyA]`

ph-fritsche:
**What**:

Select all per key combination.

**Why**:

Default behavior in the browser.

**How**:

Add a behaviorPlugin for `keydown` on `{Control>}[KeyA]`.

**Checklist**:
- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

**Additional information**
This requires proper handling of selection ranges.
https://github.com/testing-library/user-event/blob/3fa471f511663607807aa9e792446d67c4f27586/src/utils/edit/selectionRange.ts#L15

Renders our implementation of `userEvent.keyboard('{selectall}')` obsolete which allows to remove the only `specialChar` and `behaviorPlugin` that has no corresponding behavior in the browser.
Also allows us to replace the implementation of `userEvent.clear()`.


ph-fritsche:
See https://github.com/testing-library/user-event/issues/778#issuecomment-966393903
This behavior is triggered by `Cmd`+`A` on Mac. Maybe this should be made configurable like the `keyboardMap` in the future.

PR #784, opened 2021-11-15
feat!: remove support for user provided `MouseEventInit`

ph-fritsche:
BREAKING CHANGE: `init` parameter has been removed from these APIs:
- `userEvent.click`
- `userEvent.dblClick`
- `userEvent.tripleClick`
- `userEvent.hover`
- `userEvent.unhover`
- `userEvent.selectOptions`
- `userEvent.deselectOptions`

BREAKING CHANGE: `userEvent.upload` no longer supports `clickInit`
as part of its `init` parameter.

**What**:
Remove `init` parameters for `EventInit` of mouse/pointer events.

**Why**:
The `init` parameters are escape hatches.
They were used for different events that don't necessarily share the same properties.

The new pointer implementation renders supplying own `MouseEventInit` properties obsolete as the events are created according to specs / living standard.
If our code lacks non-deprecated features from the specs, we should add them.
If our users want to test features outside of the specs, they should be on their own using lower level implementations.

The `init` parameter encourages bad practice as it makes it convenient to just tweak the events instead of describing a user interaction that *might* lead up to those events.

**Checklist**:
- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged -- The code is ready but I'd like a second opinion on this.


kentcdodds (approved):
I'm in favor of this change. I never liked this API 👍

Issue #504, opened 2020-11-28
Re-thinking approach to async event handling


nickmccurdy:
See https://github.com/testing-library/dom-testing-library/pull/616#issuecomment-643711133

We've been discussing some differences in Testing Library behavior with asyncronous renders and data fetching. The official Vue Test Utils docs (which Vue Testing Library uses) recommend awaiting its events so the library could internally wait for Vue to finish rendering. However, User Event is used with some libraries like React and Angular that don't have async renders like this. This causes potential race conditions with the UI when awaiting things like fireEvent or userEvent.

I'm starting to feel that perhaps this test example shouldn't rely on the race condition being fixed. For example in a real world app, if a web app loads very quickly its loader may not be displayed (perhaps it's already cached or the website has a threshold to prevent it from loading). A real human manually testing a website wouldn't give up after not seeing the loader, they would just continue as long as the page changed and they would wait for a loader if they saw one. As a result, I don't think `waitForElement` is the right function to make sure a loader disappears, especially since frameworks like React may batch updates when async functions are mocked. What matters here is that the element isn't displayed, so `.not.toBeInTheDocument()` is more accurate and reliable.

I think that instead of trying to get User Event to run the original code without modifications, we should update waitForElement's error message for initial null elements to explain that the element may have already been removed and give alternatives like using Jest DOM's `.not.toBeInTheDocument()` instead, and potentially document other common issues. Then we'll be able to make User Event functionality async, which makes it more consistent both with real user behavior and some frameworks like Vue and Svelte which already deal with async renders in tests. It's also worth noting that because userEvent isn't async, its APIs are harder to wrap for async Vue tests and users have to learn more implementation details of Vue as a result (see https://github.com/testing-library/vue-testing-library/pull/182).

CC @kentcdodds @afontcu @ITEnthusiasm

kentcdodds:
Thanks for bringing this up. I've never felt comfortable making all of these synchronous. The user never fires these events synchronously, so I don't like that these fire synchronously. But you explain the problem with the alternative really well. Specifically about asserting on the loading state. The suggestion is:

> What matters here is that the element isn't displayed, so `.not.toBeInTheDocument()` is more accurate and reliable.

[...]

PR #797, opened 2021-12-03
fix(tab): skip elements with `visibility:hidden`

Dennis273:
**What**:

Calling `userEvent.tab` focuses on element with `visibility: hidden`, which is not expected.

**Why**:
It breaks use cases related to `visibility:hidden` and `userEvent.tab`

**How**:

It is fixed by checking whether the target element has style `visibility: hidden`.

**Checklist**:

- [N/A] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/tab.js
+++ b/src/__tests__/tab.js
@@ -324,14 +324,8 @@ test('should stay within a focus trap', () => {
     document.querySelector('[data-testid="div1"]'),
     document.querySelector('[data-testid="div2"]'),
   ]
-  const [
-    checkbox1,
-    radio1,
-    number1,
-    checkbox2,
-    radio2,
-    number2,
-  ] = document.querySelectorAll('[data-testid="element"]')
+  const [checkbox1, radio1, number1, checkbox2, radio2, number2] =
+    document.querySelectorAll('[data-testid="element"]')
 
   expect(document.body).toHaveFocus()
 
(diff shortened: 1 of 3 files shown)


ph-fritsche:
Thanks a lot for your contribution :heart: 

I'm not sure yet if it's worth the merge conflict with [v14.0.0-beta](https://github.com/testing-library/user-event/tree/v14.0.0-beta). We'll drop support for `v13` as soon as `v14` is considered stable.
What do you think about targeting `beta` with this change?

Dennis273:
@ph-fritsche 

It'll be nice if this issue can be fixed on `v13` since we'll still be using `v13` for a while before we can migrate to `v14`

I can make another PR to `beta`  fixing this issue later this week.

ph-fritsche:
I've merged the other PR. Would you share why you expect the migration to be put on hold for a while?
We merged the remaining PRs on `main` before starting `alpha` and pushed through all breaking changes so that if the need for a fix or feature arises, anyone with that need could switch to the new version.
This change is small and the fixed code exists similarly in both branches so it might be reasonable to merge this in both. But I'm a little hesitant to even start gauging upcoming changes in that regard.

Dennis273:
🤔    We have to make tests using `userEvent` from syncronous to asyncronous to use `v14`. I was not sure whether it is worthless making those changes. But the discussion in #504 make great sense so I'm willing to do the migration as soon as `v14` is released.

I'll close this PR.

PR #823, opened 2022-01-04
feat(pointer)!: introduce `pointerEventsCheck` option

ph-fritsche:
BREAKING CHANGE: `skipPointerEvents` has been removed.
Use `pointerEventsCheck: PointerEventsCheckLevel.Never` instead.

**What**:

Replace boolean `skipPointerEvents` with bit flag `pointerEventsCheck`.
Move pointer-events check from convenience APIs  to `pointer` API.

**Why**:

The pointer-events check had been implemented on the pointer related APIs which have one target element.
Direct calls to `pointer` didn't implement the check as the `pointer` API does not necessarily have *one* target element. Therefore just moving or replicating the old implementation in pointer was impossible.
Just adding checks wasn't feasible as we know how expensive the check can be (especially in Jsdom).

**How**:

Let a "trigger" be one UI interaction that possibly leads to multiple events. E.g. releasing a mouse button which can trigger `pointerup`, `mouseup`, `click` ...  
Move check into `pointer` API and call it on every "trigger" on every target element.

Add internal `ApiLevel` refs and reorganize the export so that multiple internal API calls from one user API call can be treated as one call.

Cache the result of the pointer-events check on the element.

Allow the user to specify how often we actually perform the check or potentially use a cached result per `pointerEventsCheck` option:
https://github.com/testing-library/user-event/blob/4b6a58e63bf3a1614061dc1689748e461535922c/src/options.ts#L6-L18

This PR includes a drive-by fix for options per main `setup`. These have been dropped before.

**Checklist**:
- [ ] Documentation
- [x] Tests
- [x] Ready to be merged


ph-fritsche:
@jesperorb You implemented the previous option. What do you think about this change?

jesperorb:
> @jesperorb You implemented the previous option. What do you think about this change?

I think it is a good change. As disabling it is not the ideal solution and more granualar control is a good, The flag is a better solution than just a boolean. 👍🏽 

As I saw in my profiling it was a combination of the click being slow but also that a click triggered a hover or some other event. Which doubled the time although the check for pointer events should be the same, the elements are the same. And this change seems to adress that so that is nice :)

That said, I didn't understand everything but I think I got the gist of it. But I can't really visualize how I would use every option of the flag practically although I understand it conceptually.

ph-fritsche:
Thanks for the feedback.
Yeah I don't know how much each of the values will be used. (That's why I discarded to add this on the event-level for now. Might reconsider if someone actually has a use case for this.)

`EachApiCall` should be a fine default. If an element becomes non-interactive per `pointer-events` between `userEvent.anyApi()` calls this will throw.

`EachTarget` ignores that an element might change its `pointer-events` and keeps the value across multiple calls (and possibly tests) as long as that element still exists. This might be an acceptable optimization for most projects.

`Never` can be used if one knows that there can be no `pointer-events` style.

`EachTrigger` should allow to test cases where `pointer-events` is changed in a pointer event handler.

PR #825, opened 2022-01-10
docs: streamline README

ph-fritsche:
**What**:

Streamline the README.

**Why**:

The published README should always be up-to-date and quickly point people in the right direction.

**How**:

* Remove table of content. With the API documentation removed from the README, the ToC is longer than most sections. It does no longer improve readability of the README.
* Remove the duplicate links to the docs. The docs are already linked at the top of the README.
* Remove installation section. The installation is described in the docs with further explanation.
* Remove known limitations. The list gives the impression to be exhaustive. This is not the case and maintaining an exhausting list in the README is not feasible.
* Move list of contributors to own file. The README is part of the published package, but the contributors to each version are usually added after the fix/feature was merged. This results in the package shipping with a list of contributors that is usually at least one step behind.

**Checklist**:
- [x] Ready to be merged


timdeschryver (approved):
Looks good to me.
The only nit I have is that this config is a bit different from the other TL libraries, because these have the contributors added to the README.

ph-fritsche:
Yes, I guess a nicer way to resolve this would be a Github Action:
Get code and test contributors from commits.
Get bug/feature contributions from issues.
Get review contributions from pull requests.
Automatically add a docs commit before the release.

Is it worth it though?

timdeschryver:
> Is it worth it though?

Probably not 😅

PR #862, opened 2022-02-24
fix(paste): skip input according to `defaultPrevented`

### LINKED ISSUE #861: Paste event ignores `preventDefault`
- `@testing-library/user-event` version: 14.0.0-beta.11

- Testing Framework and version: jest@26.6.3
- DOM Environment: jsdom@11.5.1

What you did:

I used the new `userEvent.paste` API to test a form in my application. This form intercepts paste events to split contents among numerous fields, so it is critical that it prevents the clipboard contents from being inserted into the focused input field.

What happened:

During the test, the clipboard contents were all inserted into the focused input element. Calling `event.preventDefault` in the `onPaste` handler had no effect.

Reproduction repository:

I would be happy to create a reproduction repository if that would be useful, but I saw a similar unit test here in this repository that I might be able to copy as a simpler representation of the problem. I will pursue that first instead.

Problem description:

When using the new `userEvent.paste` API with the focus on an editable element, calling `event.preventDefault` in the `onPaste` handler has no effect. There is no way to prevent the clipboard contents from being inserted into the editable context.

Suggested solution:

As is already the case for other APIs in this library such as `upload`, the default behavior of the paste event should be skipped if `preventDefault` is called on the event.

It looks like a similar problem with the `upload` API was recently fixed in this PR: https://github.com/testing-library/user-event/pull/847
I am hoping this problem is able to be fixed in a similar manner.

Gudahtt:
**What**:
The paste event now uses the central event dispatching system introduced in #847 to implement the effects of the `paste` event, which ensures the effects are skipped if `event.preventDefault` is called.

**Why**:
This ensures that the effect of the `paste` event is ignored if `event.preventDefault()` is called on the paste event. This matches the behavior of the paste event in the browser, and is necessary to test sites that rely on this behavior. This fixes #861.

**How**:
The behavior of the `paste` event (which is to insert the clipboard content into the current context if it is editable) has been moved from the `paste` module to a separate module in the `behavior` directory. This effect of pasting is now called by the event dispatching system, which also ensures this effect is suppressed if `event.preventDefault` is called.

**Checklist**:

- [X] Tests
- [X] Ready to be merged

### DIFF
--- a/src/event/behavior/paste.ts
+++ b/src/event/behavior/paste.ts
@@ -1,0 +1,17 @@
+import {input, isEditable} from '../../utils'
+import {behavior} from './registry'
+
+behavior.paste = (event, target, config) => {
+  if (isEditable(target)) {
+    return () => {
+      if (event.clipboardData) {
+        input(
+          config,
+          target,
+          event.clipboardData.getData('text'),
+          'insertFromPaste',
+        )
+      }
+    }
+  }
+}
--- a/src/event/behavior/paste.ts
+++ b/src/event/behavior/paste.ts
@@ -4,13 +4,9 @@ import {behavior} from './registry'
 behavior.paste = (event, target, config) => {
   if (isEditable(target)) {
     return () => {
-      if (event.clipboardData) {
-        input(
-          config,
-          target,
-          event.clipboardData.getData('text'),
-          'insertFromPaste',
-        )
+      const insertData = event.clipboardData?.getData('text')
+      if (insertData) {
+        input(config, target, insertData, 'insertFromPaste')
       }
     }
   }
--- a/tests/clipboard/paste.ts
+++ b/tests/clipboard/paste.ts
@@ -12,7 +12,7 @@ test('paste with empty clipboard', async () => {
   expect(getEvents('input')).toHaveLength(0)
 })
 
-test('paste with file data', async () => {
+test('do not trigger input for paste with file data', async () => {
   const {getEvents, user} = setup(`<input/>`)
 
   const f0 = new File(['bar'], 'bar0.txt', {type: 'text/plain'})
--- a/tests/clipboard/paste.ts
+++ b/tests/clipboard/paste.ts
@@ -78,15 +78,12 @@ test('does not paste when disabled', async () => {
 })
 
 test('prevent input per paste event handler', async () => {
-  const {element, getEventSnapshot, user} = setup(`<input />`)
+  const {element, eventWasFired, user} = setup(`<input />`)
   element.addEventListener('paste', e => e.preventDefault())
 
   await user.paste('hi')
-  expect(getEventSnapshot()).toMatchInlineSnapshot(`
-    Events fired on: input[value=""]
-
-    input[value=""] - paste
-  `)
+  expect(eventWasFired('paste')).toBe(true)
+  expect(eventWasFired('input')).toBe(false)
 })
 
 test.each(['input', 'textarea'])(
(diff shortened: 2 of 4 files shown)


src/event/behavior/paste.ts
```diff
@@ -0,0 +1,17 @@
+import {input, isEditable} from '../../utils'
+import {behavior} from './registry'
+
+behavior.paste = (event, target, config) => {
+  if (isEditable(target)) {
+    return () => {
+      if (event.clipboardData) {
```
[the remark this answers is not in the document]
Gudahtt: Oh, this is causing test to fail because of the drop in coverage.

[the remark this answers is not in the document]
Gudahtt: Ah, good point re: the `input()` util. I will try this. This provides a path for testing it as well, as I can test the `else` case by pasting non-text data.

tests/clipboard/paste.ts
```diff
@@ -65,6 +77,18 @@ test('does not paste when disabled', async () => {
   )
 })
 
+test('does not paste when preventDefault is called', async () => {
```
ph-fritsche, line 80:
```suggestion
test('prevent input per paste event handler', async () => {
```
In fact it does `paste` it just doesn't manipulate the DOM.

tests/clipboard/paste.ts
```diff
@@ -65,6 +77,18 @@ test('does not paste when disabled', async () => {
...
+  const {element, getEventSnapshot, user} = setup(`<input />`)
+  element.addEventListener('paste', e => e.preventDefault())
+
+  await user.paste('hi')
+  expect(getEventSnapshot()).toMatchInlineSnapshot(`
+    Events fired on: input[value=""]
+
+    input[value=""] - paste
+  `)
```
ph-fritsche, line 89:
```suggestion
  const {element, eventWasFired, user} = setup(`<input />`)
  element.addEventListener('paste', e => e.preventDefault())

  await user.paste('hi')
  expect(eventWasFired('paste')).toBe(true)
  expect(eventWasFired('input')).toBe(false)
```
Let's prefer explicit assertions over snapshots.

tests/clipboard/paste.ts
```diff
@@ -11,6 +12,17 @@ test('paste with empty clipboard', async () => {
   expect(getEvents('input')).toHaveLength(0)
 })
 
+test('paste with file data', async () => {
```
ph-fritsche, line 15:
```suggestion
test('do not trigger input for paste with file data', async () => {
```

ph-fritsche (changes requested):

tests/clipboard/paste.ts
Gudahtt: Oops, I knew that, not sure how I came up with that wording. Thanks!

ph-fritsche (approved):
LGTM. Thanks for taking the time to contribute to this project :heart:

ShaneYu:
Any idea when this PR is going to be merged and a new package released? As far as I can tell, there is no workaround(s) at the moment for this issue and is therefore needing much complicated tests to achieve something simple.

ph-fritsche:
@all-contributors add @Gudahtt bug, code

laboriosi:
I have the same question and I reread the entire documentation and I couldn't find a solution.

In version 13, when I wanted to paste something into a specific input I would do:
```jsx
import userEvent from '@testing-library/user-event'
import { screen } from '@testing-library/react'
import Screen from '..'

it('should paste', () => {
  render(<Screen />)
    
  userEvent.paste(screen.getByRole('textbox'), '1234')
})
```

But in version 14 it is not clear how this can be done, as now paste no longer receives the element

PR #872, opened 2022-03-04
fix: check for inherited `:disabled`

### LINKED ISSUE #844: Events ignore disabled property in fieldset
- `@testing-library/user-event` version: 13.5.0
- Testing Framework and version: jest 27.4.7
- DOM Environment: jsdom (version unknown)
  
- node version: 16.13.2

What you did:  test input checkbox in a disabled fieldset

What happened: checkbox is modified even though fieldset is disabled

Reproduction repository: https://codesandbox.io/s/busy-sky-u1thc

It works in the browser, but fails when tested with testing-library. 

Testing-library understands that the checkbox is disabled, but userEvent appears to ignore it.

It looks like #96 resolved it for disabled checkboxes but didn't take fieldsets into account.

ph-fritsche:
**What**:

Check if an element is disabled according to specs.

**Why**:

Closes #844 

**How**:

Traverse the element and ancestors and check if any of them cause the element being disabled.

`:disabled` pseudo-class should yield the correct value but doesn't in JSDOM.

**Checklist**:
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/utils/misc/isDisabled.ts
+++ b/src/utils/misc/isDisabled.ts
@@ -1,5 +1,37 @@
-// This should probably be extended with checking the element type
-// https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled
+import {isElementType} from './isElementType'
+
+// This should probably just rely on the :disabled pseudo-class, but JSDOM doesn't implement it properly.
 export function isDisabled(element: Element | null): boolean {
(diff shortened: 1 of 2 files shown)


ph-fritsche:
https://codesandbox.io/s/frosty-shape-jbsj54?file=/src/App.test.js

PR #877, opened 2022-03-05
Fix URL to keyboard parsing

joshunger:
Fix URL to keyboard parsing that was broken


ph-fritsche:
We might want to output the appropriate link for the API used by the user so that this doesn't cause confusion.

Inferring from the function name fails with the current build:
https://github.com/testing-library/user-event/commit/97b929651b5e26f9a630912c28f858201ab34263
Maybe we should just pass down which API uses the util in the arguments.

joshunger:
Nice! Feel free to close this PR. Sounds good to me.

ph-fritsche:
Closing in favor of #881

PR #892, opened 2022-03-31
fix: work around shadowed globals

ph-fritsche:
**What**:

Remove direct references to global constructors for DOM(-related) implementation.

**Why**:

Codesandbox environment has shadowed global constructors which break `instanceof` checks.
See https://github.com/codesandbox/codesandbox-client/issues/6580

**How**:

Get the `window` from the elements/document.
Get the constructor from that `window`.

**Checklist**:
- [x] Tests
- [x] Ready to be merged


ph-fritsche:
https://codesandbox.io/s/userevent-react-forked-8tu34o?file=/src/App.js

PR #898, opened 2022-04-01
fix: track calls to `HTMLInputElement.select()`

### LINKED ISSUE #897: Selection issue with .type()
### Reproduction example

https://codesandbox.io/s/testing-lib-user-event-897-4q5egq?file=/src/App.test.js

### Prerequisites

1. Render an `input` that automatically selects its own value on focus
2. `.type()` in the input (with the default `skipClick: false`)
3. Observe selection is correct in the browser, but incorrect in tests (which passed in v13)

### Expected behavior

I expected calling `await user.type(input, text)` would result in a focused input (due to the implied click in the default usage where skipClick is false), which in turn would result in the value of the input being selected (input.selectionStart = 0, input.selectionEnd = input.value.length). This is what happens in the DOM, but now fails in tests (as of `@testing-library/user-event` v14).

### Actual behavior

input.selectionStart is set to input.value.length (unexpected) and input.selectionEnd is set to input.value.length (expected).

### User-event version

14.0.3

### Environment

Testing Library framework: `@testing-library/react@14.0.3`

JS framework: `react@18.0.0`

Test environment: `jest@27.5.1`

DOM implementation: `jsdom@16.6.0`

### Additional context

Similar to #894

ph-fritsche:
**What**:

Support setting the UI selection per `HTMLInputElement.select()`.

**Why**:

Closes #897 

**How**:

Add an interceptor on the method and change `UISelection` accordingly.

**Checklist**:
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/document/selection.ts
+++ b/src/document/selection.ts
@@ -69,6 +69,19 @@ export function prepareSelectionInterceptor(
       return {realArgs: v}
     },
   )
+
+  prepareInterceptor(
+    element,
+    'select',
+    function interceptorImpl(this: HTMLInputElement | HTMLTextAreaElement) {
+      this[UISelection] = {
+        anchorOffset: 0,
+        focusOffset: getUIValue(element).length,
+      }
+
+      return {realArgs: [] as []}
+    },
+  )
 }
 
 export function setUISelection(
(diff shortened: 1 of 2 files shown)


ph-fritsche:
Fixed sandbox:
https://codesandbox.io/s/testing-lib-user-event-897-forked-plu3wz?file=/src/App.test.js

Issue #904, opened 2022-04-05
Enable setting global default setup options


CreativeTechGuy:
### Problem description

This is related to #833.

If we are using fake timers in all tests, that means that we now must specify `{ delay: null }` in every test when we call `userEvent.setup`.

### Suggested solution

In the spirit of [dom-testing-library's configure method](https://testing-library.com/docs/dom-testing-library/api-configuration) it'd be nice to have a top-level function which can set the default options for `userEvent.setup` so we don't have to pass them in every time.

Additionally, having an option to bypass the fake timers for this library rather than having to turn off the delay entirely would be even better! (I'd argue this should be default since it's a very difficult bug to track down when upgrading this library and suddenly having tests timeout.)

CreativeTechGuy:
I took a shot at implementing what I'm imagining here and when using it to upgrade my projects, it was a much better experience. I'd love to hear what you think!

ph-fritsche:
First of all thanks for taking the time to contribute to this library :heart: 

Let's try to keep issues separate. That makes it easier to discuss them and reference them later. (No harm in filing multiple short issues at once.)

I fear a global config does more harm than good.
We want to motivate people to use `userEvent.setup` after the DOM is ready because it ensures that our workarounds are applied (see #714 ) and it gives a clear entry point where manually reenacting a test workflow would start which makes it a lot easier to help people with their tests.
If you want to apply a config across all your tests, wrapping `render` and `userEvent.setup` in a [setup function](https://testing-library.com/docs/user-event/intro/#writing-tests-with-userevent) and applying your config there is probably the best way to solve this.

Regarding the fake timers: A PR resolving #585 would be welcome.
Note that Jest isn't the only test environment this library is used in. I think the user should supply the callback.
```ts
type Config = {
  // ...
  advanceTimers: (delay: number) => Promise<void>
}
```

CreativeTechGuy:
Understood! I'll separate the fake timers into another PR and we can discuss more over there.

My thought with the global config was a few things:

* It aligns with the rest of the `@testing-library` ecosystem which has the same approach.
* It'll be easier to motivate people to use the new `.setup` API since it'll be easier to use. (less duplication and no need for everyone to write another abstraction) It doesn't replace it or take away any of the benefits of it, it just avoids the duplication given that almost every setup call with have the same arguments by default.
[...]

Issue #585, opened 2021-03-17
feat: advanceTimers callback


ph-fritsche:
@ph-fritsche - Would it be possible to add an `advance` option to type?
The typeImpl function could then call that with the delay before `await` ing on the promise.

As an example of the approach, I'm using this "wrapper" around type (it only works for plaintext):
```
async function typeWithDelay(input, text, delayInMilliseconds) {
  let previous = Promise.resolve();
  for (const codepoint of text) {
    await previous;
    userEvent.type(input, codepoint);

    previous = new Promise((resolve) => setTimeout(() => resolve(), delayInMilliseconds));

    act(() => {
      jest.advanceTimersByTime(delayInMilliseconds);
    });
  }
}
```
So instead of above, I could just call `userEvent.type(input, 'some text', { delay: 20, advance: jest.advanceTimersByTime });`

_Originally posted by @icedtoast in https://github.com/testing-library/user-event/issues/565#issuecomment-800694378_

nickmccurdy:
Tip: You can start block quotes on GitHub by prefixing lines with `>` or pressing the `Insert a quote` button.

ph-fritsche:
@all-contributors add @icedtoast ideas

PR #907, opened 2022-04-05
feat: add `advanceTimers` option

CreativeTechGuy:
**What**:

Implement an `advanceTimers` option from #585 and [the second half of this comment](https://github.com/testing-library/user-event/issues/904#issuecomment-1088366570)

**Why**:

The internal delays in the library don't play nice with fake timers in testing frameworks

**How**:

_I'm not really sure how to answer this question. With code? haha_

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


src/options.ts
```diff
@@ -120,6 +120,13 @@ export interface Options {
    * Defaults to `true` when calling the APIs per `setup`.
    */
   writeToClipboard?: boolean
+
+  /**
+   * A function to be called internally to advance your fake timers (if applicable)
+   *
+   * @example jest.advanceTimersByTime
+   */
+  advanceTimers?: (delay: number) => void
```
ph-fritsche, line 129: I think this should return `Promise<void>`. The resulting `await` on this enables the user to let third-party code run after the timers were advanced and before we continue interacting. Otherwise asynchronous code during the timeout might not be executed until after the next interaction which might result in hard to debug issues.

ph-fritsche (changes requested):

src/utils/misc/wait.ts
```diff
@@ -1,3 +1,11 @@
-export function wait(time?: number) {
-  return new Promise<void>(resolve => setTimeout(() => resolve(), time))
+import type {Options} from '../../options'
+
+export function wait(
+  time: number,
+  advanceTimers: Exclude<Options['advanceTimers'], undefined>,
+) {
```
ph-fritsche, line 6: If we handle timing here, we should accept the Config and take the properties we need.
```suggestion
export function wait(
  config: Config,
) {
```

ph-fritsche (changes requested):

src/options.ts
CreativeTechGuy: The place where this `advanceTimers` function would be used is already inside of a promise and resolving the promise is already from a `setTimeout`. Can you give an example of some code which would need `advanceTimers` to return a promise?

ph-fritsche:
```ts
function wait(t: number) {
    return new Promise<void>(r => setTimeout(r, t))
}
function waitVoidAdvance(t: number, advance: (t: number) => void) {
    return new Promise<void>(resolve => {
        setTimeout(() => resolve(), t)
        advance(t)
    })
}
function waitAwaitAdvance(t: number, advance: (t: number) => Promise<unknown> | unknown) {
    return Promise.all([
        new Promise<void>(resolve => setTimeout(() => resolve(), t)),
        advance(t)
    ])
}

async function impl(
    log: string[],
    waitImpl: (t: number) => Promise<unknown> | unknown
) {
    // event handler
    void (async () => {
        log.push('event')
        await wait(0)
        log.push('then')
        void wait(20).then(() => {
            log.push('after 20')
        })
        await wait(0)
        log.push('next')
        await wait(0)
        log.push('x')
        await wait(0)
        log.push('y')
        await wait(0)
        log.push('z')
        await wait(100)
        log.push('delayed')
    })()

    // delay
    await waitImpl(50)
    log.push('continue')
}

describe.each([false, true])('usePromise: %s', (useAwaitAdvance) => {
    test.each([false, true])('fakeTimers: %s', async (useFakeTimers) => {
        const log: string[] = []
    
        if (useFakeTimers) {
            jest.useFakeTimers()
        }
    
        await impl(
            log,
            useAwaitAdvance
                ? (t) => waitAwaitAdvance(
                    t,
                    useFakeTimers
                        ? async (u: number) => {
                            do {
                                jest.advanceTimersByTime(1)
                                for(let before;;) {
                                    before = jest.getTimerCount()
                                    await new Promise<void>(r => r())
                                    if (jest.getTimerCount() === before) {
                                        break
                                    }
                                    jest.advanceTimersByTime(0)
                                }
                            } while(--u > 0)
                        }
                        : () => Promise.resolve()
                )
                : (t) => waitVoidAdvance(
                    t,
                    useFakeTimers
                        ? jest.advanceTimersByTime
                        : () => undefined
                ),
        )
 
        jest.useRealTimers()
        await wait(10)

        expect(log).toEqual([
            'event',
            'then',
            'next',
            'x',
            'y',
            'z',
            'after 20',
            'continue',
        ])
    })

})
```

ph-fritsche: The code above just demonstrates the problem in a reproducible manner.
In real-world scenarios this usually isn't that obvious, the pushes to the event loop can be implemented in nested/imported modules and/or can involve event-driven APIs.

> ```ts
> (delay: number) => {
>    (async () => {
>        /* do your crazy async work */
>    })();
>}
>```
^ This would not allow the asynchronous code after any `await` to run before we continue with the next interaction that is supposed to happen "later".
The problem is that fake timers create a situation that normally doesn't exist: Microtasks being queued and not executed before picking the next "macrotask". This is because the next "macrotask" is actually queued synchronously while the microtask still is queued through the event loop.

CreativeTechGuy: Got it. I don't understand why anyone would ever do this. But sure, I can make it support promises too if someone wants to do this.

CreativeTechGuy:
@ph-fritsche let me know if there are any other changes you'd like to see. I think this is ready to go though! :)

ph-fritsche (approved):
LGTM. Thanks for the contribution.
Just giving everyone a chance to weigh in on this here or in the docs repo.

stokesman:
This looks like a nice addition 👏 . I wonder if the name shouldn't be more general. I can't think of another use case besides advancing timers but who knows. I think calling it `advance` would be better. Maybe even a callback-sounding name would be worth a look: `onAdvance` `onDelay` `onNext`.

CreativeTechGuy:
@stokesman I'd totally agree with you if there were ideas on how else it could be used. But I'd rather not rename something to be more ambiguous (and possibly confusing) because of some hypothetical future that may not exist. 😄

If you can think of other places in the library where a similar, abstract, version of this pattern could be used, let's do it!

ph-fritsche:
@all-contributors add @CreativeTechGuy code

Issue #839, opened 2022-01-20
_userEvent.default.setup is not a function


Tommoore96:
- `@testing-library/user-event` version: ^14.0.0-beta
- DOM Environment: 8.11.1
- Node version: 14.17.4

Relevant code or config

```javascript
const user = userEvent.setup()
```

What you did:
Tests worked perfectly before updating, I just wanted the latest version for the pointer method.

Followed the latest documentation and tried multiple versions 14 and above.

Imported like so:
```js
import userEvent from '@testing-library/user-event';
```
and initialised like so:
```js
const user = userEvent.setup()
```

What happened:
Tests failed and receive following error:
`TypeError: _userEvent.default.setup is not a function`

ph-fritsche:
`_userEvent.default.setup` suggests that the `import` statement is transpiled before being executed.

Could you please provide a reproduction?

ph-fritsche:
@bijay-ps `userEvent.setup()` was introduced in `v14`.

nickmccurdy:
What build tools are you using (it would help to see your configs as well)?

mattxyzeth:
The same thing happened to me. I'm using typescript with jest Node 16 lts.

```json
"devDependencies": {
    "@testing-library/jest-dom": "^5.16.2",
    "@testing-library/react": "^12.1.3",
    "@testing-library/user-event": "^14.0.0-beta",
    "jest": "^27.5.1",
    "ts-jest": "^27.1.3",
    "ts-node": "^10.5.0"
}
```

```json
{
  "compilerOptions": {
    "allowUmdGlobalAccess": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "noUnusedParameters": false,
    "noUnusedLocals": true,
    "skipLibCheck": true,
    "moduleResolution": "node",
    "removeComments": true,
    "target": "ESNEXT",
    "module": "ESNext",
    "noEmit": true,
    "esModuleInterop": true,
    "jsx": "react"
  },
  "include": ["src"]
}
```

```javascript
export default {
  clearMocks: true,
  collectCoverage: true,
  coverageDirectory: "coverage",
  coverageProvider: "v8",
  preset: 'ts-jest/presets/default-esm',
  testEnvironment: 'jest-environment-jsdom',
  transform: {},
  transformIgnorePatterns: [
    "node_modules/(?!(ts-invariant)/)",
  ],
  extensionsToTreatAsEsm: [
    ".jsx",
    ".ts",
[...]

PR #927, opened 2022-04-13
Use .js extension for build outputs

strmer15:
**What**:
Fixes https://github.com/testing-library/user-event/issues/839 by separating the ESM build and CommonJS builds into different directories, instead of using `.cjs` and `.mjs` extensions.

**Why**:
Some build tools don't work well with the `.cjs` / `.mjs` extensions, so using a `.js` extension is more compatible. Users with existing tooling setup for `.js` don't need to update their configurations to work with the new extensions.

**How**:
The entry points in `package.json` were changed to use `index.js` with the ESM build using an `esm/` directory. The `build.js` script now tells `esbuild` to put the files in the appropriate space with the `.js` extension.

**Checklist**:

- [ ] Documentation - N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/build.js
+++ b/build.js
@@ -1,9 +1,9 @@
 ;(async () => {
   const child = require('child_process')
-  const {build} = require('esbuild')
+  const { build } = require('esbuild')
 
   await build({
-    outfile: 'dist/index.mjs',
+    outfile: 'dist/esm/index.js',
     format: 'esm',
     target: 'es6',
     bundle: true,
@@ -12,7 +12,7 @@
   })
 
   await build({
-    outfile: 'dist/index.cjs',
+    outfile: 'dist/index.js',
     format: 'cjs',
     target: 'node12',
     bundle: true,
--- a/build.js
+++ b/build.js
@@ -1,7 +1,9 @@
 ;(async () => {
   const child = require('child_process')
-  const { build } = require('esbuild')
+  const {build} = require('esbuild')
+  const fs = require('fs')
 
+  // put the ESM output in a separate directory to avoid using the .mjs extension
   await build({
     outfile: 'dist/esm/index.js',
     format: 'esm',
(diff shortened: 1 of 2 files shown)


ph-fritsche:
First of all: Thanks for taking the time to contribute to this library.

The [package.json](https://github.com/testing-library/user-event/blob/3184504f7ea0ec138e0436900b37288287c783cd/package.json) does not have a `type` field. The `dist/esm/index.js` will be [treated as CommonJS](https://nodejs.org/dist/latest-v17.x/docs/api/packages.html#packagejson-and-file-extensions) by Node and any environment following their module resolution.

strmer15:
> First of all: Thanks for taking the time to contribute to this library.
> 
> The [package.json](https://github.com/testing-library/user-event/blob/3184504f7ea0ec138e0436900b37288287c783cd/package.json) does not have a `type` field. The `dist/esm/index.js` will be [treated as CommonJS](https://nodejs.org/dist/latest-v17.x/docs/api/packages.html#packagejson-and-file-extensions) by Node and any environment following their module resolution.

Ah, ok - good point. I was thinking of this in terms of Webpack, which will use either the `module` or the `exports` declarations. I think a good way to deal with this would be to generate a `package.json` file that only contains `{ "type": "module" }` and place it in the `dist/esm` directory. Does that sound good to you?

ph-fritsche:
Sounds as if it *should* work. But the current export *should* work too.
So I guess I'll keep this open for a while. Hopefully some people try [the build from this PR](https://pkg.csb.dev/testing-library/user-event/commit/acbe109c/@testing-library/user-event) in different environments and give some feedback.

pgarciacamou:
I pulled [the build](https://pkg.csb.dev/testing-library/user-event/commit/acbe109c/@testing-library/user-event) from @ph-fritsche (note: it is a compressed .zip file without extension, so you might have to manually add the `.zip` extension to be able to decompress it).

So far is the only thing that has worked for me to solve the `TypeError: _userEvent.default.setup is not a function` issue mentioned in https://github.com/testing-library/user-event/issues/839.

```json
    "@testing-library/dom": "8.13.0",
    "@testing-library/jest-dom": "5.16.4",
    "@testing-library/react": "13.2.0",
    "@testing-library/user-event": "14.2.0",
```

All the jest setups mentioned in the comments or installing "@testing-library/dom" (as you can see above) which was suggested in another other issue by Kent did not work for me.

pgarciacamou:
Any news?

ph-fritsche:
closed in favor of #949

PR #947, opened 2022-05-06
fix: add missing html tag `details` in FOCUSABLE_SELECTOR

zthxxx:
**What**:
The `<details >` can be focused by default, but couldn't pass the test using `user-event`.

**Why**:

Due to it is missing from the helper `FOCUSABLE_SELECTOR` variable in `user-event`.

**How**:

Total focusable(accessibility) native elements see list in [Accessibility concerns docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex#accessibility_concerns)

**Checklist**:

- [N/A] Documentation
- [N/A] Tests
- [x] Ready to be merged

### DIFF
--- a/src/utils/focus/selector.ts
+++ b/src/utils/focus/selector.ts
@@ -7,4 +7,5 @@ export const FOCUSABLE_SELECTOR = [
   '[contenteditable="true"]',
   'a[href]',
   '[tabindex]:not([disabled])',
+  'details',
 ].join(', ')


ph-fritsche:
First of all thanks for taking the time to contribute to this library. :heart: 

I think we need to do more to properly support `<details>`. It is an `interactive element` but it is not `focusable`. The `<summary>` is `focusable` and `tabable`.  It also has `click` events per `Space` and `Enter` key.
The elements inside a `<details>` element without `open` attribute are hidden and need to be excluded to taborder. Interacting with such an element should throw an error.

zthxxx:
@ph-fritsche  Sure? Here is a `details` below, without a `summary`

<details>
  --- something ---
</details>

the code and also the dom are

```html
<details>
  --- something ---
</details>
```

ph-fritsche:
@zthxxx I checked again and the browser behavior is even more complicated:

For a `<details>` without `<summary>` you can focus `<details>`. But in Firefox if you click it (per `Space`/`Enter` or per mouse) `document.activeElement` is updated to `document.body`.
For a `<details>` with `<summary>` the behavior seems to be consistent. `<details>` is not `focusable` while `<summary>` and seems to behave like any other focusable element across browsers.

https://codesandbox.io/s/hungry-feather-ywg7bi?file=/src/App.js

zthxxx:
Thanks for review, I see it's a inconsistent behavior

PR #967, opened 2022-06-07
Improve performance of tabbing

camchenry:
**What**:

This PR changes the implementation of `getTabDestination` to only calculate the visibility of an element "just in time," rather than calculating the visibility of all elements ahead of time.

(As an additional change, this also fixes elements with a `tabindex` less than `-1` from being focusable.)

**Why**:

Currently, `getTabDestination` computes the visibility of all elements ahead of time, which can be a costly operation due to calling `getComputedStyle` repeatedly. Even though `getComputedStyle` only takes a few milliseconds to complete, this time can add up for tests that have many calls to `userEvent.tab()` or lots of focusable DOM elements.

Computing the visibility only for adjacent elements in the tab order saves many calls to `getComputedStyle` which results in dramatically improved performance for calls to `tab`. Assuming there few hidden elements in the tab order, then this makes the visibility run in constant time rather than linear time in the number of focusable elements.

Here is a rough benchmark with just tabbing between a few elements with 100 focusable buttons in the DOM.

**Before**:
Screen Shot 2022-06-07 at 3 58 20 PM
**After**:
Screen Shot 2022-06-07 at 3 59 01 PM

This results in a roughly ~10x (-90%) improvement in the peformance of `userEvent.tab`.

[Benchmark Code](https://gist.github.com/camchenry/cb919375f5cb25b5fbaa52660d5b1d91)

**How**:

- Added a new (internal) utility function to check the focusable state of an element, with an option to skip checking visibility
- Changed `getTabDestination` to use new focusable utility function and check focusability in a loop until we find the next focusable element in the tab order
- Fixes elements with a `tabindex` of less than `-1` by checking for negative `tabindex`, rather than just `-1`

**Checklist**:

- [N/A] Documentation
- [x] Tests
- [x] Ready to be merged


ph-fritsche:
Thanks for contributing to this project. :heart: 

Avoiding unnecessary calls to `getComputedStyle` is a great performance improvement.
It doesn't matter much in the browser where this call is fast, but it is notoriously slow in DOM implementation written in JS (e.g. Jsdom).

I've opened a PR simplifying the implementation: https://github.com/camchenry/user-event/pull/1

camchenry:
> Thanks for contributing to this project. ❤️
> 
> Avoiding unnecessary calls to `getComputedStyle` is a great performance improvement. It doesn't matter much in the browser where this call is fast, but it is notoriously slow in DOM implementation written in JS (e.g. Jsdom).
> 
> I've opened a PR simplifying the implementation: [camchenry#1](https://github.com/camchenry/user-event/pull/1)

Thank you for the suggestions! I have gone ahead and merged those in this branch. I noticed that CI was failing previously due to a typecheck issue, which I'm not sure what that was about.

camchenry:
👋 @ph-fritsche Is there anything else that I can do to help get this merged? I'm more than happy to help contribute any additional changes or tests as needed.

ph-fritsche:
Sorry about the delay. Wanted to dispel one minor concern:

Feeding our logic for radio groups with hidden elements does in fact match what is observed in the browser (Chrome):
If a checked radio is hidden, the group is ignored.

https://github.com/testing-library/user-event/blob/e732ee9994ffa7a2204397bd2ca6dda07166eefc/src/utils/focus/getTabDestination.ts#L42-L44

ph-fritsche:
@all-contributors add @camchenry code

PR #983, opened 2022-07-13
Expose `UserEvent` type

vitalybaev:
**What**:
I exposed `UserEvent` type from `types/setup.ts` to be available in `index.d.ts`

**Why**:
We have been using `user-event` a lot with `eslint-import-resolver-typescript`, but after upgrade of the resolver package, we've got the following error:

> Resolve error: Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath `./dist/types/setup` is not defined by "exports" in `/path/to/project/node_modules/@testing-library/user-event/package.json`

So we should either add `types/setup` to exports or just expose this type in `index.d.ts`.

**How**:
-

**Checklist**:
- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

I think it's ready to be merged.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,4 @@
-export {userEvent as default} from './setup'
+export {userEvent as default, UserEvent} from './setup'
 export type {keyboardKey} from './keyboard'
 export type {pointerKey} from './pointer'
 export {PointerEventsCheckLevel} from './options'


vitalybaev:
> Why does eslint-import-resolver-typescript complain about types subpaths not being part of the export ?

Hmm, I thought, even though TS resolves it, it is correct behavior.

Anyway, as for me, it's better to have all types from the main entry point. Exposing this type allows doing
```ts
import { UserEvent } from '@testing-library/user-event';
```

instead of
```ts
import { UserEvent } from '@testing-library/user-event/dist/types/setup';
```

ph-fritsche:
> Hmm, I thought, even though TS resolves it, it is correct behavior.

Is there any spec you can link to?
TS resolves it and the TS docs don't mention `exports`. But according to comments e.g. on the issues about resolving ESM imports, the correct way is to import a path that is a valid JS import and TS takes it from there.
According to the docs it uses the `types` and `typesVersions` entries in `package.json`.
Because we only support the main export and there is no JS file for the subpath, I don't think adding it to `exports` would be correct.

So if you wanted to use this subpath, you would need to use `import type` to declare that this import is not part of the resulting JS.
```ts
import type { UserEvent } from '@testing-library/user-event/dist/types/setup'
```

But what do you need this type for in the first place?

(I mean you could just retrieve from our default export, but I don't understand the use case yet.)
```ts
import UserEvent from '@testing-library/user-event'
type UserEventAfterSetup = ReturnType<(typeof UserEvent)['setup']>
```

Issue #1014, opened 2022-08-02
The upgrade from 14.3.0 to 14.4.0 of @testing-library/user-event no longer supports node 14


sesam:
### Reproduction example

Not a website

### Prerequisites

Reproduce
1. Run with node 14, so e.g. `nvm use 14`
2. run any test that imports this library

### Expected behavior

With no major version bump this package should still be possible to load.

### Actual behavior

It fails, saying `??=` isn't supported.

### User-event version

14.4.0

### Environment

~Testing Library framework:~

~JS framework:~

~DOM implementation:~

### Additional context

Suggest to either make this a major version bump,

or look into what went wrong with the build pipeline change in this version -> https://github.com/testing-library/user-event/commit/fd8f2194b28848923858a0404b09044a89af0781

ph-fritsche:
Thanks for the report. :heart:

ph-fritsche:
@all-contributors add @sesam bug

PR #1015, opened 2022-08-02
Avoid usage of logical nullish assignment for node 14 compatibility

alexander-schranz:
**What**:

Replace logical nullish assignement `test ??= other` with `test = test ?? other`.

**Why**:

For better compatibility.

See https://github.com/testing-library/user-event/issues/1014

**How**:

Adopting the source files.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/system/keyboard.ts
+++ b/src/system/keyboard.ts
@@ -98,7 +98,7 @@ export class KeyboardHost {
     const target = getActiveElementOrBody(config.document)
     this.setKeydownTarget(target)
 
-    this.pressed[code] ??= {
+    this.pressed[code] = this.pressed[code] ?? {
       keyDef,
       unpreventedDefault: false,
     }
--- a/src/system/pointer/index.ts
+++ b/src/system/pointer/index.ts
@@ -41,7 +41,7 @@ export class PointerHost {
     private registry = {} as Record<string, Device>
 
     get(k: string) {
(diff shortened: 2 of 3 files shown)


src/system/keyboard.ts
```diff
@@ -98,7 +98,7 @@ export class KeyboardHost {
     const target = getActiveElementOrBody(config.document)
     this.setKeydownTarget(target)
 
-    this.pressed[code] ??= {
+    this.pressed[code] = this.pressed[code] ?? {
```
alexander-schranz, line 101: not sure about the `eslint` as that would mean if I understand correclty the `??=` is unnecessary here and we not even need a fallback?

src/system/pointer/index.ts
```diff
@@ -41,7 +41,7 @@ export class PointerHost {
     private registry = {} as Record<string, Device>
 
     get(k: string) {
-      this.registry[k] ??= new Device()
+      this.registry[k] = this.registry[k] ?? new Device()
```
alexander-schranz, line 44: same as above here 🤔

ph-fritsche:
Thanks for contributing and trying to fix this :heart: 

But I'll fix it in the build script instead of the source. Otherwise the problem would just be reintroduced later.

Issue #1019, opened 2022-08-03
Rethink testing strategy


ph-fritsche:
### Problem description

1. We run linting and tests on multiple node versions that we support according to
  https://github.com/testing-library/user-event/blob/ea5023141740123aede05561913cb2ec14952e2b/package.json#L13-L14
  This has no merits.
  Our source code has no dependency on any NodeJS API, so that the result for the source itself will always be the same.
  The tested code is not the distributed code, so that if the different NodeJS versions yield different results, it only flags problems with our testing environment but don't indicate anything about the build we would distribute from that commit.

2. We don't test our build. Misconfigurations or bugs in our build tools result in broken builds being published and any potential fix is verified only by manually testing the build.

3. We only test on Jsdom. We try to be platform agnostic, but out automation doesn't help us to develop software that works as intended in different environments, and the differences between Jsdom and e.g. Chrome are too numerous and wide for relying on manual testing only.

### Suggested solution

- [ ] Remove source tests on older node versions.
- [ ] Split up our tests in
  a) those that rely on mocks or test internals (unit) and
  b) those that interact with the DOM (integration/e2e - I don't think making a distinction makes much sense here).
- [ ] Refactor the e2e tests so that they could be run in different environments.
- [ ] Run the tests in a headless browser during CI.
- [ ] Add at least a smoke test on different node versions using the build.

nickmccurdy:
I was thinking of getting some Testing Library projects running on test frameworks other than Jest. Vitest would probably be a good one to focus on, as it's fast, easy to configure, mostly compatible with Jest, and supports enabling/disabling global injection.

ph-fritsche:
Can you leverage Vitest to run tests in a headless browser?

MatanBobi:
@ph-fritsche great initiative, I highly agree with all of the points.
Small thing about 3, I agree that the differences between JSDOM and Chrome are enormous (don't forget about happy-dom too), but our code is based on the standards and not on a specific engine implementation. If we'll also run tests using happy-dom for examples, some of our tests won't work and that's because happy-dom has some missing implementation.
So what's our suggestion here? Should we run all our tests on a headless browser too? Running our tests on JSDOM is nice because a big part of our users are using JSDOM to run our library.

ph-fritsche:
I don't know if we should run (some?) tests in happy-dom too, but I think we need to run our tests in a headless browser.
In the end the browser has been our point of reference all the time.

[...]

Issue #1025, opened 2022-08-06
Allow to pass an element to paste, copy, cut methods


Christian24:
### Problem description

Hello,

We want to use this library to write component tests for our web components. Since web components use builtin components within them (an `input` for example), the element that is currently focused might be different from the element we want to simulate user input on. 

A simple example of this is a text field. A text field is web component with an `input` inside of it. If you will it just wraps the `input` element. So while the text field is focused, the user input is actually supposed to happen on the `input`. Sadly currently with Shadow DOM enabled there is no way to use `paste`, because it just uses `document.activeElement`: https://github.com/testing-library/user-event/blob/ab1a6c44f7a5895f7650dd6f97c160eca5b2ac43/src/clipboard/paste.ts#L13
 

### Suggested solution

It would be great to have a utility API similar to `type()` to which you can just pass the element directly: https://testing-library.com/docs/user-event/utility/#type

This would give people using Shadow DOM a way to use this library.

### Additional context

It looks like Shadow DOM support is still further away: https://github.com/testing-library/dom-testing-library/issues/413

ph-fritsche:
Thanks for opening this issue.

We should aim to support Shadow DOM throughout all APIs. There are actually some lines in the internal utils to explicitly handle shadow hosts, but I think the considerations weren't properly documented (neither in the code nor in an issue or PR), so it's unclear which parts of the library already interact correctly with Shadow DOM and what would be necessary to add full support. (That is within the constraint to only handle `open` Shadom DOM, as providing workarounds for testing `closed` Shadow DOM would be another issue.)

The problem with any `Utility API` is that they introduce room for interpretation. Wherever possible we want to avoid this and provide APIs which describe a workflow that could manually be reproduced in a browser to provide a source of truth for the correct behavior.

Would you rewrite this or file a new issue that targets (open) Shadow DOM support in the existing APIs?
If you could research how focus and selection are handled in the browser and if there are other things to be considered, that would be a huge contribution towards this feature. :green_heart:

Christian24:
Thanks for getting back to me. I opened a new issue: https://github.com/testing-library/user-event/issues/1026

> The problem with any `Utility API` is that they introduce room for interpretation. Wherever possible we want to avoid this and provide APIs which describe a workflow that could manually be reproduced in a browser to provide a source of truth for the correct behavior.
> 

[...]

Issue #1002, opened 2022-07-28
Inputs are not focused when clicked (v14)


IanVS:
### Reproduction example

https://github.com/IanVS/storybook-testing-user-event-click-focus

### Prerequisites

1. Using V14, try to click an input field: `await userEvent.click(getByRole('textbox'));` for example

### Expected behavior

The input element should be fully focused, ready to accept user input (flashing cursor).

### Actual behavior

The input is half-way focused, with a focus ring, but no flashing cursor.

### User-event version

14.3.0

### Environment

Testing Library framework: `@testing-library/react@13.3.0`

JS framework: `react@18.2.0` 

Test environment:  storybook@6.5.9

DOM implementation: browser

### Additional context

This was working correctly in V13.  I have instructions in the README of the reproduction of how to verify.

ph-fritsche:
Are you sure this is an issue with `user-event` ?
If yes, please break it down to a reproduction example that doesn't wrap `user-event` in third-party code.
If not, please raise this with https://github.com/storybookjs/testing-library. I don't know if their integration is correct and it seems they reverted and postponed the upgrade to `v14`.

IanVS:
I did also experience the issue when using user-event directly, and as I describe in the readme, forcing npm to download v13 causes user-event to work correctly.  But I can modify the reproduction if that's helpful.

IanVS:
I've pushed an update to the reproduction to import from `@testing-library/user-event` directly, with the same behavior.

ph-fritsche:
Thanks for the report.

This is odd:
https://codesandbox.io/s/young-https-oghigy?file=/src/App.js

In Firefox the focus is set into the element per `.click` API, in Chrome it isn't.
But calling `.focus()` directly on the element works in both. Although this is how the target is focused by the click API internally - resulting in the outline that indicates that the element gained focus :thinking: 

`paste()` doesn't work in either browser.

jaworek:
I'm currently working on updating `@testing-library/user-event` in my project and it seems that I've stumbled upon the same bug. At first I thought that this is related to #1018, but this seems to be actual reason. Downgrading to v13 fixes the issue for the time being.

ph-fritsche:
With `v14.4` both the `.click()` and `.paste()` yield the expected result with one exception:
In Chrome the cursor is missing and `.selectionStart`/`.selectionEnd` are always `0` after the click although they should be at `value.length`

aaron-nance:
[...]

PR #1033, opened 2022-08-12
Support cut, copy and paste on open shadow DOM inputs

### LINKED ISSUE #1026: Support open shadow DOM components
### Problem description

Hello,

this is a follow up to #1025. Currently the library doesn't support web components which use shadow dom. This feature request only deals with **open** shadow doms. The underlying issue seems to be that with web components the element that is focused can be different from the one where input needs to take place: Web Components wrap builtin elements on which the input takes place. To get the element to interact with `document.activeElement` seems to be used. In case of web components this would return the web component, while the element that would need to be interacted with (e. g. an input element within the shadow DOM) is hidden from the library. 

 

### Suggested solution

There has been prior art in the form of `getActiveElement` (https://github.com/testing-library/user-event/blob/main/src/utils/focus/getActiveElement.ts) which iterates over potential shadow DOM's `activeElement` property, until the element to interact with is reached. 

This could be used as a replacement for `document.activeElement` to make the library open shadow DOM compatible. I quickly hacked this together on `paste()` and it seems to get the job done: 
https://github.com/Christian24/user-event/blob/f4242842c1ed51d2ec8042263bee7abdbafea21e/src/clipboard/paste.ts#L14

In some cases like keyboard input it seems to be more an issue of getting the element to focus properly first. 

Interestingly, `user.click(element)` doesn't seem to focus the input within the web component, not sure why. I had to manually `querySelector` the input to get my paste test to work: 
https://github.com/Christian24/user-event/blob/main/tests/webcomponents/index.ts#L29

### Additional context

I have been playing a little bit with this this afternoon: https://github.com/Christian24/user-event 
Not sure it will evolve into a PR though, would like to hear seasoned contributors opinion first. Also for some reason I couldn't get the tests to debug in VS Code...

Christian24:
**What**:
This allows support for copy, cut & paste in inputs in open shadow dom custom elements. 

**Why**:
Because I filled #1025 and #1026. This doesn't completely resolve #1026 though, however it should ease the situation when using native inputs in custom elements.   

**How**:
Changed the implementation of getting the activeElement. 

**Checklist**:

- [ ] Documentation (I am uncertain about this, maybe the state of #1026 should be at some point be part of the documentation?)
- [x ] Tests
- [ x] Ready to be merged


src/clipboard/copy.ts
```diff
@@ -1,10 +1,10 @@
 import {copySelection} from '../document'
-import type {Instance} from '../setup'
-import {writeDataTransferToClipboard} from '../utils'
+import {Instance} from '../setup'
+import {getActiveElement, writeDataTransferToClipboard} from '../utils'
 
 export async function copy(this: Instance) {
   const doc = this.config.document
-  const target = doc.activeElement ?? /* istanbul ignore next */ doc.body
+  const target = getActiveElement(doc) ?? doc.activeElement ?? doc.body
```
ph-fritsche, line 7:
```suggestion
const target getActiveElementOrBody(doc)
```
We should use the util that already provides this fallback.

tests/dom/customElement.ts
```diff
@@ -84,3 +85,94 @@ test('types text inside custom element', async () => {
...
+  const inputElement = element.shadowRoot?.querySelector('input')
+  if (inputElement) {
+    await user.click(inputElement)
+  }
+  await user.keyboard('test')
+
+  expect((element as ShadowInput).value).toEqual('test')
+})
+
+test('Render open shadow DOM element - copy', async () => {
```
ph-fritsche, line 124:
```suggestion
describe('on shadow DOM', () => {
  test('copy', async () => {
```

tests/dom/customElement.ts
```diff
@@ -84,3 +85,94 @@ test('types text inside custom element', async () => {
...
+
+  const inputElement = element.shadowRoot?.querySelector('input')
+  if (inputElement) {
+    await user.click(inputElement)
+  }
+  await user.keyboard('test')
+
+  if (inputElement) {
+    setUISelection(inputElement, {anchorOffset: 0, focusOffset: 4})
+  }
```
ph-fritsche, line 137: The tests should make clear which feature is tested and be narrowed down to this - i.e. no other user-event API being used. Otherwise it might be hard for the next developer to determine what is tested, what lacks tests and which tests are redundant. See the comment in this particular file above.

tests/dom/customElement.ts
```diff
@@ -1,5 +1,6 @@
 import userEvent from '#src'
-import {addListeners} from '#testHelpers'
+import {addListeners, setup} from '#testHelpers'
+import {setUISelection} from '#src/document'
 
 // It is unclear which part of our implementation is targeted with this test.
 // Can this be removed? Is it sufficient?
```
ph-fritsche, line 6: I don't think this test suite should be extended.

ph-fritsche:
First of all: Thanks for contributing to this library. This is much appreciated :heart: 

I think we should add tests to `tests/clipboard/*.ts` testing just the interaction that happens on the `.copy()`/`.cut()`/`.paste()`.
If we feel like `.click()` and `.keyboard()` should have additional tests to, we should add them there.
Testing the features separately makes it much easier for future developers to identify if a test breakage is expected on a change. It's really bad if you're working on something, 50 tests break and you have to spend extra time to figure out which of those 50 tests have false assumptions that you're just about to correct and which actually warn you that you're in the process of breaking something else by mistake.

Also the thing we just imply here is the internal abstractions working correctly. So this is probably where the real testing should take place to pave the way for possible other changes required by #1026.
So we need to test `util/focus/getActiveElement and the `document/copySelection` and `event/input` to properly interact with shadow DOM.
If we know about a limitation of the current change and we don't want to wait for all aspects to be resolved, this can also mean adding some tests with comments that outline what is missing.

src/clipboard/copy.ts
Christian24: I had tried this, but it didn't work. I will investigate a bit.

tests/dom/customElement.ts
Christian24: So your idea is to move these tests into copy, paste, cut test suites?

ph-fritsche: The test in this file was introduced in https://github.com/testing-library/user-event/pull/320/files#diff-a1c186e56ef39b3e82d6781e7a934af052817d6ae42102d3c12368c9896e6e07

The thought there might have been to test for `.type()` what you're trying to test for `.copy()/.paste()/.cut()`, that the target for changes (the editable context) is correctly recognized to be inside the shadow tree.
Since then the implementation for `.type()` completely changed and so did the only assertion - an event screenshot.

I'd like to remove/replace this test, but therefore we need to understand which (if any) part of this test is important and which is just noise. If it's really just about recognizing the correct target, we should make sure to assert this on any API in the respective test suite. Maybe we should add a linting rule that ensures that `document.activeElement` is only used per the util that should document the edge cases - which of course we should test on it directly.

Anyway I would not add any new tests under that code comment, it just makes it harder to be confident when removing/replacing it.

Christian24: I am working on moving the tests at the moment. And I am throwing in a few goodies. :)

src/clipboard/copy.ts
Christian24: All of these should be fixed now.

tests/dom/customElement.ts
Christian24: So I added support for `selection` and `focus` in `setup()`, so this is no longer necessary.

tests/dom/customElement.ts
Christian24: So, I moved the tests and removed the test that was there. My reasoning is this: From what I see this is not a test of open shadow dom as the input element was directly passed to `user.type()`. I adapted the simple test I had to make sure that `type()` works directly on open shadow dom elements, although testing is going to be limited since `delegatesFocus` is not implemented in jsdom. I also think we should rather test recognizing the correct target, because the rest should be identical in open shadow dom.

Christian24:
I added some very simple tests for `getActiveElement`. So far I have only dealt with input related stuff. This is the area most relevant to me.

tests/utility/type.ts
```diff
@@ -90,3 +92,14 @@ test('do nothing on disabled element', async () => {
 
   expect(getEvents()).toHaveLength(0)
 })
+
+describe('on shadow DOM', () => {
+  test('type into an input element', async () => {
+    const {element, user} = setup('<shadow-input></shadow-input>')
+
+    // Skip click because delegatesFocus is not implemented in jsdom
+    await user.type(element, 'test', {skipClick: true})
```
ph-fritsche, line 101: This only works in our specific `setup` that includes the focus workaround.
We should make sure this works inside our own tests and for our users.

tests/utils/focus/getActiveElement.ts
```diff
@@ -0,0 +1,32 @@
+import '../../_helpers/shadow-input'
+import {setup} from '#testHelpers'
+import {getActiveElement, getActiveElementOrBody} from '#src/utils'
+
+describe('focus tests with not shadow root', () => {
```
ph-fritsche, line 5: Unnecessary describe block. Not all edge cases are necessarily in the binary `with shadow` vs `without shadow` so this might create a legacy of misleading structure.

tests/utils/focus/getActiveElement.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    expect(getActiveElement(document)).toBe(
+      element.shadowRoot?.querySelector('input'),
+    )
+  })
+  test('focus just body', async () => {
+    setup('<shadow-input></shadow-input>', {focus: false})
+
+    expect(getActiveElementOrBody(document)).toBe(document.body)
+  })
```
ph-fritsche, line 31: This is an anti-pattern. This tests asserts that our test setup is doing nothing while suggesting that has something to do with util being used here.

tests/utils/focus/getActiveElement.ts
```diff
@@ -0,0 +1,32 @@
+import '../../_helpers/shadow-input'
+import {setup} from '#testHelpers'
+import {getActiveElement, getActiveElementOrBody} from '#src/utils'
+
+describe('focus tests with not shadow root', () => {
+  test('focus input element', async () => {
+    const {element} = setup('<input />')
+
+    expect(getActiveElement(document)).toBe(element)
```
ph-fritsche, line 9:
```suggestion
    expect(getActiveElementOrBody(document)).toBe(element)
```

We should test through the util that is probably what should be used in all our APIs when retrieving the focused element.

tests/utils/focus/getActiveElement.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  test('focus should be body', async () => {
+    setup('<input />', {focus: false})
+
+    expect(getActiveElementOrBody(document)).toBe(document.body)
+  })
+})
+
+describe('on shadow DOM', () => {
+  test('focus contained input element', async () => {
```
ph-fritsche, line 20:
```suggestion
  test('get focused element inside shadow tree', async () => {
```

src/clipboard/copy.ts
```diff
@@ -1,10 +1,10 @@
 import {copySelection} from '../document'
-import type {Instance} from '../setup'
-import {writeDataTransferToClipboard} from '../utils'
+import {Instance} from '../setup'
```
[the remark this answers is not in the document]
Christian24: This is a mistake. I will fix it.

src/clipboard/copy.ts
```diff
@@ -1,10 +1,10 @@
...
+import {getActiveElementOrBody, writeDataTransferToClipboard} from '../utils'
 
 export async function copy(this: Instance) {
-  const doc = this.config.document
-  const target = doc.activeElement ?? /* istanbul ignore next */ doc.body
+  const doc: Document = this.config.document
```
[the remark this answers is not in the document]
Christian24: It is not in my environment. I honestly don't know why but my eslint fails saying that the type is any and therefore it is unsafe.

tests/_helpers/shadow-input.ts
```diff
@@ -0,0 +1,52 @@
...
+      this.$input?.setAttribute(name, value)
+    }
+  }
+
+  public get value(): string {
+    return this.$input?.value ?? ''
+  }
+
+  /**
+   * Overwrite focus handling because delegatesFocus is not implemented in jsdom
```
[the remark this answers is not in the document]
Christian24: I don't know. Depends upon what you want to achieve. We used exactly this when Firefox did not yet support `delegatesFocus`. And I know of some libraries which do this too. It is a question of which browser (aka test environment you target). My tests use karma, so overwriting `focus()` is not necessary, `delegatesFocus` would work. However, as a component author if you use `delegatesFocus` or overwrite `focus()` is your choice. I don't think we should make assumptions here, especially ones that wouldn't be true in a real browser. I think this should rather be fixed in jsdom or we should document that `delegatesFocus` is not implemented in jsdom. But I don't think that this is a problem of this library or this library should do something about it. Maybe the comment is misleading?

tests/utility/type.ts
Christian24: Maybe this is more related to #1002? The `click` just means it looses focus.

tests/utils/focus/getActiveElement.ts
Christian24: Removed it.

tests/utils/focus/getActiveElement.ts
Christian24: I tried making it more explicit.

tests/utils/focus/getActiveElement.ts
Christian24: I did change all usages to use `getActiveElementOrBody`. Are users expected to use this? I would love to remove the export for `getActiveElement`, but this would be a breaking change, so holding back on that for now.

tests/clipboard/copy.ts
```diff
@@ -1,5 +1,6 @@
 import userEvent from '#src'
 import {render, setup} from '#testHelpers'
+import '../_helpers/shadow-input'
```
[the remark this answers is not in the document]
Christian24: I made this explicit everywhere.

tests/_helpers/setup.ts
```diff
@@ -28,24 +28,26 @@ export function render<Elements extends Element | Element[] = HTMLElement>(
   if (typeof focus === 'string') {
     ;(assertSingleNodeFromXPath(focus, div) as HTMLElement).focus()
   } else if (focus !== false) {
-    ;(div.querySelector(FOCUSABLE_SELECTOR) as HTMLElement | undefined)?.focus()
+    const element: HTMLElement | null = div.querySelector(FOCUSABLE_SELECTOR)
+    if (element) {
+      element.focus()
+    } else if (div.firstChild) {
+      ;(div.firstChild as HTMLElement).focus()
+    }
```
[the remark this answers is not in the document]
Christian24: No, I assumed it is a custom element. I changed this to check for firstElementChild and check the tag for `hyphen`, thus making clear it is a custom element and therefore focusable.

tests/_helpers/shadow-input.ts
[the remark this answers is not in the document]
Christian24: I changed the comment a bit to make more clear that this is actually functionality of this test component and not a hack.

tests/_helpers/setup.ts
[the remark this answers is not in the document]
ph-fritsche: Not every custom element is focusable. And the behavior depends on `ShadowRoot.delegatesFocus` or the overloaded `HTMLElement.focus`.
This util is supposed to provide a good default for our tests that makes it easy for future developers to write tests that focus on the tested feature without polluting the tests with too much arrange logic.
I want to make sure there is nothing subtle about this function. It currently focuses the first focusable element in the rendered fragment by default.
Let's add a function that ensures this behavior when one or more elements with shadow DOM are used here.

tests/_helpers/shadow-input.ts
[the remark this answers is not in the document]
ph-fritsche: I just checked in the browser: If I click the `shadow-input` outside the nested `input` (you need to change the style to achieve this, but this doesn't matter here) in the browser, the `document.activeElement` does not change - the `focus` method is not triggered.
Therefore this behavior depends on our test setup calling `focus` on this element which would not be triggered by any user interaction in the browser.

If a workaround for Jsdom not implementing `ShadowRoot.delegatesFocus` is needed, this should be applied when setting up the environment.

tests/utility/type.ts
ph-fritsche: Which is correct. `shadow-input` attaches the shadow DOM with `delegatesFocus: false` . Clicking the `shadow-input` outside of `input` focuses the closest focusable element - which is `document.body` here.
If this is not intended, we need to fix the component and/or our implementation instead of bending the test ;)

#1002 is environment-specific. While it still needs someone to investigate why it happens and how we might be able to fix it, it doesn't affect our tests running in Jsdom.

tests/utils/focus/getActiveElement.ts
```diff
@@ -0,0 +1,37 @@
...
+import {setup} from '#testHelpers'
+import {getActiveElementOrBody} from '#src/utils'
+
+test('focused input element', async () => {
+  const {element} = setup('<input />')
+
+  expect(getActiveElementOrBody(document)).toBe(element)
+})
+
+test('focus should be body', async () => {
```
ph-fritsche, line 14:
```suggestion
test('default to body as active element', async () => {
```

tests/utils/focus/getActiveElement.ts
```diff
@@ -0,0 +1,37 @@
...
+    )
+  })
+  test('focus just body', async () => {
+    defineShadowInputCustomElementIfNotDefined()
+    const {element} = setup<ShadowInput>('<shadow-input></shadow-input>', {
+      focus: false,
+    })
+    element.blur()
+    expect(getActiveElementOrBody(document)).toBe(document.body)
+  })
```
ph-fritsche, line 36: It still doesn't test anything about `getActiveElementOrBody`. It merely ensures that after `element.blur()` the test setup is the same as in the other test with `{focus: false}` above.

Christian24: What I wanted to test is that nothing is currently focused. I don't see how this is wrong, but I can remove the test if you want.

ph-fritsche: The test suggests that `getActiveElementOrBody` should have a specific behavior on this state of the document. But the created state is exactly the initial state which is already covered by a test above. Therefore it just obfuscates the requirements for `getActiveElementOrBody` instead of describing them.

Christian24: Gotcha. I removed the test.

src/clipboard/copy.ts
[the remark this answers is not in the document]
Christian24: Fixed.

tests/_helpers/setup.ts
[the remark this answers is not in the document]
Christian24: At work we only have custom elements that are focusable. I just tried this in a variety of browsers and got different results per browser. I am super confused now ;) I will try to write a little helper, so we can figure out the focusable element and call it a day.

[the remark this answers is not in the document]
ph-fritsche:
```ts
function findFocusable(container: Element | ShadowRoot) {
  for(const el of container.querySelectorAll('*')) {
    if (el.matches(FOCUSABLE_SELECTOR)) {
      return el
    } else if (el.shadowRoot) {
      const f = findFocusable(el.shadowRoot)
      if (f) {
       return f
      }
    }
}
```

tests/_helpers/shadow-input.ts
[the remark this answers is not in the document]
Christian24: How would you apply that though?

tests/_helpers/setup.ts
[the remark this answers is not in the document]
Christian24: I added a helper myself which also skips searching inside the shadowRoot if delegatesFocus is defined.

tests/_helpers/shadow-input.ts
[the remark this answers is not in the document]
Christian24: Removed since the helper it is no longer necessary.

tests/utility/type.ts
Christian24: I removed the test. This looked like a quick win, but it is not it seems, so removing this out of this PR.

tests/_helpers/setup.ts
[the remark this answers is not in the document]
ph-fritsche: `ShadowRoot.delegatesFocus` can be `false`.
`ShadowRoot.delegatesFocus` can be `true`, but the element could not contain a focusable element.
A focusable element inside a shadow tree might precede or follow on a focusable element outside of it.

tests/utility/type.ts
```diff
@@ -95,11 +95,12 @@ test('do nothing on disabled element', async () => {
...
+    const {element, user} = setup<ShadowInput>('<shadow-input></shadow-input>')
 
-    // Skip click because delegatesFocus is not implemented in jsdom
+    // Skip click because the element is already focused
     await user.type(element, 'test', {skipClick: true})
 
-    expect((element as ShadowInput).value).toEqual('test')
+    expect(element.value).toEqual('test')
   })
 })
```
ph-fritsche, line 106: If the test skips the click because it wouldn't work with it, we should add a test describing the missing implementation on `.pointer()` and move the typing test to `.keyboard()`.

tests/_helpers/setup.ts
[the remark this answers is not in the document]
Christian24: This is true. My idea was to fallback to the engine's implementation if there is one. I think overriding this behavior in general if it exists in the browser engine is a bad idea. It is true though, we will not know what the element will do with the focus.

tests/utility/type.ts
Christian24: I removed this, because my primary focus is cut, copy and paste. Maybe this can be part of another PR.

tests/_helpers/setup.ts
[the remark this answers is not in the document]
Christian24: Done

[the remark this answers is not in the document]
ph-fritsche: The goal is to provide a state to the document that can be used to test a certain feature.
The util needs to provide the same state in different environments.
We will probably run these tests in Jsdom and in a browser. See #1019

tests/utility/type.ts
ph-fritsche: But the test was there before. We shouldn't just remove it to merge this.
We need to make sure that we have a good understanding of what is missing and needs to be fixed later if we don't deal with it in this PR.

tests/_helpers/setup.ts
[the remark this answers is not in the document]
Christian24: I am not convinced. This feels like reimplementing the browser. If I want to test pasting support in my custom element for example, I rely on the proper element being focused. What the browser does on `delegatesFocus` is important to get the proper element focused. 

I may want to use karma to run this test in a real browser and I would assume that `delegatesFocus` would be used, because that's why I am running it in a real browser. While in reality what would happen would be that `getFocusableElement` would be used, yet as a user I would be unaware of this. If our implementation differs from the browser's implementation people would never notice this, because we would always use our implementation without them knowing.

tests/utility/type.ts
Christian24: As I said I assumed it was a quick win. It was not, so I removed it for now. It is not part of what this PR describes anyway.

tests/_helpers/setup.ts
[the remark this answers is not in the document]
ph-fritsche: > If I want to test pasting support in my custom element for example, I rely on the proper element being focused. What the browser does on delegatesFocus is important to get the proper element focused.

That's exactly the issue here. The setup util provides a certain document state, e.g. `elementA` being focused, so that a test can assert a feature is correctly implemented, e.g. the pasted content being inserted in `elementA`.
If the setup util provides a different document state, e.g. `elementB` being focused, it is expected that pasted content is not being inserted in `elementA`.

Now that `elementA` has to get focused on some interaction is a different feature. On that one it is important that there is a certain document state, e.g. `elementA` not being focused.

If the document state created by the setup util depends on the environment, this might make it very difficult to find the relevant bit when multiple tests break and point at different features.

tests/utility/type.ts
Christian24: I added a new test for `keyboard`.

tests/_helpers/setup.ts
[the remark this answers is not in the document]
Christian24: Oh, sorry, I was mistaken. `setup` is only really used internally right? Then I agree this is a good idea. Sorry for the confusion. I updated the code.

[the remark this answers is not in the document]
ph-fritsche: Yes, it's part of the `#testHelpers`. When we implement handling focus on the `mousedown` on the shadow tree, we'll need to provide the same behavior as the browser would.

tests/_helpers/setup.ts
```diff
@@ -105,3 +103,25 @@ export function setup<Elements extends Element | Element[] = HTMLElement>(
...
+      const possibleFocusableChildElement = getFocusableElement(
+        child.shadowRoot,
+      )
+      if (possibleFocusableChildElement) {
+        return possibleFocusableChildElement
+      }
+    }
+  }
+  return null
+}
```
ph-fritsche, line 127: Could you explain your preference over:
https://github.com/testing-library/user-event/pull/1033#discussion_r947111604

Christian24: None really. I can also use that if that works better for you. This basically just comes from my tinkering with delegatesFocus.

ph-fritsche: If there is no advantage, i'd prefer the shorter implementation that takes the first focusable regardless of whether it's inside a shadow tree and there are other focusable elements after that.

Christian24: Done

ph-fritsche:
@Christian24 Thanks for your work on this. I'll review again with fresh eyes and try to figure out if - based on what we learned here - there are some easy improvements towards full shadow DOM support that we could add to that release as well.

Christian24:
> @Christian24 Thanks for your work on this. I'll review again with fresh eyes and try to figure out if - based on what we learned here - there are some easy improvements towards full shadow DOM support that we could add to that release as well.

Thanks for taking the time to review it!

Christian24:
@ph-fritsche is there something I can do to help get this, #1038, #1039, #1040 merged?

ph-fritsche:
@Christian24 I'd like to resolve #1019 before merging the pending PRs so that we have proper testing utils in place to verify our changes and any future bug reports. Also the CI is broken - as it seems due to conflicts with more recent versions of Typescript and Eslint. We need to fix these before we can move forward.
I opened a draft for updating the test environment in #1081 . Any help on this would be highly welcome.

Christian24:
@ph-fritsche Sorry, I got somewhat swallowed. I would like to help. I will try to rebase this to get started.

PR #1038, opened 2022-08-18
feat(pointer): move focus in shadow DOM

ph-fritsche:
**What**:

**Why**:

One step closer to full shadow DOM support #1026 

**How**:

**Checklist**:
- [x] Tests
- [ ] Ready to be merged


ph-fritsche:
@Christian24 Interestingly we'll need to apply a workaround both on Jsdom and the browser to handle `delegatesFocus`. Invoking `shadowHost.focus()` does not delegate the focus in nested shadow trees while the `mousedown` which we will use this for does delegate through multiple shadow hosts to the first focusable in the browser.

Christian24:
This entire conversation has caused me into looking into focus handling in Shadow DOM in a different light. It is a mess. It seems to work oddly differently everywhere. How do you want to get this consistent? Also do you want to apply the workarounds to the library's tests or in user land? How do you see chances of jsdom implementing `delegatesFocus`? 

A bit of context maybe as well: We use `delegatesFocus` mostly to get an application that can be easily navigated by tab.

ph-fritsche:
> This entire conversation has caused me into looking into focus handling in Shadow DOM in a different light. It is a mess. It seems to work oddly differently everywhere.

That isn't exclusive to this feature. When you work on this library, you get to know a few quirks of browsers and frameworks you might not expect. ;)


> How do you want to get this consistent? Also do you want to apply the workarounds to the library's tests or in user land?

It looks as if the behavior in the browser regarding the user interaction is consistent - according to MDN:
> The `delegatesFocus` read-only property of the [ShadowRoot](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot) interface returns `true` if the shadow root delegates focus, and `false` otherwise.
> If `true`, when a non-focusable part of the shadow DOM is clicked, the first focusable part is given focus, and the shadow host is given any available `:focus` styling.

On `delegatesFocus: true` we need to find the focusable element and apply focus there so that the click behavior matches the behavior in the browser.
I think it makes sense to handle `delegatesFocus: undefined` as `true`. Users who want a different behavior in Jsdom will need to set the property themselves.


> How do you see chances of jsdom implementing `delegatesFocus`?

I filed an issue. But even if it's implemented, I don't think it is relevant enough to be backported. So most users would still use versions without it.


> A bit of context maybe as well: We use `delegatesFocus` mostly to get an application that can be easily navigated by tab.

How keyboard navigation handles shadow trees needs to be researched, but there might be no special handling, so our tab order helper would just need to consider all focusable elements no matter if they are in a shadow tree.

ph-fritsche:
Looks good now.

Changes on top of the other PR:
https://github.com/testing-library/user-event/pull/1038/files/cbe1391f14f8319ec737e48b3878caad80c6f3a5..HEAD

src/event/focus.ts
```diff
@@ -1,30 +1,72 @@
...
-    wrapEvent(() => target.focus(), element)
+  }
+
+  if (target) {
+    if (delegatesFocus(target)) {
+      const effectiveTarget = findFocusable(target.shadowRoot)
+      if (effectiveTarget) {
+        doFocus(effectiveTarget, true, element)
+      } else {
+        // This is not consistent across browsers if there is a focusable ancestor.
```
Christian24, line 31: Should we document this and related gotchas in the docs?

Issue #1048, opened 2022-08-31
Radio inputs - using Arrow keys to navigate focuses incorrect inputs


vgpena:
### Reproduction example

https://codesandbox.io/s/ancient-forest-44xrzm?file=/src/App.js

### Prerequisites

Render a group of at least 3 radio inputs that:
- share the same `name` attribute
- can be navigated to using the keyboard (i.e., are not `disabled`)

### Expected behavior

Codesandbox test: https://codesandbox.io/s/ancient-forest-44xrzm?file=/src/App.test.js

Keyboard user events should follow the patterns outlined [here](https://www.w3.org/WAI/ARIA/apg/example-index/radio/radio-rating.html#kbd_label).

Specifically, `ArrowRight` and `ArrowDown` should advance the focus within the radio group, wrapping back to the beginning if needed. `ArrowLeft` and `ArrowUp` should reverse the focus within the group, wrapping back to the end if needed.

### Actual behavior

Arrow keys send focus to the wrong input in the Radio Group. E.g.,`ArrowRight` send focus through all the inputs to the last input in the group.

### User-event version

14.4.3

### Environment

*Bug is reproducible in CodeSandbox, so skipping this section.*

Testing Library framework: 

JS framework: 

Test environment: 

DOM implementation: 

### Additional context

Radio group arrow functionality was added here - https://github.com/testing-library/user-event/pull/995

ph-fritsche:
Thanks for the report :heart:

sjwilczynski:
@ph-fritsche, @kentcdodds could we get this issue resolved? Also #1198 is a duplicate. There are 3 very similar PRs opened to get it fixed, could one be picked and merged?

sjwilczynski:
That is fair, I tagged you only because I saw you both as committers in the last ten commits to main. Do you know if there is anyone else apart from you and Philipp who could have the time to take a look?

kentcdodds:
I'm thinking maintenance of this library has kind of fallen off I'm afraid.

If you're blocked I recommend just using fireEvent directly or jumping into a more wholistic solution with playwright.

sjwilczynski:
Sad to hear it, particularly as tools like Storybook depend on it but thanks for sharing the suggestions

kentcdodds:
Perhaps someone from storybook would be interested in taking maintenance! 🙏

PR #1049, opened 2022-08-31
Radio inputs - fix arrow button functionality

vgpena:
**What**:

This PR fixes a bug where keyboard navigation through radio inputs did not work as expected (#1048).

**Why**:

This gives us true-to-life outputs when radio inputs are tested with the keyboard.

**How**:

1. *Sending correct directions for up/down arrows* - Based on native browser interactions [plus this W3C spec for a widget based on radio inputs](https://www.w3.org/WAI/ARIA/apg/example-index/radio/radio-rating.html#kbd_label), Down/Right arrows should move us forward in the group, and Up/Left should move us back. I adjusted `keydown.ts` to send the correct directions for Down and Up key events.
2. *Only advancing one element at a time* - The initial behavior seemed to move us through every valid radio input within a group. I've updated `radio.ts` so that we only move forwards or backwards by 1 valid radio input.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/event/radio.ts
+++ b/src/event/radio.ts
@@ -14,13 +14,20 @@ export function walkRadio(
         ? `input[type="radio"][name="${window.CSS.escape(el.name)}"]`
         : `input[type="radio"][name=""], input[type="radio"]:not([name])`,
     ),
-  ).filter(elt => !isDisabled(elt))
-  // 1 or 0 radio buttons found
-  if (group.length <= 1) {
-    return;
+  )
+  for (let i = group.findIndex(e => e === el) + direction; ; i += direction) {
+    if (!group[i]) {
+      i = direction > 0 ? 0 : group.length - 1
+    }
+    if (group[i] === el) {
+      return
+    }
+    if (isDisabled(group[i])) {
+      continue
+    }
+
+    focusElement(group[i])
+    instance.dispatchUIEvent(group[i], 'click')
+    return
   }
-  // Multiple buttons found
-  const nextRadioElement = group[(group.indexOf(el) + direction + group.length) % group.length];
-  focusElement(nextRadioElement)
-  instance.dispatchUIEvent(nextRadioElement, 'click')
 }
--- a/src/event/radio.ts
+++ b/src/event/radio.ts
@@ -2,6 +2,11 @@ import {type Instance} from '../setup'
 import {getWindow, isDisabled} from '../utils'
 import {focusElement} from './focus'
 
+function setNewActiveRadio(instance: Instance, el: HTMLInputElement & {type: 'radio'}) {
+  focusElement(el)
+  instance.dispatchUIEvent(el, 'click')
+}
+
 export function walkRadio(
   instance: Instance,
   el: HTMLInputElement & {type: 'radio'},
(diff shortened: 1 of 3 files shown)


src/event/radio.ts
```diff
@@ -14,19 +14,13 @@ export function walkRadio(
...
-    instance.dispatchUIEvent(group[i], 'click')
+  ).filter(elt => !isDisabled(elt))
+  // 1 or 0 radio buttons found
+  if (group.length <= 1) {
+    return;
   }
+  // Multiple buttons found
+  const nextRadioElement = group[(group.indexOf(el) + direction + group.length) % group.length];
+  focusElement(nextRadioElement)
+  instance.dispatchUIEvent(nextRadioElement, 'click')
```
ph-fritsche, line 25:
```suggestion
  )
  for (let i = group.findIndex(e => e === el) + direction; ; i += direction) {
    if (!group[i]) {
      i = direction > 0 ? 0 : group.length - 1
    }
    if (group[i] === el) {
      return
    }
    if (isDisabled(group[i])) {
      continue
    }

    focusElement(group[i])
    instance.dispatchUIEvent(group[i], 'click')
    return
  }
```

Just adding the missing `return` prevents calling `isDisabled` on irrelevant elements.

tests/event/behavior/keydown.ts
```diff
@@ -359,15 +360,25 @@ cases(
...
       focus: '//input[@value="a"]',
       key: 'ArrowUp',
-      expectedTarget: '//input[@value="d"]',
+      expectedTarget: '//input[@value="e"]',
+    },
+    'ArrowDown around the corner': {
+      focus: '//input[@value="e"]',
+      key: 'ArrowDown',
+      expectedTarget: '//input[@value="a"]',
     },
```
ph-fritsche, line 382: The additional test cases are trivially inferred from the previous cases and the test containing more than two valid targets.

ph-fritsche (commented):
Thanks for taking the time to contribute to this library :heart:

tests/event/behavior/keydown.ts
vgpena: Sounds good!

vgpena:
@ph-fritsche Please let me know if I should do anything to address the [failing linting job](https://github.com/testing-library/user-event/runs/8142781151?check_suite_focus=true). I would volunteer to make the fixes, but the issues are all in files that this PR didn't change.

If it helps in debugging, I get these linting issues when running `npm run lint` locally on [vgpena:pr/radio-keyboard](https://github.com/vgpena/user-event/tree/pr/radio-keyboard) and on [testing-library:main](https://github.com/testing-library/user-event), even though the linting job is passing in CI on `testing-library:main`.

Thanks!

filipw01:
@ph-fritsche @vgpena I'm willing to work on this. If there is something I can do to push this forward just let me know

joshferrell:
Just saw this PR, looks like it's the same as the implementation I had, but I added an additional test to ensure that aria-disabled was still selectable.

https://github.com/testing-library/user-event/pull/1199

Happy for either fix to go through, but would be great to an eye on one of these from a maintainer. Is there anything blocking this particular PR at this point?

joshferrell:
@ph-fritsche Is there anything blocking this PR still? Would be happy to make any edits, looks like this is just blocked on maintainer review.

ph-fritsche (approved):

PR #1064, opened 2022-09-22
fix(upload): accept type should be more lenient

### LINKED ISSUE #1060: userEvent.upload does not work when input `accept` contains space
### Reproduction example

https://codesandbox.io/s/sad-david-5wel5v?file=/src/App.test.js

### Prerequisites

1. Render an file input where `accept` was configured with command AND space (`, `), e.g.
   ```tsx
   <input type="file" accept="image/png, image/jpeg" />
   ```
2. Trigger `userEvent.upload(...)` on this input with a file type that matches the `accept`.

### Expected behavior

`onChange` event should be triggered on the input and `input.files`  should be updated.

### Actual behavior

Neither `input.files` is updated nor `onChange` was fired.

### User-event version

14.4.3

### Environment

Testing Library framework: 

JS framework: 

Test environment: 

DOM implementation: 

### Additional context

The [`accept`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#limiting_accepted_file_types) attribute on file input does accept file types separated by command + space, both in terms of web standard and what browsers can actually handle.

ktmud:
**What**:

Fixes a bug where file upload `input[accept]` is not respected when there are spaces in the `accept` attribute. Closes #1060.

Also raises an error in case the explicitly provided list of files did not have any match. 

**Why**:

HTML attributes are notoriously lenient---if something is valid in HTML, then it should be valid in test cases. The `accept` attribute ignores both letter cases and whitespaces (including new lines). Plus it is very common for users to provide at least 1 space separator for for multi types, e.g. `<input type="file" accept="image/png, image/jpeg" />`.

Raising an error in case of empty filtered file list is more user-friendly than silently fail as a user's test case may fail anyway if an explicit call of `upload` was not successful.

**How**:

The PR handles these special cases using string replacement and case coercion.

**Checklist**:

- [x] <strike>Documentation</strike> N/A as this should be the default expected behavior.
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/utility/upload.ts
+++ b/src/utility/upload.ts
@@ -61,13 +61,22 @@ function isAcceptableFile(file: File, accept: string) {
 
   const wildcards = ['audio/*', 'image/*', 'video/*']
 
-  return accept.split(',').some(acceptToken => {
-    if (acceptToken.startsWith('.')) {
-      // tokens starting with a dot represent a file extension
-      return file.name.endsWith(acceptToken)
-    } else if (wildcards.includes(acceptToken)) {
-      return file.type.startsWith(acceptToken.substr(0, acceptToken.length - 1))
-    }
-    return file.type === acceptToken
-  })
+  return accept
+    .replace(' ', '')
+    .toLowerCase()
+    .split(/,/)
+    .some(acceptToken => {
+      if (acceptToken.startsWith('.')) {
+        // tokens starting with a dot represent a file extension
+        return file.name
+          .toLowerCase()
+          .replace(/\.jpeg$/, '.jpg')
+          .endsWith(acceptToken.replace(/^\.jpeg$/, '.jpg'))
+      } else if (wildcards.includes(acceptToken)) {
+        return file.type
+          .toLowerCase()
+          .startsWith(acceptToken.slice(0, acceptToken.length - 1))
+      }
+      return file.type.toLowerCase() === acceptToken
+    })
 }
--- a/src/utility/upload.ts
+++ b/src/utility/upload.ts
@@ -27,16 +27,11 @@ export async function upload(
   }
   if (isDisabled(element)) return
 
-  const selectedFiles = Array.isArray(fileOrFiles) ? fileOrFiles : [fileOrFiles]
-  const files = selectedFiles
+  const files = (Array.isArray(fileOrFiles) ? fileOrFiles : [fileOrFiles])
     .filter(
       file => !this.config.applyAccept || isAcceptableFile(file, input.accept),
     )
     .slice(0, input.multiple ? undefined : 1)
-
-  if (selectedFiles.length > 0 && files.length === 0) {
-    throw new Error('No files were accepted by the `accept` attribute')
-  }
 
   const fileDialog = () => {
     // do not fire an input event if the file selection does not change
--- a/src/utility/upload.ts
+++ b/src/utility/upload.ts
@@ -59,6 +59,11 @@ export async function upload(
   input.removeEventListener('fileDialog', fileDialog)
 }
 
+// When matching files, browsers ignore case and consider jpeg/jpg interchangeable.
+function normalize(nameOrType: string) {
+  return nameOrType.toLowerCase().replace(/(\.|\/)jpg\b/g, '$1jpeg')
+}
+
 function isAcceptableFile(file: File, accept: string) {
   if (!accept) {
     return true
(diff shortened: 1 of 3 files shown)


src/utility/upload.ts
```diff
@@ -61,13 +66,24 @@ function isAcceptableFile(file: File, accept: string) {
...
+          .toLowerCase()
+          .replace(/\.jpeg$/, '.jpg')
+          .endsWith(acceptToken)
+      } else if (wildcards.includes(acceptToken)) {
+        return file.type
+          .toLowerCase()
+          .startsWith(acceptToken.slice(0, acceptToken.length - 1))
+      }
+      return file.type.toLowerCase().replace('/jpeg', '/jpg') === acceptToken
+    })
```
liamcmitchell-sc, line 88: Your replacements aren't global so will only replace the first instance of each.

Regarding JPEG/JPG; JPEG is the only valid MIME type but after playing with Chrome I see browsers are really lenient. If this is the same as other browsers then lenient matching here makes sense and will save devs from dealing with jpg/jpeg typos that don't matter. On the other hand, I would lean towards stricter tests and writing more lenient code (`accept=".jpg, .jpeg"`).

Suggestion with global replacements and lenient matching:
```js
// When matching files, browsers ignore case and consider jpeg/jpg interchangeable.
function normalize(nameOrType: string) {
  return nameOrType.toLowerCase().replace(/\bjpg\b/g, 'jpeg')
}

function isAcceptableFile(file: File, accept: string) {
  if (!accept) {
    return true
  }

  const wildcards = ['audio/*', 'image/*', 'video/*']

  return normalize(accept)
    .trim()
    .split(/\s*,\s*/)
    .some((acceptToken) => {
      // tokens starting with a dot represent a file extension
      if (acceptToken.startsWith('.')) {
        return normalize(file.name).endsWith(acceptToken)
      } else if (wildcards.includes(acceptToken)) {
        return normalize(file.type).startsWith(acceptToken.replace('*', ''))
      }
      return normalize(file.type) === acceptToken
    })
}
```

liamcmitchell-sc (changes requested):
Just ran into this issue today, thanks for submitting a fix!

src/utility/upload.ts
ktmud: Great suggestions! I incorporated changes but made a smaller adjustment to the regex---just to make it a little stricter just in case `jpg` also appears in the middle of a file or folder name.

liamcmitchell-sc (approved):

src/utility/upload.ts
```diff
@@ -54,20 +59,28 @@ export async function upload(
   input.removeEventListener('fileDialog', fileDialog)
 }
 
+// When matching files, browsers ignore case and consider jpeg/jpg interchangeable.
+function normalize(nameOrType: string) {
+  return nameOrType.toLowerCase().replace(/(\.|\/)jpg\b/g, '$1jpeg')
```
xjamundx, line 64: yes came here to complain about them being case sensitive :D

ktmud:
@ph-fritsche @Gpx @kentcdodds do you mind taking a look at this?

kentcdodds:
This looks good to me. I'm not an active maintainer on this package anymore so I don't think I should merge it, but if you don't get a reply in a day or so then I'll go ahead and merge it 👍

ktmud:
@ph-fritsche @Gpx @kentcdodds can we merge this?

kentcdodds:
Sorry nobody has replied on this PR.

Unfortunately it looks like in the time since this PR was updated, there were some dependencies that got changed which is likely what is causing the build failure. I'm afraid I don't have more time than to just click the "merge" button on this project, but it won't do any good if the build is failing. So this will have to wait until someone can adjust this PR to fix any build issues. Sorry.

ph-fritsche:
@ktmud I'm sorry about the delay here. Some updates in our dependencies blocked our previous CI. We also learned that the previous CI wasn't sufficient to ensure that bug fixes we apply don't cause new problems in other environments.
A new testing environment is already in place, but there still are some uncertainties that need to be resolved before a new version will be released. See the pinned issues and #1091 for details

src/utility/upload.ts
```diff
@@ -27,12 +27,17 @@ export async function upload(
...
+  const files = selectedFiles
     .filter(
       file => !this.config.applyAccept || isAcceptableFile(file, input.accept),
     )
     .slice(0, input.multiple ? undefined : 1)
 
+  if (selectedFiles.length > 0 && files.length === 0) {
+    throw new Error('No files were accepted by the `accept` attribute')
+  }
+
```
ph-fritsche, line 40: Throwing an error here is a breaking change.

ph-fritsche (changes requested):
Thanks for contributing to this library. :blue_heart: 

We should remove the breaking change. Otherwise this looks good.

ph-fritsche (approved):

PR #1070, opened 2022-10-11
Add named export for userEvent

jroru:
**What**:

Adding a new named export for `userEvent`

**Why**:

Allows users who use TypeScript 4.7+ with `moduleResolution` set to `Node16` to avoid configuring `esModuleInterop` to import the default CJS export.

**How**:

By importing the named export instead of the default one

**Checklist**:

- [N/A] Documentation
- [N/A] Tests
- [x] Ready to be merged

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,5 @@
 export {userEvent as default} from './setup'
+export {userEvent} from './setup'
 export type {keyboardKey} from './system/keyboard'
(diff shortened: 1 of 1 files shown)


kentcdodds:
I didn't notice this PR and pushed this support myself yesterday. Sorry for not noticing! Thanks for the help :)

PR #1081, opened 2022-11-23
chore: rewrite tests

ph-fritsche:
**What**:

Edit tests to work outside of Jest.
Add script and CI to run tests in multiple environments.

**Why**:

This allows us to discover conflicts with environments.

#1019 

**How**:

Eliminate dependencies on Jest specific features.
Use `@ph.fritsche/toolbox` to build, serve and test the project code in Chrome and Node+Jsdom.

**Issue with code coverage**:

Without additional work this results in Jest only running one set of dependencies. Therefore the resulting coverage lacks the cases specific to React17.
The coverage threshold has to be removed. We have `codecov/codecov-action` that still warns us about drops in code coverage.

The coverage produced by `@ph.fritsche/toolbox` (per `@swc/core` and `istanbul-lib-instrument`) differs from the coverage produced by `kcd-scripts test` (per `@babel` and ?).
It seems that it reports stricter coverage e.g. on nullish coalescing. I'm not sure if it's safe to just merge these with the coverage from `kcd-scripts`.

In the end the coverage threshold doesn't tell us if the tests are sufficient. Running the tests in other environments provides more value than the coverage threshold, I'd merge this so that we can address the pending PRs as we now have the feedback if they introduce more/new problems in real browsers.

**Checklist**:
- [x] Tests
- [x] Ready to be merged


timdeschryver (commented):
I didn't review all files, but I like the idea behind this PR.
But, I reviewed the CI output, and here I saw two things that need some investigation:

- the build step warns for a circular dependency
- the test step has failed tests, but the step is green

ph-fritsche:
Thanks for the review.

> * the build step warns for a circular dependency

Because they end up in the same IIFE, this isn't a problem here.
The circular dependency probably should be resolved, but this is an issue for `rollup-plugin-node-builtins` - for our use case it's good.

> * the test step has failed tests, but the step is green

Yes, I thought about this for a while.
We'll need time to investigate and fix the issues we have in different environments and a failing CI step would block us until all those issues are resolved. So for now the step is successful if it can create a report, no matter if the report includes failed tests or even errors in the test code.
Also the Toolbox isn't mature enough to replace Jest, so I'd keep the "Test" step with Jest although it should provide a duplicate result of the `Node, DTL8, React18` run.

src/_interop/dtl.ts
```diff
@@ -0,0 +1,5 @@
+/* eslint-disable @typescript-eslint/no-unnecessary-condition */
+
+import def, * as named from '@testing-library/dom'
```
IanVS, line 3: This seems to cause [some problems](https://github.com/testing-library/user-event/issues/1160) since there is no default export from `@testing-library/dom`.  I haven't seen this kind of approach before, just curious, what is the reason for writing it this way?

Issue #553, opened 2021-02-01
.focus() does not fire focus events if window is not focussed


FritsvanCampen:
Browsers (confirmed in Chrome/Firefox) will not dispatch focus events when the window is not focused when calling `.focus()`.
Here is a codepen that demonstrates this behavior: https://codepen.io/fritsvancampen/pen/OJbPaLX

This becomes a problem if you're using a test-runner that uses a real browser to test (such as Karma) and you're running tests locally, or testing with more than one browser at once (ie Chrome and Firefox). Only one browser window can have focus at a time so your code under test is going to behave differently. You can imagine the kind of debug-hell you end up in, it's an actual [Heisenbug](https://en.wikipedia.org/wiki/Heisenbug)!

I understand that if you're using a headless browser or a mocked browser this is not an issue.

Is this a use-case that this library wants to support?

---

I have been tinkering with a workaround but I haven't found a perfect solution but perhaps we don't need a perfect solution.

Instead of calling `.focus()` you can fire the focus events manually (with dispatchEvent/fireEvent), and set the `activeElement`. However the latter doesn't get picked up by CSS which might be a problem because of `:focus` rules.

Another possibility is to fire the focus events _and_ call `.focus()` (to move the `activeElement`). This has the same problem: it also doesn't trigger `:focus` CSS-rules. This also generates duplicate events, which is acceptable for my use case.

Trying to retro-actively apply the workaround to this library is a bit of a headache so an integrated solution would be nice to have. I would find it acceptable to activate this workaround behavior through some kind of config.

Here is a codepen with the workaround: https://codepen.io/fritsvancampen/pen/VwmYVQe

---

I would offer a PR but I can't seem to get the test-suite running (detects 0 tests).

ph-fritsche:
I don't think this is entirely "fixable" as it is immanent to how the real browser applies the concept of focus:
* `document.activeElement` refers to the active element inside that document.
* If the user switches out of the context of a document, the `activeElement` does not change.
* `:focus` and the `blur`/`focus` events refer to the element a user interaction e.g. per keyboard would act upon.

If you *programmatically* focus an element, the browser does change the `activeElement` no matter if the document is the context the user is in. However if there is a `BlurEvent` and/or `FocusEvent` depends on an element actually losing/gaining the focus of user interaction.
1. start with an element A that is both `document.activeElement` and `:focus`.
2. switch out of the context of the document - e.g. focus the address bar or tab to another window
[...]

PR #1092, opened 2023-01-05
New option: `autoFocusDocumentBeforeTyping` to reduce real browser testing flakiness 

KevinBon:
**What**:

Adding a new optional option, `autoFocusDocumentBeforeTyping`, to fix tests running against a real browser (e.g: [jasmine-browser-runner](https://github.com/jasmine/jasmine-browser-runner)) that may be interfered with a real user interacting with their system.

**Why**:

My company's front-end integration tests can be run against a headless or a real browser context.

We noticed that during a real browser test, users can interfere negatively with the test result, simply by focusing the Chrome dev tools or on another software while and before the test is running. When the `.keyboard` interaction is being called, the element won't receive the key events.

**How**:
Send an event to refocus the currently active element or `config.document` before sending `keyboard` events.

**Checklist**:

- [x] Documentation
- [ ] Tests: I didn't find a testing strategy that could replicate the issue
- [x] Ready to be merged
      

** Additional comments **

A similar issue was experienced by https://github.com/testing-library/user-event/issues/553, but I couldn't replicate locally the issue against `.focus()`

## Example 

The test is using `.type` against an input.

1. Run with the page being focused
2. Click on the Chrome Devtools
3. F5, to re-run the test, the test is now failing

Kapture 2023-01-05 at 13 07 46

### DIFF
--- a/src/keyboard/index.ts
+++ b/src/keyboard/index.ts
@@ -1,6 +1,7 @@
+import {fireEvent} from '@testing-library/dom'
 import type {Instance} from '../setup'
 import type {keyboardKey} from '../system/keyboard'
-import {wait} from '../utils'
+import {getActiveElementOrBody, wait} from '../utils'
 import {parseKeyDef} from './parseKeyDef'
 
 interface KeyboardAction {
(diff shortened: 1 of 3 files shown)


ph-fritsche:
resolved by #1252

PR #1109, opened 2023-03-25
Add failing test when reading uploaded file with FormData

esamattis:
**What**:
Just adding a failing test for a bug I found. It seems to be impossible to read a file added to a form with `userEvent.upload()` using `FormData`. See the test.

I was trying the test React Zorm file capabilities with this but got stuck due to this bug

See https://github.com/esamattis/react-zorm/pull/37

**Why**:

I think this should work.

**How**:

It's not.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/tests/utility/upload.ts
+++ b/tests/utility/upload.ts
@@ -1,4 +1,5 @@
-import {setup} from '#testHelpers'
+import {render, setup} from '#testHelpers'
+import userEvent from '#src'
 
 test('change file input', async () => {
   const file = new File(['hello'], 'hello.png', {type: 'image/png'})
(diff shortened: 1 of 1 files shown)


ph-fritsche:
The DOM implementations prevent setting the `HTMLInputElement.files` property to a non-empty value (or manipulating or even creating a `FileList`.
We work around by replacing the property on the element.
The `FormData` constructor populates the object with data from the internal implementations of the input elements though.
We could only replace the `window.FormData` constructor with our own during setup.

esamattis:
Thanks for the reply! I used the FormData replacement as a workaround for React Zorm if interested:

https://github.com/esamattis/react-zorm/pull/37/commits/af51726ed6044e432c27d24d1380c6b18e46e6d4

PR #1132, opened 2023-05-13
fix: error trying to do in on a null value in is ui function call

kalvenschraut:
I am having a hard time narrowing down exactly when this is occurring, but overall the error is easy to address so haven't come up with exact reproduction. If required I can spend more time doing so.

Overall issue seems to be that sometimes the value of an input is set to null which is causing the isUIValue check to be called from the interceptorImpl function. With null being passed in an error is thrown which causes vitest to exit non zero.

My stack is vitest/quasar/happy-dom so maybe an issue somewhere else if it is felt that this function should not take in null as a type.

### DIFF
--- a/src/document/UI.ts
+++ b/src/document/UI.ts
@@ -19,9 +19,9 @@ export type UIValueString = String & {[UIValue]: true}
 export type UISelectionStart = Number & {[UISelection]: true}
 
 export function isUIValue(
-  value: string | UIValueString,
+  value: string | UIValueString | null,
 ): value is UIValueString {
-  return typeof value === 'object' && UIValue in value
+  return typeof value === 'object' && value !== null && UIValue in value
 }
 
 export function isUISelectionStart(


ph-fritsche:
Thanks for contributing to this library.

`HTMLInputElement.value` and `HTMLTextAreaElement.value` are always `string` in browsers. If `happy-dom` returns `null`, I'd consider this a bug in their DOM implementation.
Is this still an issue?

kalvenschraut:
I was maintaining a patch with this PR in it but updating and not replying the patch and there is no longer an issue occurring for me. Possible happy dom fixed this issue, if it pops up again I will open an issue over there.

thanks

robbieaverill:
This change is probably worth having anyway - `typeof value === 'object'` is true when `value` is null, which is confusing and leads to a runtime error as noted in this PR. At the least, it might be worth throwing an explicit warning or error:
```
if (value === null) {
  throw new TypeError('value cannot be null')
}
```

PR #1138, opened 2023-05-25
fix: pressing arrow keys behavior on Radio buttons

### LINKED ISSUE #1048: Radio inputs - using Arrow keys to navigate focuses incorrect inputs

tnyo43:
**What**:

Fix a bug in the behavior of the keyboard operations on the radio buttons.

**Why**:

closes #1048 

**How**:

- updating test cases
  - increase amount of radio buttons in test situations to make the behavior easier to understand
  - fix the assertion of the test cases as it should be
- fix the code to behave correctly

**Checklist**:

- [N/A] Documentation
- [x] Tests
- [x] Ready to be merged
      

**FYI**

I know there is already another PR https://github.com/testing-library/user-event/pull/1049 for this bug. However, it has past several months since it stop its activity.

### DIFF
--- a/src/event/behavior/keydown.ts
+++ b/src/event/behavior/keydown.ts
@@ -28,7 +28,7 @@ const keydownBehavior: {
   ArrowDown: (event, target, instance) => {
     /* istanbul ignore else */
     if (isElementType(target, 'input', {type: 'radio'} as const)) {
-      return () => walkRadio(instance, target, -1)
+      return () => walkRadio(instance, target, 1)
     }
   },
   ArrowLeft: (event, target, instance) => {
@@ -46,7 +46,7 @@ const keydownBehavior: {
   ArrowUp: (event, target, instance) => {
     /* istanbul ignore else */
     if (isElementType(target, 'input', {type: 'radio'} as const)) {
-      return () => walkRadio(instance, target, 1)
+      return () => walkRadio(instance, target, -1)
     }
   },
   Backspace: (event, target, instance) => {


tests/event/behavior/keydown.ts
```diff
@@ -317,6 +317,8 @@ cases(
       <input type="radio" name="" value="nameless2"/>
       <input type="radio" name="group" value="c" disabled/>
       <input type="radio" name="group" value="d"/>
+      <input type="radio" name="group" value="e"/>
+      <input type="radio" name="group" value="f"/>
```
tnyo43, line 321: There were only two available radio buttons actually (the values are `"a"` and `"d"`).
I've added some other elements to make the behavior easier to understand.

src/event/behavior/keydown.ts
```diff
@@ -28,7 +28,7 @@ const keydownBehavior: {
   ArrowDown: (event, target, instance) => {
     /* istanbul ignore else */
     if (isElementType(target, 'input', {type: 'radio'} as const)) {
-      return () => walkRadio(instance, target, -1)
+      return () => walkRadio(instance, target, 1)
```
tnyo43, line 31: As [the W3C specification about radio group](https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio-rating/#kbd_label) describes, pressing the Down Arrow key behaves the same way as the Right Arrow key. "Moves focus to and checks the next radio button in the group" when the Down/Right arrow pressed, so the `direction` of `walkRadio` should be 1.

The same is true for the Up and Left Arrow keys, so the `direction` should be -1.

src/event/radio.ts
tnyo43, line 1: I think the previous implementation is a bit too hard to understand. I have rewritten it to make it easier to understand what each process does.

1. find the starting element
2. get the next element according on the direction. If it's over the last / first element, go to the other end.
3. if the element is...
3.1. disabled, skip it and go to the next one.
3.2. avaiable, and if it is equal to the starting element, do nothing. Otherwise, check it.

tnyo43:
@ph-fritsche @Gpx @kentcdodds can we merge this PR? If I can help something to merge, I will gladly cooperate with you.

ph-fritsche:
resolved by #1049

PR #1151, opened 2023-08-08
Trim accept tokens as in spec.

sandinmyjoints:
**What**:
Trim whitespace around tokens from the `input` element `accept` attribute.

**Why**:
The spec says `accept` is a [set of comma-separated tokens](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#set-of-comma-separated-tokens), and to obtain the list of tokens from the attribute, the user agent must [split the attribute value on commas](https://infra.spec.whatwg.org/#split-on-commas). The [algorithm for splitting](https://infra.spec.whatwg.org/#split-on-commas) says to "[Strip leading and trailing ASCII whitespace](https://infra.spec.whatwg.org/#strip-leading-and-trailing-ascii-whitespace) from token." as step 2.

**How**:
Call `String.trim` on each token.

**Checklist**:

- [na] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/utility/upload.ts
+++ b/src/utility/upload.ts
@@ -61,13 +61,18 @@ function isAcceptableFile(file: File, accept: string) {
 
   const wildcards = ['audio/*', 'image/*', 'video/*']
 
-  return accept.split(',').some(acceptToken => {
-    if (acceptToken.startsWith('.')) {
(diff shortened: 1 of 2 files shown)


ph-fritsche:
resolved by #1064

PR #1165, opened 2023-09-15
respect elements that are hidden by <details>

VinceMalone:
**What**:

stop considering elements hidden by a collapsed `<details>` element to be _visible_ or focusable, via `getTabDestination()`.

for example, `user.tab()` should not focus elements _hidden_ by a collapsed `details` ancestor.

**Why**:

same reasons for [how `.toBeVisible` works in jest-dom](https://github.com/testing-library/jest-dom#tobevisible)

> An element is visible if all the following conditions are met:
> - […]
> - if `<details />` it has the `open` attribute

given the following code:

```html
<details>
  <button id="hidden"></button>
</details>
<details open>
  <button id="visible"></button>
</details>
```

`button#hidden` should not be considered _visible_, nor should it be focusable, but `button#visible` should.

> **Note**
> the single exception for this is a `summary` element that is a direct descendant of a `details` element. the `summary` element acts as an [always visible] trigger for controlling the expanded state of the `details` element.

**How**:

at least for conveniences sake, i copied [the same logic from jest-dom for determining if an element should be considered _visible_ when walking the element tree](https://github.com/testing-library/jest-dom/blob/af1845383ee2cba007f43460104f73409e7618ac/src/to-be-visible.js#L16-L30) (with one exception: looking for the `hidden` attribute — this caused an issue that i think might be a bug in that library)

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/utils/misc/isVisible.ts
+++ b/src/utils/misc/isVisible.ts
@@ -1,18 +1,37 @@
 import {getWindow} from './getWindow'
 
+function isAttributeVisible(element: Element, previousElement: Element | null) {
+  let detailsVisibility
+
+  if (previousElement) {
+    detailsVisibility =
+      element.nodeName === 'DETAILS' && previousElement.nodeName !== 'SUMMARY'
+        ? element.hasAttribute('open')
+        : true
+  } else {
+    detailsVisibility =
+      element.nodeName === 'DETAILS' ? element.hasAttribute('open') : true
+  }
+
+  return detailsVisibility
+}
+
 export function isVisible(element: Element): boolean {
   const window = getWindow(element)
 
   for (
-    let el: Element | null = element;
+    let el: Element | null = element, prev: Element | null = null;
     el?.ownerDocument;
-    el = el.parentElement
+    prev = el, el = el.parentElement
   ) {
     const {display, visibility} = window.getComputedStyle(el)
     if (display === 'none') {
       return false
     }
-    if (visibility === 'hidden') {
+    if (visibility === 'hidden' || visibility === 'collapse') {
+      return false
+    }
+    if (!isAttributeVisible(el, prev)) {
       return false
     }
   }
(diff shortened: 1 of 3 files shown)


tests/utils/focus/getTabDestination.ts
```diff
@@ -76,6 +76,9 @@ test('exclude hidden elements', () => {
         <input hidden />
         <input style="visibility: hidden"/>
         <input style="display: none"/>
+        <details>
```
snowystinger, line 79: these tests are invalid html, a details must have a summary followed by flow content

src/utils/misc/isVisible.ts
```diff
@@ -1,18 +1,37 @@
...
     el?.ownerDocument;
-    el = el.parentElement
+    prev = el, el = el.parentElement
   ) {
     const {display, visibility} = window.getComputedStyle(el)
     if (display === 'none') {
       return false
     }
-    if (visibility === 'hidden') {
+    if (visibility === 'hidden' || visibility === 'collapse') {
```
snowystinger, line 31: where is visibility collapse tested? I removed that or and every test is passing

src/utils/misc/isVisible.ts
```diff
@@ -1,18 +1,37 @@
 import {getWindow} from './getWindow'
 
+function isAttributeVisible(element: Element, previousElement: Element | null) {
+  let detailsVisibility
+
+  if (previousElement) {
+    detailsVisibility =
+      element.nodeName === 'DETAILS' && previousElement.nodeName !== 'SUMMARY'
+        ? element.hasAttribute('open')
```
snowystinger, line 9: better to check `element.open`? you can set a `details` open with js by doing `detailsElement.open = true`

snowystinger (commented):
Thanks for the PR, a few items
If you don't have time or don't recall, no worries, just let me know and I can take it over.

PR #1166, opened 2023-09-17
refac: refactoring setup.ts

ssi02014:
**What**:

Refactoring `setup.ts`

**Why**:

I reused existing declared types and unified the argument order of the functions. 

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/setup/setup.ts
+++ b/src/setup/setup.ts
@@ -18,7 +18,7 @@ import {DirectOptions} from './directApi'
 /**
  * Default options applied when API is called per `userEvent.anyApi()`
  */
-const defaultOptionsDirect: Required<Options> = {
+const defaultOptionsDirect: Config = {
   applyAccept: true,
   autoModify: true,
   delay: 0,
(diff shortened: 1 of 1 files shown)


src/setup/setup.ts
```diff
@@ -18,7 +18,7 @@ import {DirectOptions} from './directApi'
 /**
  * Default options applied when API is called per `userEvent.anyApi()`
  */
-const defaultOptionsDirect: Required<Options> = {
+const defaultOptionsDirect: Config = {
```
ssi02014, line 21: I changed it to the existing defined `Config`.
```ts
export type Config = Required<Options>
```

src/setup/setup.ts
```diff
@@ -168,11 +168,7 @@ export function createInstance(
   }
 }
 
-function getDocument(
-  options: Partial<Config>,
-  node: Node | undefined,
-  defaults: Required<Options>,
-) {
+function getDocument(options: Options, defaults: Config, node?: Node) {
```
ssi02014, line 171: It seems like `createConfig` and `getDocument` take the same arguments, but why not unify the order?

PR #1176, opened 2023-11-03
Verify window exists before resetting clipboard stub

boblauer:
**What**:

Ensure that `globalThis.window` exists before resetting/detaching the clipboard stub in the `afterEach`/`afterAll` hooks.

**Why**:

Using this library in a test suite does not imply that every test in the suite will run in a browser-like environment. I have a test suite that initializes `jsdom` to test client-side code, but then tears it down after the tests. I wanted to use this library, but I couldn't because as soon as a test ran that didn't initialize `jsdom`, this `afterEach` hook would run and throw an error because `globalThis.window` was undefined.

**How**:

Verify that `globalThis.window` exists inside of the hook. I didn't want to put the check inside of `resetClipboardStubOnView` because this file is meant to be run in a browser-like environment, so it seemed most appropriate to put the check inside the hook.

**Checklist**:

- [N/A] Documentation
- [N/A] Tests
- [x] Ready to be merged
      

There were no existing tests related to the `afterEach`/`afterAll` hooks, and I wasn't sure how to go about writing a test for those, so this code does not have any unit tests around it.

### DIFF
--- a/src/utils/dataTransfer/Clipboard.ts
+++ b/src/utils/dataTransfer/Clipboard.ts
@@ -216,10 +216,18 @@ const g = globalThis as {
 }
 /* istanbul ignore else */
 if (typeof g.afterEach === 'function') {
-  g.afterEach(() => resetClipboardStubOnView(globalThis.window))
+  g.afterEach(() => {
+    if (typeof globalThis.window !== 'undefined') {
+      resetClipboardStubOnView(globalThis.window)
+    }
+  })
 }
 
 /* istanbul ignore else */
 if (typeof g.afterAll === 'function') {
-  g.afterAll(() => detachClipboardStubFromView(globalThis.window))
+  g.afterAll(() => {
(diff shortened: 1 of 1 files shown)


src/utils/dataTransfer/Clipboard.ts
```diff
@@ -216,10 +216,18 @@ const g = globalThis as {
...
+    }
+  })
 }
 
 /* istanbul ignore else */
 if (typeof g.afterAll === 'function') {
-  g.afterAll(() => detachClipboardStubFromView(globalThis.window))
+  g.afterAll(() => {
+    if (typeof globalThis.window !== 'undefined') {
+      resetClipboardStubOnView(globalThis.window)
```
abarajithan, line 230:
```suggestion
      detachClipboardStubFromView(globalThis.window)
```

ph-fritsche (approved):

valentinpalkovic:
Is there any chance to get this fix released soon?

PR #1178, opened 2023-11-14
Clipboard - Check for window existence in `afterAll` functions

EgoziE:
**What**:
Added a verification that window exists in functions `resetClipboardStubOnView` and `detachClipboardStubFromView` which are called on `afterAll`.

**Why**:
When testing in SSR environment, window does (and should) not exist, therefore my tests fail.

**How**:
Just added a conditional chaining.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged
      

I believe these functions should not be called in `afterAll` at all unless the user is actively using clipboard functionality.
Thanks 💛

### DIFF
--- a/src/utils/dataTransfer/Clipboard.ts
+++ b/src/utils/dataTransfer/Clipboard.ts
@@ -148,7 +148,7 @@ export function attachClipboardStubToView(window: Window & typeof globalThis) {
 }
 
 export function resetClipboardStubOnView(window: Window & typeof globalThis) {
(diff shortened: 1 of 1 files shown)


EgoziE:
I didn't notice, it's almost a copy of https://github.com/testing-library/user-event/pull/1176
Please accept either PR 🙏🏼

ph-fritsche:
resolved by #1176

PR #1130, opened 2023-05-11
fix: prevent click event on non-focusable control

a0js:
**What**:

Resolves #1057. This change stops click events from being forwarded to non-focusable inputs from labels.

**Why**:

Currently, click events are being forwarded to inputs from labels. This is expected behavior unless the input is disabled or hidden or otherwise not focusable. This behavior break tests that rely on clicking labels of disabled inputs.

**How**:

The function call forwarding the click was moved inside the conditional checking if the control was focusable.

**Checklist**:

- [N/A] Documentation
- [x] Tests
- [x] Ready to be merged
      

Also, updating the "How to Contribute" egghead tutorial link in CONTRIBUTING.md.

Based on @nknapp's [suggestion](https://github.com/testing-library/user-event/issues/1057#issuecomment-1505985844).

PR #1192, opened 2024-01-30
Fix dead link in CONTRIBUTING.md

jakeboone02:
**What**:

Updated link to KCD's PR course on Egghead.io.

**Why**:

Existing link leads to 404 page.

**Checklist**:

- [x] Ready to be merged

I'm wondering if it should be `http://kcd.im/pull-request` instead of the direct link.

Also, the same bad link appears on the `testing-library/jest-dom` contributing page, too:

https://github.com/testing-library/jest-dom/blob/main/CONTRIBUTING.md

I can submit a similar PR for that repo if necessary.


ph-fritsche:
resolved as a drive-by fix in #1130

Issue #1198, opened 2024-02-27
Arrow keys on radio buttons are backwards


joshferrell:
### Reproduction example

https://codesandbox.io/p/devbox/github/joshferrell/user-event-radio-demo/tree/main/

### Prerequisites

Create a series of radio inputs
```html
<label><input type="radio" name="group" value="a" />A</label>
<label><input type="radio" name="group" value="d" />D</label>
<label><input type="radio" name="group" value="e" />E</label>
```

Create a test that confirms radio behavior using the userEvent keyboard function
```
describe('Radio implementation', () => {
    test.each([
      {
        name: "per [ArrowDown]",
        focus: 'A',
        key: "[ArrowDown]",
        expectedTarget: 'D',
      },
      {
        name: "per [ArrowLeft]",
        focus: 'D',
        key: "[ArrowLeft]",
        expectedTarget: 'A',
      },
      {
        name: "per [ArrowRight]",
        focus: 'A',
        key: "[ArrowRight]",
        expectedTarget: 'D',
      },
      {
        name: "per [ArrowUp]",
        focus: 'D',
        key: "[ArrowUp]",
        expectedTarget: 'A',
      },
      {
        name: "forward around the corner",
        focus: 'E',
        key: "[ArrowRight]",
        expectedTarget: 'A',
      },
      {
        name: "backward around the corner",
        focus: 'A',
        key: "[ArrowUp]",
        expectedTarget: 'E',
      },
    ])('$name', async ({ focus, key, expectedTarget }) => {
      const user = userEvent.setup()
        render(<App />)
        screen.getByLabelText(focus).focus()
        await user.keyboard(key)
        expect(screen.getByLabelText(expectedTarget)).toHaveFocus()
    })
})
```

### Arrow Right
If A is selected and ArrowRight is pressed, D should be selected
If E is selected and ArrowRight is pressed, A should be selected
### Arrow Down
If A is selected and ArrowDown is pressed, D should be selected
If E is selected and ArrowDown is pressed, A should be selected
### Arrow Left
If D is selected and ArrowLeft is pressed, A should be selected
If A is selected and ArrowLeft is pressed, E should be selected
### Arrow Up
If D is selected and ArrowUp is pressed, A should be selected
If A is selected and ArrowUp is pressed, E should be selected

### Arrow Right
If A is selected and ArrowRight is pressed, E is selected
If E is selected and ArrowRight is pressed, D is selected
### Arrow Down
[...]

PR #1199, opened 2024-02-27
Fix radio events

joshferrell:
**What**:
There is currently an issue with the behavior of radio inputs and the user-event library. Because the `walkRadio` method does not break out of the loop when iterating through options, it can sometimes skip over the correct option.

There's also an issue where the direction that the arrow up and arrow down keys were swapped. Arrow down should increment and arrow up should decrement.

Fixes issue: https://github.com/testing-library/user-event/issues/1198

**Why**:

There was an issue with the robustness of the tests where going through the radio options, only A and D were possible. So the tests were just flipping between the only two available radio options. Adding an additional radio input breaks the tests. 

**How**:

I've updated the tests to include an additional radio input to ensure that behavior is correct. I've also added an additional tests to ensure that radio inputs behave correctly for aria-disabled as they should still be focusable. This matches the browser behavior.

**Checklist**:

- [N/A] Documentation 
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/event/behavior/keydown.ts
+++ b/src/event/behavior/keydown.ts
@@ -28,7 +28,7 @@ const keydownBehavior: {
   ArrowDown: (event, target, instance) => {
     /* istanbul ignore else */
     if (isElementType(target, 'input', {type: 'radio'} as const)) {
-      return () => walkRadio(instance, target, -1)
+      return () => walkRadio(instance, target, 1)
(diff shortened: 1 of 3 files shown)


jemega1545 (approved):

ph-fritsche:
resolved by #1049

PR #1208, opened 2024-04-17
Remove label-click-implies-input-click as this behavior is handled by DOM libs

Lanny:
There's this common UA behavior where clicking on a label which controls an input behaves the same way as a click on the input. As far as I can tell this isn't part of any standard, but it's pretty much universal among browsers in practice. There is some code here in user-event to simulate this behavior, however [jsdom](https://github.com/jsdom/jsdom/blob/main/lib/jsdom/living/nodes/HTMLLabelElement-impl.js#L87) and [happy-dom](https://github.com/capricorn86/happy-dom/blob/c29f36c5644eacd0546b5c302ecd6bd7feacf50f/packages/happy-dom/src/nodes/html-label-element/HTMLLabelElement.ts) both implement this.

In the case of happy-dom this causes a bug producing double-clicks on elements when their controlling label is clicked: https://github.com/capricorn86/happy-dom/issues/1410 . This doesn't affect jsdom because the way it handles `preventDefault` is different, but that's purely an implementation detail that we shouldn't be relying on here.

This MR removes the logic to attempt to click controlled inputs when their controlling labels are clicked and leaves it to the DOM lib. In so doing, to prevent interfering with the dom lib's handling of this behavior, this PR also removes the logic to half-way preventing default when user-event has some additional behavior to apply on top of dispatching an event (e.g. in the case of click or clipboard events). It seems to have only really been needed for the label-click scenario. If we find this is necessary, we can let the individual behavior handlers make their own determination about preventing default or not.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/event/behavior/click.ts
+++ b/src/event/behavior/click.ts
@@ -1,8 +1,8 @@
-import {cloneEvent, getWindow, isElementType, isFocusable} from '../../utils'
+import {getWindow, isElementType, isFocusable} from '../../utils'
 import {blurElement, focusElement} from '../focus'
 import {behavior} from './registry'
 
-behavior.click = (event, target, instance) => {
+behavior.click = (event, target) => {
   const context = target.closest('button,input,label,select,textarea')
   const control = context && isElementType(context, 'label') && context.control
   if (control) {
@@ -10,7 +10,6 @@ behavior.click = (event, target, instance) => {
       if (isFocusable(control)) {
         focusElement(control)
       }
-      instance.dispatchEvent(control, cloneEvent(event))
     }
   } else if (isElementType(target, 'input', {type: 'file'})) {
     return () => {
--- a/src/event/dispatchEvent.ts
+++ b/src/event/dispatchEvent.ts
@@ -40,24 +40,13 @@ export function dispatchEvent(
       )
 
   if (behaviorImplementation) {
-    event.preventDefault()
-    let defaultPrevented = false
-    Object.defineProperty(event, 'defaultPrevented', {
-      get: () => defaultPrevented,
-    })
-    Object.defineProperty(event, 'preventDefault', {
-      value: () => {
-        defaultPrevented = event.cancelable
-      },
-    })
-
     wrapEvent(() => target.dispatchEvent(event), target)
 
-    if (!defaultPrevented as boolean) {
+    if (!event.defaultPrevented) {
       behaviorImplementation()
     }
 
-    return !defaultPrevented
+    return !event.defaultPrevented
   }
 
   return wrapEvent(() => target.dispatchEvent(event), target)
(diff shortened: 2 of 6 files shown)


Lanny:
@nickmccurdy if you could take a look at this when you get a chance, I'd appreciate it!

ph-fritsche:
This is a bug in happy-dom. It works as expected in Jsdom and in Chrome.

```ts
import { setup } from '#testHelpers'
test('reproduce bug', async () => {
    const {user, xpathNode} = setup(`<label><input type="checkbox"/><span>text</span></label>`)

    let i = 0
    xpathNode('//input').addEventListener('click', () => i++)

    await user.click(xpathNode('//input'))
    expect(i).toBe(1)
    
    await user.click(xpathNode('//label'))
    expect(i).toBe(2)

    await user.click(xpathNode('//span'))
    expect(i).toBe(3)
})
```

This PR would also allow the default behavior of the DOM implementation to run before and in addition to our behavior implementation.

Lanny:
hi @ph-fritsche , please see the explanation in the PR description and my comment on the happy-dom bug report: https://github.com/capricorn86/happy-dom/issues/1410#issuecomment-2062478962

It's true that this works in jsdom, but it's because user-event is relying on an implementation detail of jsdom in a way that I think is fairly clearly undesirable. Specifically this code in user-event:

```
  if (behaviorImplementation) {
    event.preventDefault()
    let defaultPrevented = false
    Object.defineProperty(event, 'defaultPrevented', {
      get: () => defaultPrevented,
    })
    Object.defineProperty(event, 'preventDefault', {
      value: () => {
        defaultPrevented = event.cancelable
      },
    })
```

I'm not sure what the motivation for this code existing is, user-event is calling `preventDefault` but then overrides the `defaultPrevented` field. Because of how jsdom works, this prevents its built in behavior of label click -> input click delegation in jsdom. In happy-dom doing this does not end up preventing the same behavior because happy-dom later refers back to the `defaultPrevented` field. There's really no reason to do this, the only apparent motivation is to prevent the dom library from doing it itself, which is probably can do faster (or at least the option for optimization is out there)

ph-fritsche:
It works in Jsdom which is what most of our users use for testing. It works in browsers which is our source of truth.
We won't break this library in those environments because Happy-dom decides to read `Event.defaultPrevented` instead of using an internal state as the user agent is supposed to.

```js
element.addEventListener('click', e => e.preventDefault())
element.addEventListener('click', e => Object.defineProperty(e, 'defaultPrevented', {value: false}))

element.click() // This should not do anything. If it does, this is a bug in the DOM implementation
```

This library is about simulating events on the DOM like they happen in the browser when a user interacts with the UI layer which is not programmatically accessible.
When we implement a default behavior (that's what the behavior plugins do), we replace the behavior of the DOM implementation which is incomplete or different from the behavior of browsers on a true UI event.
The code prevents default behavior of the DOM implementation (at least those without a bug) and makes the event handlers receive an event that still has `defaultPrevented: false` until one of them invokes `.preventDefault()`.

PR #1224, opened 2024-06-11
Update PointerEvent to have default width / height set to 1

chuganzy:
**What**:

This PR updates `PointerEvent` to have default width / height set to `1` as per the spec.

* https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/width
* https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/height

**Why**:

`new PointerEvent('xxx')` on most browser implementations should set them `1` by default.

Also some libraries check if a pointer event is _virtual_ or not by looking at the geometry values, while we expect this library to work exactly as the actual interactions.

* ex: [Floating UI](https://github.com/floating-ui/floating-ui/blob/3656e71dbd59078debc393696a50b4b5aee57f65/packages/react/src/utils.ts#L90)

**How**:

Just use `1` as default values on them.

**Checklist**:

- [x] Documentation
  - Not needed?
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/event/createEvent.ts
+++ b/src/event/createEvent.ts
@@ -262,8 +262,8 @@ function initPointerEvent(
 ) {
   assignProps(event, {
     pointerId: sanitizeNumber(pointerId),
-    width: sanitizeNumber(width),
-    height: sanitizeNumber(height),
+    width: width ?? 1,
+    height: height ?? 1,
     pressure: sanitizeNumber(pressure),
     tangentialPressure: sanitizeNumber(tangentialPressure),
     tiltX: sanitizeNumber(tiltX),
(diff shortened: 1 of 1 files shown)


src/event/createEvent.ts
```diff
@@ -135,7 +135,7 @@ function initUIEvent(event: UIEvent, {view, detail}: UIEventInit) {
   })
 }
 
-function initUIEventModififiers(
+function initUIEventModifiers(
```
chuganzy, line 138: Updated as I thought this is typo?

devongovett:
Any chance this PR could be merged? This affects pointer event handling by React Aria, which checks width / height to determine if events are coming from virtual clicks (e.g. screen readers).

ph-fritsche (approved):

chuganzy:
Thank you for the release!

Issue #577, opened 2021-03-08
Performance issues with `userEvent.type()`


Etheryte:
- `@testing-library/user-event` version: 12.6.0
- Testing Framework and version: `@testing-library/react` version 11.2.1
- DOM Environment: `jsdom` version 16.4.0

Relevant code or config:

```javascript
describe("demo", () => {
  const text = "2001:db8:ac10:fd01::20";

  test("time type", () => {
    render(<input data-testid="input" type="text" />);
    const input = screen.getByTestId('input');
    const before = new Date();
    userEvent.type(input, text);
    console.log(`type took ${new Date() - before}ms`);
    expect(input.value).toEqual(text);
  });

  test("time delayed type", async () => {
    render(<input data-testid="input" type="text" />);
    const input = screen.getByTestId('input');
    const before = new Date();
    await userEvent.type(input, text, {
      delay: Number.MIN_VALUE,
    });
    console.log(`delayed type took ${new Date() - before}ms`);
    expect(input.value).toEqual(text);
  });

  test("time paste", () => {
    render(<input data-testid="input" type="text" />);
    const input = screen.getByTestId('input');
    const before = new Date();
    userEvent.paste(input, text);
    console.log(`paste took ${new Date() - before}ms`);
    expect(input.value).toEqual(text);
  });
});
```

Problem description:

Running the above test suite with `yarn test --no-cache` logs the following values on my machine:  

```
type took 29ms
delayed type took 63ms
paste took 1ms
```
The results are fairly consistent between reruns (plus-minus a few ms).  

The performance difference is starker with longer strings: replacing the input text with `const text = "2001:db8:ac10:fd01::20".repeat(100);`, the results are as follows:  

```
type took 1380ms
delayed type took 5940ms
paste took 1ms
```

I understand that some of this difference stems from the letter-by-letter nature of `type()` which is crucial for certain test scenarios. For the time being however, `type()` is too slow for many of the tests we run and we've had to opt for `paste()` instead.  

I haven't profiled `type()`'s internals so I don't know what the source of the performance penalty is.

ph-fritsche:
I don't think there is much to gain optimizing the test utility code for performance.
After all we're talking about miliseconds in a test environment that can run multiple tests parallel.
The approach here is to trade manual labor (analyzing code) for automated labor (running test containers) by running these utilities in the first place.

If you put in the work and optimize some part of the code, a PR will be welcome.
But keep in mind we'll err on the side of maintainability.

[...]

PR #1233, opened 2024-09-17
Optimize typing a single character

Janpot:
**What**:

This improves performance of typing a single character by ~2x

**Why**:

Users complain about performance https://github.com/testing-library/user-event/issues/577. Also related to [this thread](https://github.com/mui/material-ui/pull/43757#discussion_r1763133261)

**How**:

The desired behavior is to (I suppose) yield to the next macrotask in between keyboard actions. We can avoid the initial `wait`. Some quick benchmarking suggested those `wait` calls make up the bulk of the runtime of this function. When there is only 1 action to run, this brings the `wait` calls from 2 => 1

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged
      

I don't believe documentation/tests would be required here.

### DIFF
--- a/src/utils/misc/wait.ts
+++ b/src/utils/misc/wait.ts
@@ -1,11 +1,11 @@
 import {type Instance} from '../../setup'
 
-export function wait(config: Instance['config']) {
+export async function wait(config: Instance['config']): Promise<void> {
   const delay = config.delay
   if (typeof delay !== 'number') {
     return
   }
-  return Promise.all([
+  await Promise.all([
     new Promise<void>(resolve => globalThis.setTimeout(() => resolve(), delay)),
     config.advanceTimers(delay),
   ])
--- a/src/keyboard/index.ts
+++ b/src/keyboard/index.ts
@@ -14,7 +14,9 @@ export async function keyboard(this: Instance, text: string): Promise<void> {
   const actions: KeyboardAction[] = parseKeyDef(this.config.keyboardMap, text)
 
   for (let i = 0; i < actions.length; i++) {
-    await wait(this.config)
+    if (i > 0) {
+      await wait(this.config)
+    }
 
     await keyboardAction(this, actions[i])
   }


ph-fritsche:
The initial `wait` ensures that micro tasks added during (re)render are done before we dispatch events or manipulate the DOM.

Could you verify that resolving the promise on a new event loop step carries that performance hit?
Or is this just caused by other tasks that run in between? Wouldn't these other tasks run afterwards if we remove the `setTimeout` call and result in about the same overall execution time?

src/utils/misc/wait.ts
```diff
@@ -1,11 +1,11 @@
 import {type Instance} from '../../setup'
 
-export function wait(config: Instance['config']) {
+export async function wait(config: Instance['config']): Promise<void> {
   const delay = config.delay
   if (typeof delay !== 'number') {
     return
   }
-  return Promise.all([
+  await Promise.all([
```
ph-fritsche, line 8: This makes the function return `Promise<undefined>` instead of `undefined` for `delay: null` and results in `await wait()` to add a new micro task.
Are we sure this has no effect?

Janpot: Doesn't `await` always create a new microtask regardless? Doesn't matter what the right-hand-side is, right?

ph-fritsche: I'm not sure. I guess it depends on implementation.

Typescript reports `'await' has no effect on the type of this expression. ts(80007)` if you try `await undefined`.

The following results in `sync` in `node@10` and `async` in `node@12`:
```js
async function waitAsync() {
    return undefined
}

function waitSync() {
    return undefined
}

(async function () {
    await Promise.race([
        (async () => {
            await waitAsync()
            return 'async'
        })(),
        (async () => {
            await waitSync()
            return 'sync'
        })(),
    ]).then(console.log)
})()
```

If we're not sure that this has no effect on any platform that is still supported, I wouldn't touch it if I don't have to.

snowystinger:
I'm going to close this one, it looks like it's risky and there are some open questions still.

PR #1235, opened 2024-09-24
allow keyboard events on shadow root hosts

### LINKED ISSUE #1153: User Events: keyboard does not target focused element with shadow dom
### Reproduction example

https://codesandbox.io/s/trusting-panka-gvg259

### Prerequisites

1. Render an element with an open shadow dom 
2. Focus said element
3. Use `keyboard` to trigger keys

### Expected behavior

The focused element should register the keyboard events

### Actual behavior

The element does not register the keyboard events

### User-event version

^14.4.3

### Environment

Testing Library framework: `@testing-library/dom@9.3.1` 
JS framework: `lit@2.7.5`
Test environment: `jest@29.5.0`
DOM implementation: `jsdom@20.0.3`

### Additional context

I identified that the issue is that when `getActiveElement` gets an element with shadowDom, it gets the active element of the shadowDom. But, the element with shadow dom could be the active one, even if no shadow children are active (`tabindex`). Then it returns null, instead of returning the element.

mattclough1:
**What**:
Resolves #1153 

**Why**:
The current code only allows keyboard entry on elements with a ShadowRoot if the ShadowRoot has a focusable element

**How**:
The `getActiveElement` function previously checked for the existence of a `shadowRoot` on the currently active element, since active elements in a shadow DOM make the host element the active element in the light DOM, but this logic ignored the fact that host elements can also be active elements on their own.

**Checklist**:

- [N/A] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/utils/focus/getActiveElement.ts
+++ b/src/utils/focus/getActiveElement.ts
@@ -6,18 +6,20 @@ export function getActiveElement(
   const activeElement = document.activeElement
 
   if (activeElement?.shadowRoot) {
-    return getActiveElement(activeElement.shadowRoot)
-  } else {
-    // Browser does not yield disabled elements as document.activeElement - jsdom does
-    if (isDisabled(activeElement)) {
-      return document.ownerDocument
-        // TODO: verify behavior in ShadowRoot
-        ? /* istanbul ignore next */ document.ownerDocument.body
-        : document.body
+    const activeShadowRootElement = getActiveElement(activeElement.shadowRoot)
+    if (activeShadowRootElement) {
+      return activeShadowRootElement
     }
-
-    return activeElement
   }
+  // Browser does not yield disabled elements as document.activeElement - jsdom does
+  if (isDisabled(activeElement)) {
+    return document.ownerDocument
+      ? // TODO: verify behavior in ShadowRoot
+        /* istanbul ignore next */ document.ownerDocument.body
+      : document.body
+  }
+
+  return activeElement
 }
 
 export function getActiveElementOrBody(document: Document): Element {
--- a/src/utils/focus/getActiveElement.ts
+++ b/src/utils/focus/getActiveElement.ts
@@ -14,8 +14,8 @@ export function getActiveElement(
   // Browser does not yield disabled elements as document.activeElement - jsdom does
   if (isDisabled(activeElement)) {
     return document.ownerDocument
-      ? // TODO: verify behavior in ShadowRoot
-        /* istanbul ignore next */ document.ownerDocument.body
+      // TODO: verify behavior in ShadowRoot
+      ? /* istanbul ignore next */ document.ownerDocument.body
       : document.body
   }
 
--- a/src/utils/focus/getActiveElement.ts
+++ b/src/utils/focus/getActiveElement.ts
@@ -6,9 +6,9 @@ export function getActiveElement(
   const activeElement = document.activeElement
 
   if (activeElement?.shadowRoot) {
-    const activeShadowRootElement = getActiveElement(activeElement.shadowRoot)
-    if (activeShadowRootElement) {
-      return activeShadowRootElement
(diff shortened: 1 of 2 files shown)


tests/keyboard/index.ts
```diff
@@ -163,3 +163,30 @@ test('disabling activeElement moves action to HTMLBodyElement', async () => {
...
+  class CustomElement extends HTMLElement {
+    constructor() {
+      super()
+
+      this.attachShadow({mode: 'open'})
+      this.addEventListener('keypress', keypressHandler)
+    }
+  }
+
+  customElements.define('custom-element', CustomElement)
```
ph-fritsche, line 178: The element definition should be moved out of the test function to the bottom of the test file because the test should have no side-effect and the custom element will be defined for any tests that might be added below.

Naming it e.g. `focusable-shadow-root` might help someone working on this file and I think inlining the class definition into the `define()` might help if someone decides to refactor the file later.

tests/keyboard/index.ts
```diff
@@ -163,3 +163,30 @@ test('disabling activeElement moves action to HTMLBodyElement', async () => {
...
+  )
+
+  // Tab to body
+  await user.tab()
+  // Tab to custom element
+  await user.tab()
+  expect(element).toHaveFocus()
+
+  await user.keyboard('[Space]')
+  expect(keypressHandler).toHaveBeenCalled()
```
ph-fritsche, line 191: The element (or the first focusable element child) is automatically focused in `setup` (unless `focus: false` option is set). We can simplify the test case.

```suggestion
  const {user, eventWasFired} = setup(
    '<custom-element tabindex="0"></custom-element>',
  )

  await user.keyboard('[Space]')
  expect(eventWasFired('keypress')).toBe(true)
```

src/utils/focus/getActiveElement.ts
```diff
@@ -6,18 +6,20 @@ export function getActiveElement(
...
-  } else {
-    // Browser does not yield disabled elements as document.activeElement - jsdom does
-    if (isDisabled(activeElement)) {
-      return document.ownerDocument
-        ? // TODO: verify behavior in ShadowRoot
-          /* istanbul ignore next */ document.ownerDocument.body
-        : document.body
+    const activeShadowRootElement = getActiveElement(activeElement.shadowRoot)
+    if (activeShadowRootElement) {
+      return activeShadowRootElement
```
ph-fritsche, line 11:
```suggestion
    const activeElementInShadowTree = getActiveElement(activeElement.shadowRoot)
    if (activeElementInShadowTree) {
      return activeElementInShadowTree
```

ph-fritsche (changes requested):
Thanks for contributing to this library. :blue_heart:

tests/keyboard/index.ts
mattclough1: The custom element isn't actually focusable by itself, but we add the `tabindex` attribute to it in the test to make it focusable. I'm going to (conditionally for best practice) add a tabindex in the constructor so that this custom element serves this single purpose. Does that make sense?

mattclough1: Like this:

```suggestion
  customElements.define(
    'focusable-custom-element',
    class FocusableCustomElement extends HTMLElement {
      constructor() {
        super()
        this.attachShadow({mode: 'open'})
      }
    
      connectedCallback() {
        if (!this.hasAttribute('tabindex')) {
          this.setAttribute('tabindex', '0')
        }
      }
    }
  )
```

ph-fritsche: I think either would be okay:
`<custom-element tabindex="0"></custom-element>` or `<focusable-custom-element></focusable-custom-element>`.

Just move it out of the test.

ph-fritsche (approved):

Artur-:
Ran into this same issue today. Can we get a release with it?

coder-se:
Same here, just ran into this issue. Fantastic that it's already solved. Would love a release with it.

PR #1238, opened 2024-11-06
fix(click): skip explicit re-dispatch for face

### LINKED ISSUE #1237: click leads to endless loop for custom elements in label
### Reproduction example

self contained html provided in prerequisites

### Prerequisites

```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <title></title>
  </head>
  <body>
    <script type="module">
      import userEvent from "https://esm.sh/@testing-library/user-event@14.5.2?bundle";

      class MyFoo extends HTMLElement {
        static formAssociated = true;
      }
      customElements.define("my-foo", MyFoo);

      const label = document.createElement("label");
      const myFoo = document.createElement("my-foo");
      label.append(myFoo);

      document.body.append(label);

      user.click(document.querySelector("label"));
    </script>
  </body>
</html>
```

loading this throws the following error:
```sh
isElementType.js:5 Uncaught (in promise) RangeError: Maximum call stack size exceeded
    at Function.isArray (<anonymous>)
    at c (isElementType.js:5:17)
    at T.click (click.js:17:32)
    at Object.ne (dispatchEvent.js:25:160)
    at c.type (click.js:23:22)
    at Object.ne (dispatchEvent.js:39:13)
    at c.type (click.js:23:22)
    at Object.ne (dispatchEvent.js:39:13)
    at c.type (click.js:23:22)
    at Object.ne (dispatchEvent.js:39:13)
```

### Expected behavior

Should not lead to endless loog

### Actual behavior

leads to endless loop

### User-event version

14.5.2

### Environment

nope.

### Additional context

I tried debugging it a bit and found that the PR #850 in which the closest label of a click target is selected and that is supposed to break the loop - I'm not sure how?

hesxenon:
**What**:

closes #1237 

**Why**:

Re-dispatching a click event on the control of a label is already skipped for labeled controls that are builtin. This change also skips form associated custom elements (FACEs) since they are always classified as labelable elements and are thus handled by the user agent already.

**How**:

Check whether the `formAssociated` static property exists by reading it from the targets constructor.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,10 +3,8 @@ Thanks for your interest in the project. Bugs filed and PRs submitted are apprec
 
 Please make sure that you are familiar with and follow the Code of Conduct for
 this project (found in the CODE_OF_CONDUCT.md file).
-
 Also, please make sure you're familiar with and follow the instructions in the
 contributing guidelines (found in the CONTRIBUTING.md file).
-
 If you're new to contributing to open source projects, you might find this free
 video course helpful: http://kcd.im/pull-request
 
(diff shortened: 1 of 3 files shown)


hesxenon:
there's a bit of a problem with the test setup... jsdom does not yet support FACEs as far as I can see: https://github.com/jsdom/jsdom/issues/3444

So while rudimentary CE tests are possible it seems that e.g. the HTMLLabel implementation does not find the nested FACE :/

Any tips on how to progress here?

hesxenon:
@ph-fritsche would you mind taking a look at this? It's been some time since I fully grokked the behaviour in question and upon updating jsdom to the latest version (in which appropriate tests for this change would even be possible) a few more tests than I was hoping for fail.

ph-fritsche:
@hesxenon I've rebased the commits onto current main and applied some changes on top. What do you think?

ph-fritsche (approved):

PR #1252, opened 2025-01-15
feat: dispatch `FocusEvent` in hidden documents

ph-fritsche:
**What**:

Dispatch `focus`, `focusin`, `blur` and `focusout` events when `document.visibilityState` is `hidden`.

**Why**:

Browsers don't dispatch focus events on hidden windows.
Our simulation of the UI layer relies on these events and userEvent should act like a user does on a visible window.

**How**:

Patch `HTMLElement.focus()` and `HTMLElement.blur()` and dispatch the events if the document is hidden.

**Checklist**:

- [x] Tests
- [x] Ready to be merged


donaldpipowitch:
It looks like this change broke some integrations. See https://github.com/testing-library/user-event/discussions/1258. For me it broke react-aria here:

https://github.com/adobe/react-spectrum/blob/4b8b33a022a17c7df2d6ee6d23b6208b503973fe/packages/%40react-aria/interactions/src/useFocusVisible.ts#L138

> TypeError: Cannot set property focus of [object HTMLElement] which has only a getter

PR #1270, opened 2025-02-03
feat: export callbacks for hooks

ph-fritsche:
### What

Export `@experimental` functions `prepare`, `reset` and `detach` to improve support for different test environments.

```js
import userEvent, {prepare, reset, detach } from '@testing-library/user-event'
import { beforeAll, afterEach, afterAll, test } from 'some-test-runner'
import { DomImplementation } from 'some-dom-implementation'

const window = new DomImplementation()

beforeAll(() => prepare(window))
afterEach(() => reset(window))
afterAll(() => detach(window))

test('something', async () => {
  const user = userEvent.setup({document: window.document})
  // ...
```

### Why

UserEvent should work in various environments. Exposing these functions allows our users to integrate UserEvent in environments that don't use global hooks or global document.

### How

Includes a fix to `restoreFocus`.

Includes a check if `globalThis.window` exists.

### Checklist
- [ ] Ready to be merged

### DIFF
--- a/src/document/patchFocus.ts
+++ b/src/document/patchFocus.ts
@@ -5,7 +5,7 @@ const patched = Symbol('patched focus/blur methods')
 
 declare global {
   interface HTMLElement {
-    readonly [patched]?: Pick<HTMLElement, 'focus' | 'blur'>
+    [patched]?: Pick<HTMLElement, 'focus' | 'blur'>
   }
 }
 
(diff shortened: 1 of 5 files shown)


brynshanahan:
How about calling `userEvent.setup()` and `userEvent.${directApis}` calls the prepare functions automatically then, you would only need to `afterEach(() => detach())`

ph-fritsche:
@brynshanahan `prepare` is already called internally by `setup`. Exposing the function allows to call it earlier to resolve conflicts with other frameworks/tools.
```
This should work:
domImplementation.createDocument()
--> userEvent.prepare()
--> someLibrary.setup()/userEvent.setup()
  --> someComponent.render()
  --> test interaction
  --> someLibrary.teardown()/userEvent.reset() (reverse order)
[...]
--> userEvent.detach()
--> execute any code on the document as if userEvent had never been used
```

PR #1272, opened 2025-02-06
change value of number input when pressing arrow up or down

### LINKED ISSUE #1066: ArrowDown / ArrowUp events in "number" input fields do not trigger onChange
### Reproduction example

https://codesandbox.io/s/eloquent-tristan-ixhpuv?file=/src/number.test.js

### Prerequisites

1. Create a `Number` component with an `onChange` handler that calls back to the container. 
2. Create a test where `Number` component is rendered with an `onChange` func created with `jest.fn()`. Use `userEvent` to click into the field and then simulate a down-arrow keyboard press.

For example:

```
test('the value can be changed with arrow up & down', async () => {
    const initialValue = '7';
    const minValue = '5';
    const onChange = jest.fn();

    render(<NumberField label={LABEL} value={initialValue} onChange={onChange} min={minValue} />);

    const input = screen.getByRole('spinbutton');

    await userEvent.click(input);
    await userEvent.keyboard('{arrowdown}');
    expect(onChange).toBeCalledWith('6');
})
```

### Expected behavior

Expected behaviour is for the test to pass. Pressing the down arrow in a browser would cause the number value to reduce by 1 (assuming `step=1`). 

### Actual behavior

Upon running the tests, the test fails with the following result:
```
expect(jest.fn()).toBeCalledWith(...expected)

    Expected: "6"

    Number of calls: 0

    > 64 |              expect(onChange).toBeCalledWith('6');
         |                               ^
      65 |      });
      66 | });
```

### User-event version

14.4.3

### Environment

Testing Library framework: @testing-library/react@13.4.0
JS framework: `react@18.2.0`
Test environment: `jest@27.5.1`
DOM implementation: `jsdom@16.7.0` 

### Additional context

_No response_

Haberkamp:
### What

With this change you can change the value of a number input by pressing the `ArrowUp` and `ArrowDown` key, like so:

```ts
await userEvent.type(screen.getByRole('spinbutton'), "{ArrowUp}");
```

### Why

In browsers users are able to increment or decrement the value of a number input when pressing the up or down arrow key.

Adding this to `userEvent` closes the gap between this package and the browsers.

### How

Checks if the input field has a `min` or `max` value defined, yes then the value cannot exceed those boundaries. If not the user can increment or decrement the value indefinitely.

If the input field has a `step` property then it increments or decrements in those steps.

### Checklist

- [N/A] Documentation
- [x] Tests
- [x] Ready to be merged 

Resolves #1066

### DIFF
--- a/src/event/input.ts
+++ b/src/event/input.ts
@@ -11,6 +11,8 @@ import {
   EditableInputOrTextarea,
   getMaxLength,
   getNextCursorPosition,
+  isDisabled,
+  isEditable,
   isElementType,
   isValidDateOrTimeValue,
   supportsMaxLength,
@@ -217,6 +219,35 @@ function calculateNewValue(
     if (builtValue !== '' && isValidDateOrTimeValue(node, builtValue)) {
       newValue = builtValue
       newOffset = builtValue.length
+    }
+  }
+
+  if (
+    isElementType(node, 'input', {type: 'number'} as const) &&
+    inputType === 'changeNumberInput' &&
+    !isDisabled(node) &&
+    !node.readOnly
+  ) {
+    const step = node.step ? Number(node.step) : 1
+
+    const reachedMax = value === node.max
+    if (inputData === 'ArrowUp' && !reachedMax) {
+      const exceedsMax = Number(value) + step > Number(node.max)
+      if (exceedsMax && !!node.max) {
+        newValue = node.max
+      } else {
+        newValue = (Number(value) + step).toString()
+      }
+    }
+
+    const reachedMin = value === node.min
+    if (inputData === 'ArrowDown' && !reachedMin) {
+      const exceedsMin = Number(value) - step < Number(node.min)
+      if (exceedsMin && !!node.min) {
+        newValue = node.min
+      } else {
+        newValue = (Number(value) - step).toString()
+      }
     }
   }
 
--- a/src/event/behavior/keydown.ts
+++ b/src/event/behavior/keydown.ts
@@ -28,6 +28,10 @@ const keydownBehavior: {
     if (isElementType(target, 'input', {type: 'radio'} as const)) {
       return () => walkRadio(instance, target, 1)
     }
+
+    if (isElementType(target, 'input', {type: 'number'} as const)) {
+      return () => input(instance, target, 'ArrowDown', 'changeNumberInput')
+    }
   },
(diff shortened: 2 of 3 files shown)


src/event/input.ts
```diff
@@ -220,6 +222,35 @@ function calculateNewValue(
...
+    isElementType(node, 'input', {type: 'number'} as const) &&
+    inputType === 'changeNumberInput' &&
+    !isDisabled(node) &&
+    !node.readOnly
+  ) {
+    const step = node.step ? Number(node.step) : 1
+
+    const reachedMax = value === node.max
+    if (inputData === 'ArrowUp' && !reachedMax) {
+      const exceedsMax = Number(value) + step > Number(node.max)
```
greypants, line 235: What if `value` is empty? `Number(value)` will be `NaN`

greypants: in that case I think value should be treated as "zero" (seems like that's what browsers do)

src/event/input.ts
```diff
@@ -220,6 +222,35 @@ function calculateNewValue(
...
+      if (exceedsMax && !!node.max) {
+        newValue = node.max
+      } else {
+        newValue = (Number(value) + step).toString()
+      }
+    }
+
+    const reachedMin = value === node.min
+    if (inputData === 'ArrowDown' && !reachedMin) {
+      const exceedsMin = Number(value) - step < Number(node.min)
```
greypants, line 245: same here... need to handle an empty value and bump to zero or the min value? or maybe max?

greypants (changes requested):
Thanks for getting this started!  I ran into the same thing today. 

I think you're missing handling empty values. Add that, and this would be great to have added!

snowystinger (commented):
Found a failing test
```
test("increments number input's value by the defined steps when pressing the arrow up key with offset", () => {
  const {element} = render<HTMLInputElement>(
    `<input min="5" type="number" step="10"/>`,
  )

  const instance = setupInstance()

  instance.dispatchUIEvent(element, 'keydown', {key: 'ArrowUp'})

  expect(element).toHaveValue(5)

  instance.dispatchUIEvent(element, 'keydown', {key: 'ArrowUp'})

  expect(element).toHaveValue(15)
})
```

here's the algorithm we need to implement https://html.spec.whatwg.org/#dom-input-stepup

We should also have a test that shows we don't change the value if preventDefault is called

PR #1283, opened 2025-04-13
Fix dispatch wrapping for internal events

mgmolisani:
### What

There a few places in the library where events are dispatched in response to already triggered events. These seemingly fix missing/differing functionality in JSDOM for [events relating to focus](https://github.com/search?q=repo%3Atesting-library%2Fuser-event%20dispatchDOMEvent&type=code). This includes handling focus/blur when `visibilityState` is `hidden` and triggering `change` events on `blur`. However, these are also wrapped in the runtime's configured `eventWrapper`. In React, this is the `act` wrapper. Due to how React flushes the update queue in `act` environments, this can prematurely tear down the queue making future updates trigger...

>Warning: An update to %s inside a test was not wrapped in act(...).

Users cannot avoid these error logs (and in our case, we elevate them and fail our tests as if they were real errors). This can be completely avoided though as when we trigger events for `focus` and `blur`, we should already be in a wrapped scope. To further that point, imagine in user space we had a component that re-dispatches a custom `change` event:

```tsx
function Component(): ReactElement {
  const handleChange = (event: ChangeEvent<HTMLInputElement>): void => {
    const customEvent = new CustomEvent("custom", {
      detail: event.target.value,
    });

    event.target.dispatchEvent(customEvent);
  };

  return <input data-testid="input" onChange={handleChange} />;
}
```

In this case, it's obvious that we cannot wrap our `dispatchEvent` call in `act` as it's part of the component yet this is totally fine because our synthetic React `change` event presumably was triggered by a wrapped call like `userEvent.type(screen.getByTestId("input"), "some text")`. The same should be true here where the focus/blur events are already wrapped and the wrapping around the patched event is unnecessary (and, in the case of React, problematic).

### Why

In React, nested synchronous `act` calls can prematurely teardown the `act` queue creating console warnings. In my team's case, we elevate these to real errors in our tests making them more visible. These warnings can be avoided as described in the [What](#What) section. It's also possible React could warn about `act` wrapped state updates on the queue and might even handle the nesting in async cases already. Nesting seems like it should be avoided though as users can't really create the same behavior in a real app. It's still not totally clear to me why this is allowed or if it should be handled already by React and is a bug. But in any case, I don't think we have to wrap the internal calls and can solve here.

### How

I removed the call to `wrapEvent` in `dispatchDOMEvent` which is currently only used for these focus/blur event. I am not sure this was the original intention for the split dispatch compared to `dispatchEvent` in the same file, but it seems very fitting as fixing here fixes all the current cases where wrapping should not be added and doesn't touch the places wrapping is required. I also passed through the boolean return value. It's not used anywhere but seemed like it fit the intention of the function to just be a simple wrapper around the native `dispatchEvent` API.

### Checklist

- [ ] Documentation - Internal, no updates required.
- [x] Tests - None needed to be updated (which is a good sign).
- [x] Ready to be merged  - No one should be relying on this behavior as is.

https://github.com/testing-library/user-event/pull/1275 - Another user reported this issue while also introducing a new feature. I believe the solution in this PR is more universal and decouples from the feature request to hopefully move this along more quickly.


mgmolisani:
Checking in. Do PRs ever get reviewed? Been open since April.

src/event/dispatchEvent.ts
```diff
@@ -63,11 +63,15 @@ export function dispatchEvent(
...
+ */
 export function dispatchDOMEvent<K extends EventType>(
   target: Element,
   type: K,
   init?: EventTypeInit<K>,
-) {
-  const event = createEvent(type, target, init)
-  wrapEvent(() => target.dispatchEvent(event), target)
+): boolean {
+  return target.dispatchEvent(createEvent(type, target, init))
```
snowystinger, line 76: So the thing I'm worried about here is that there may be other reasons to wrap events beyond React adding `act`. If that's the case, then suddenly this event is no longer part of that.

snowystinger: angulars testing library for instance does
https://github.com/testing-library/angular-testing-library/blob/e6f4e26d8751a09056d0c610aa5be522cc73edfa/projects/testing-library/src/lib/testing-library.ts#L94

preact is the same as react, and vue and svelte appear not to do anything

and none of this includes what users may do

snowystinger: I think we may need to make this change up in https://github.com/testing-library/react-testing-library and in the preact one as well

snowystinger: Possibly related
https://github.com/testing-library/react-testing-library/issues/1231

snowystinger:
fixed in react testing library
https://github.com/testing-library/react-testing-library/pull/1468

PR #1295, opened 2025-08-08
Make sure the default document is the correct instance

Janpot:
### What

I noticed in a setup where a jsdom instance was recreated for each test and globals were reassigned, `userEvent` used an outdated version of document.

### Why

Avoid the need of always passing a `document` to `userEvent.setup()`

### How

Use a function to obtain default values, this makes it capture `globalThis.document` at the time `userEvent.setup()` is called.

### Checklist

- [N/A] Documentation
- [N/A] Tests: feels too challenging to test under the current setup. I've added a clarifying comment.
- [x] Ready to be merged

### DIFF
--- a/src/setup/setup.ts
+++ b/src/setup/setup.ts
@@ -19,8 +19,10 @@ import {DirectOptions} from './directApi'
 
 /**
  * Default options applied when API is called per `userEvent.anyApi()`
+ * We use a function to make sure the document is always the instance
+ * of the time the API is called.
  */
-const defaultOptionsDirect: Required<Options> = {
+const getDefaultOptionsDirect = (): Required<Options> => ({
   applyAccept: true,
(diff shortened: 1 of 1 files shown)


src/setup/setup.ts
```diff
@@ -19,8 +19,10 @@ import {DirectOptions} from './directApi'
 
 /**
  * Default options applied when API is called per `userEvent.anyApi()`
+ * We use a function to make sure the document is always the instance
+ * of the time the API is called.
  */
-const defaultOptionsDirect: Required<Options> = {
+const getDefaultOptionsDirect = (): Required<Options> => ({
```
Janpot, line 25: Could also do:
```suggestion
const getDefaultOptions = (overrides: Options): Required<Options> => ({
```

PR #1306, opened 2025-10-14
call initUIEvent with the correct view

### LINKED ISSUE #709: UIEvent.view is always null
- `@testing-library/user-event` version: 13.2.1

- Testing Framework and version: "@testing-library/react": "12.0.0",
  
- DOM Environment: "@testing-library/jest-dom": "5.14.1",
  

Relevant code or config

```javascript
import React from 'react';
import { FormInput } from '@fluentui/react-northstar';

import {render, screen} from '@testing-library/react';
import userEvent from '@testing-library/user-event';

test('type in form input', () => {
  render(
    <FormInput label="Test" placeholder="welcome" />
  );

  userEvent.type(screen.getByPlaceholderText('welcome'), 'hello');

  expect(screen.getByDisplayValue('hello')).toBeInTheDocument();
});
```

Trying to execute code above fails the test with `TypeError: Cannot read property 'document' of null`

According to [specs](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/view) the `view` property of UIEvent should point to the `window`.

This issue affects other libraries, like https://github.com/microsoft/fluentui/issues/19233.

The more generic `fireEvent` works fine though - the `view` property is defined as expected:
```js
act(() => {
    fireEvent.change(screen.getByPlaceholderText('welcome'), { target: { value: 'hello' } });
});

peterkogo:
### What

We need to pass window to the UIEvent constructor otherwise `event.view` will be undefined in the emitted event.

### Why

This should fix [this erroneously closed issue](https://github.com/testing-library/user-event/issues/709).
We rely on `event.view` being defined inside [React Flow](https://github.com/xyflow/xyflow). This would also fix issues with testing [ `d3-drag`](https://github.com/d3/d3-drag).

### How

Always pass `view: window` along with the init parameters.

### Checklist

- [N/A] Documentation
- [ ] Tests
- [x] Ready to be merged


snowystinger:
Logged a bug against Chrome https://issues.chromium.org/issues/545322457
Once that's been addressed we can move forward with this issue.

https://github.com/microsoft/fluentui/pull/20022#issuecomment-930306706 suggests that this should be fixed at the jsdom level. I'm not entirely sure yet if I agree. jsdom doesn't deal in trusted events (except for focus/blur, i'm not sure why), and only in trusted events is the `view` set according to the [spec](https://w3c.github.io/uievents/#event-type-input).

User event *would* be the level at which to set trusted, but I think that's likely a large undertaking and isn't something I wish to take on right now. But adding the view at this level does seem correct as we can think of the user event flows as trusted.

If you'd like to open an issue with jsdom, we could see what they have to say as well. I've pushed some tests for this that I was working on while getting up to speed.

PR #1324, opened 2026-08-15
feat: automatically advance fake timers in Jest and Vitest

### LINKED ISSUE #1115: userEvent.click() fails when used with vi.useFakeTimers(), all available solutions are not working
### Reproduction example

https://stackblitz.com/edit/vitejs-vite-askvcq?file=src/__tests__/App.test.tsx

### Describe the bug

I have a React component which is a timer (minimal reproduction), that starts automatically after mounted, and there's a button to RESET the timer.

In the test, I'm using `vi.useFakeTimers()` and `await vi.advanceTimersByTimeAsync(500)` to test the timer segmentally.

However, I'm not able to use `await user.click()` to click the button.

This is a known issue with fake timers, however, none of the existing solutions works.

None of these works
```ts
const user = userEvent.setup({
  advanceTimers: vi.advanceTimersByTime,
  // advanceTimers: vi.advanceTimersByTimeAsync
  // advanceTimers: vi.advanceTimersByTime.bind(vi)
  // advanceTimers: vi.advanceTimersByTimeAsync.bind(vi)
  // delay: null,
})
```

And I cannot use `vi.useRealTimers()` before clicking button, since it will break the fake timer to further test the component's reset functionality.

---

Those solutions above are all based on `Jest` since almost all resources on the internet are for `Jest`. But since I've never used `Jest`, so I'm not sure if it's working in `Jest`, and is this an issue with `@testing-library/user-event` or `Vitest`

### Expected behavior

The button is successfully clicked

### Actual behavior

It makes the test timed out

### User-event version

14.0.0

### Environment

System:
  OS: Windows 10 10.0.22000
  CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
  Memory: 9.23GB / 31.69GB
Binaries:
  Node: 16.18.0 - C:\Program Files\nodejs\node.EXE
  npm: 8.19.2 - C:\Program Files\nodejs\npm.CMD
Browsers:
  Edge: Spartan (44.22000.120.0), Chromium (112.0.1722.39)
  Internet Explorer: 11.0.22000.120
npmPackages:
  @vitejs/plugin-react: ^3.1.0 => 3.1.0
  vite: ^4.2.1 => 4.2.1
  vitest: ^0.30.1 => 0.30.1
  @testing-library/jest-dom: ^5.16.5 => 5.16.5
  @testing-library/react: ^14.0.0 => 14.0.0
  @testing-library/user-event: ^14.4.3 => 14.4.3
  jsdom: ^21.1.1 => 21.1.1
  react: ^18.2.0 => 18.2.0
  react-dom: ^18.2.0 => 18.2.0

### Additional context

Please also refer to https://github.com/vitest-dev/vitest/issues/3184

TrevorBurnham:
This PR makes `userEvent` automatically advance fake timers under both Jest and Vitest, without the need for users to configure `advanceTimers` manually.

Fixes #1115

## Approach

Jest's modern fake timers and Vitest's fake timers are both built on `@sinonjs/fake-timers`, which exposes the installed clock on each timer function it replaces. So tick that clock when it is present:

```ts
export function advanceFakeTimers(delay: number): void {
  const {clock} = globalThis.setTimeout as typeof globalThis.setTimeout & {
    clock?: FakeClock
  }

  if (typeof clock?.tick === 'function') {
    clock.tick(delay)
  }
}
```

This works however the framework is imported, and only while fake timers are installed, so real timers are never touched. It's the same signal `@testing-library/dom` uses in `jestFakeTimersAreEnabled()`.

Detection happens per `wait()` rather than once in `createConfig()`, so calling `userEvent.setup()` in a `beforeEach` before `useFakeTimers()` also works. An explicit `advanceTimers` option still takes precedence.

This does _not_ add support for auto-advancing Jest's legacy fake timers, which don't expose a clock. Modern fake timers have been the default since Jest 27 (2021): https://jestjs.io/blog/2021/05/25/jest-27

## Relevant docs

The "[Using Fake Timers](https://testing-library.com/docs/using-fake-timers)" and "[Options → advanceTimers](https://testing-library.com/docs/user-event/options#advancetimers)" docs should be updated once this is merged. Currently they only mention Jest, not Vitest, and recommend this piece of setup:

```js
userEvent.setup({advanceTimers: jest.advanceTimersByTime})
```

Happy to send a docs PR alongside this.

## Verification

I've confirmed that timers automatically advance under the latest Vitest and Jest with fake timers enabled.

Covered: click and delayed `type()` under fake timers with no options, fake timers installed after `setup()`, `toFake: ['Date']` only, explicit `advanceTimers` still honored, real timers clean.

This repo's test env installs `@sinonjs/fake-timers` the same way Jest and Vitest do, so the unit tests exercise the real mechanism instead of hand-assigned globals.

## Notes for review

- `setTimeout.clock` is undocumented `@sinonjs/fake-timers` internals. `@testing-library/dom` already relies on it, but it's a notable dependency on an implementation detail.
- `waitFor` in `@testing-library/dom` sniffs the `jest` global separately, so Vitest users under fake timers may still need the `globalThis.jest` shim for `findBy*` queries.
Making that "just work" would be a natural follow-up.
- Auto-advancing configs (`jest.useFakeTimers({advanceTimers: true})`, `vi.useFakeTimers({shouldAdvanceTime: true})`) now get both auto-advance and an explicit `tick(delay)`. At the default `delay: 0` no additional virtual time is advanced, since `tick(0)` fires the timers already due without moving the clock. With a non-zero delay, the clock moves by that delay on top of auto-advance. Sinon sets `clock.attachedInterval` only when auto-advance is on, so the tick could be skipped, but advancing by the delay the caller asked for seems more correct than deferring to real time.

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -122,6 +122,11 @@ export interface Options {
   /**
    * A function to be called internally to advance your fake timers (if applicable)
    *
+   * Fake timers installed by Vitest or Jest's modern fake timers are advanced
+   * automatically.
+   * Set this option if you use another implementation of fake timers,
+   * e.g. Jest's legacy fake timers.
+   *
    * @example jest.advanceTimersByTime
    */
   advanceTimers?: ((delay: number) => Promise<void>) | ((delay: number) => void)
--- a/src/setup/setup.ts
+++ b/src/setup/setup.ts
@@ -5,6 +5,7 @@ import {defaultKeyMap as defaultKeyboardMap} from '../keyboard/keyMap'
 import {defaultKeyMap as defaultPointerMap} from '../pointer/keyMap'
 import {Options, PointerEventsCheckLevel} from '../options'
 import {
+  advanceFakeTimers,
   ApiLevel,
   attachClipboardStubToView,
   getDocumentFromNode,
@@ -32,7 +33,7 @@ const defaultOptionsDirect: Required<Options> = {
   skipClick: false,
   skipHover: false,
   writeToClipboard: false,
-  advanceTimers: () => Promise.resolve(),
+  advanceTimers: advanceFakeTimers,
 }
 
 /**
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -30,6 +30,7 @@ export * from './misc/isElementType'
 export * from './misc/isVisible'
 export * from './misc/isDisabled'
 export * from './misc/level'
+export * from './misc/timerDetection'
 export * from './misc/wait'
 
 export * from './pointer/cssPointerEvents'
--- a/src/utils/misc/timerDetection.ts
+++ b/src/utils/misc/timerDetection.ts
@@ -1,0 +1,22 @@
+interface FakeClock {
+  tick: (ms: number) => void
+}
+
+/**
+ * Advance fake timers if the test framework installed some.
+ *
+ * Jest's modern fake timers and Vitest's fake timers are both built on
+ * `@sinonjs/fake-timers`, which exposes the installed clock on each timer
+ * function it replaces.
+ * Looking for that clock instead of a `jest`/`vi` global detects fake timers
+ * however the framework is imported, and only while they are installed.
+ */
+export function advanceFakeTimers(delay: number): void {
+  const {clock} = globalThis.setTimeout as typeof globalThis.setTimeout & {
+    clock?: FakeClock
+  }
+
+  if (typeof clock?.tick === 'function') {
+    clock.tick(delay)
+  }
+}
--- a/tests/utils/misc/timerDetection.ts
+++ b/tests/utils/misc/timerDetection.ts
@@ -1,0 +1,36 @@
+import {advanceFakeTimers} from '#src/utils'
+
+test('advance fake timers installed on the global scope', () => {
+  timers.useFakeTimers()
+  try {
+    let fired = false
+    globalThis.setTimeout(() => {
+      fired = true
+    }, 100)
+
+    advanceFakeTimers(100)
+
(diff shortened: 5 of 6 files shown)


snowystinger:
Thanks, I spent some time looking over this today.

The issue this PR is closing I think is actually addressed by https://github.com/testing-library/react-testing-library/pull/1443
Right now, Jest will also hang if you use fake timers and don't follow this part of the docs https://testing-library.com/docs/user-event/options#advancetimers

So if we're aiming for vitest and jest working the same to start, which I think is the best interpretation of that issue, then I think the first change is to update `react-testing-library`, not here, that way vitest `advanceTimers` in the config works without the `globalThis.jest` shim.

Once those two work the same, then I think we can consider this PR and a different/new Issue to add automatic detection of fake timers. At least right now, vitest just breaks until you have the right setup instead of sometimes working.


One more thing to note, this changes the API default for advanceTimers, it goes from `() => Promise.resolve()` to `() => void`. I *think* this doesn't matter because the only thing that uses it is `wait` which coerces it to a promise anyways. Though the config is available, so someone else could be relying on it.

TrevorBurnham:
You're right: https://github.com/testing-library/react-testing-library/pull/1443 fixes https://github.com/testing-library/user-event/issues/1115 insofar as it makes this setup work with Vitest fake timers:

```js
const user = userEvent.setup({advanceTimers: vi.advanceTimersByTime})
```

I think detecting and handling the most common fake timer implementations automatically is the better solution for users, but I also realize that it has a larger blast radius and might want to wait for a new major version (is there a roadmap for one?).

PR #1325, opened 2026-08-17
fix: default pointer event pointerType to empty string instead of the string "undefined"

### LINKED ISSUE #1291: `click` event has wrong `pointerType` of the string `"undefined"`
### Reproduction example

Unit test case in issue

### Prerequisites

```
import { createEvent } from '#src/event/createEvent'
import { render } from '#testHelpers'

test('it does not stringify click pointerType', () => {
    const { element } = render(`<input type="checkbox"/>`)
    const event = createEvent('click', element)
    expect(event).toHaveProperty('pointerType', undefined)
})
```

### Expected behavior

Test passes

### Actual behavior

```
 FAIL  tests/event/createEvent.ts
  ● it does not stringify click pointerType

    expect(received).toHaveProperty(path, value)

    Expected path: "pointerType"

    Expected value: undefined
    Received value: "undefined"

      5 |     const { element } = render(`<input type="checkbox"/>`)
      6 |     const event = createEvent('click', element)
    > 7 |     expect(event).toHaveProperty('pointerType', undefined)
        |                   ^
      8 | })
      9 |

      at Object.toHaveProperty (tests/event/createEvent.ts:7:19)
```

### User-event version

14.6.1

### Environment

Testing Library framework: `@testing-library/react@13.4.0`

JS framework: `react@18.2.0`

Test environment: `jest@29.5.0`
    

DOM implementation: `jsdom@26.1.0`

### Additional context

Simulating a `click` does produce an event with an invalid `pointerType` of `"undefined"` (as string).

It is expected that `pointerType` can be `undefined`, but it should never be that string.

The cause for this is this line of code:
https://github.com/testing-library/user-event/blob/63ac399e06bd8f2397a6c581915acd29235f2d38/src/event/createEvent.ts#L271

I tried adding a simple PR that accounts for `undefined` here, but due to the nature of `assignProps`, the `undefined` value will be turned into `null`, which is also not correct:
https://github.com/testing-library/user-event/blob/63ac399e06bd8f2397a6c581915acd29235f2d38/src/event/createEvent.ts#L101-L105

Updating `assignProps` lead to too many snapshot changes, so I didn't pursue this further, but hope I've provided enough information that allows provide a fix.

Thanks 🙌

tzh476:
### What

Set `pointerType` to an empty string on synthetic click events when no pointer type was provided, instead of setting it to the string `"undefined"`.

### Why

`initPointerEvent` used `String(pointerType)`. When `pointerType` is missing, that becomes the invalid token `"undefined"`. Real click events can have an unset `pointerType`; they should never use that string.

Fixes #1291

### How

Set `pointerType` to an empty string when it is defined. 

### Testing

- Added unit tests for a keyboard click  as well as for the constructor

### Checklist

- [x] Tests
- [x] Ready to be merged

Made with [Cursor](https://cursor.com)


snowystinger:
Thanks for the PR. I don't think this is quite right, you're editing the initialisation of the PointerEvent, but the spec says that it should be initialised to an empty string https://w3c.github.io/pointerevents/#pointerevent-interface so it shouldn't ever be undefined.

I checked this on Chrome as well, https://jsfiddle.net/f6nLaruj/ by focusing the button, then pressing "Enter". This results in `click` followed by an empty string `''`.

It'd be good to test the same user flow instead of directly asserting on createEvent, that's not a very helpful test as it's not something we can compare the results of against browsers. It's useful to stop a regression, but for proof for this PR, it's not accurate.

tzh476:
Updated in `32e4155`:
- Initialized `pointerType` to `''` (empty string) when not provided/undefined, conforming to the W3C PointerEventInit specification default.
- Added tests asserting `pointerType: ''` on default event initialization and on keyboard-triggered click events (`[Enter]` on button elements).

src/event/createEvent.ts
```diff
@@ -268,7 +268,7 @@ function initPointerEvent(
     tiltX: sanitizeNumber(tiltX),
     tiltY: sanitizeNumber(tiltY),
     twist: sanitizeNumber(twist),
-    pointerType: String(pointerType),
+    pointerType: pointerType === undefined ? '' : String(pointerType),
```
snowystinger:
```suggestion
    pointerType: String(pointerType ?? ''),
```

src/event/createEvent.ts
```diff
@@ -222,9 +222,9 @@ function initMouseEvent(
     pageX,
     pageY,
   }: MouseEventInit &
-    Partial<
-      Pick<MouseEvent, 'x' | 'y' | 'offsetX' | 'offsetY' | 'pageX' | 'pageY'>
-    >,
+      Partial<
+        Pick<MouseEvent, 'x' | 'y' | 'offsetX' | 'offsetY' | 'pageX' | 'pageY'>
+      >,
```
snowystinger:
```suggestion
    Partial<
      Pick<MouseEvent, 'x' | 'y' | 'offsetX' | 'offsetY' | 'pageX' | 'pageY'>
    >,
```

PR #1329, opened 2026-09-10
fix: treat programmatic null input values as empty string

### LINKED ISSUE #1290: Error: Diplayed value in UI can be `null`
### Reproduction example

null

### Prerequisites

Setting a value to `null` programmatically.

https://github.com/testing-library/user-event/blob/63ac399e06bd8f2397a6c581915acd29235f2d38/src/document/UI.ts#L24

### Expected behavior

Nothing happens

### Actual behavior

`TypeError: right hand side of 'in' should be an object, got null`
Since EcmaScript returns `object` on `typeof null` this function is not safe

https://github.com/testing-library/user-event/blob/63ac399e06bd8f2397a6c581915acd29235f2d38/src/document/UI.ts#L24

### User-event version

14.6.1

### Environment

_No response_

### Additional context

_No response_

cpruijsen:
### What

Treat `null` assigned by application code to a prepared `HTMLInputElement` / `HTMLTextAreaElement` `value` as an empty string, which is what the value IDL does in browsers.

Closes #1290

### Why

`userEvent.setup()` intercepts `value` so the library can tell UI keystrokes from programmatic sets. The interceptor called `isUIValue`, which did `typeof value === 'object' && UIValue in value`. `typeof null` is `'object'`, so `in` throws:

`TypeError: Cannot use 'in' operator to search for 'Symbol(Displayed value in UI)' in null`

Callers do this at runtime (for example `input.value = null` to reset a field). In the browser that is `""`, not an exception.

### How

- Guard `isUIValue` the same way `isUISelectionStart` already guards null (`!!value && typeof value === 'object' && … in value`).
- Coerce `null` to `''` at the start of the value interceptor, before the JSDOM number sanitizer. `Number(null) === 0`, so without that coerce a number input would jump to `0` instead of empty.

The one-line `isUIValue` null check from #1132, leaving `String(null)` / `Number(null)` in `sanitizeValue`, would store `"null"` on text inputs and `0` on number inputs. Native `value = null` is already `""` here and in browsers, so the interceptor coerces `null` to `''` rather than only skipping the `in` throw. Can drop the interceptor coerce and keep only the `isUIValue` guard if the interceptor should not do IDL coercion.

The `sourceDescriptor` lookup next to the interceptor is equivalent to the previous `||` + assertion. Nested `intercept` loses narrowing, so the assertion was required by typecheck but removed as unnecessary by `eslint --fix` whenever this file is formatted. Can restore the assertion plus a disable comment instead.

### Checklist

- [N/A] Documentation
- [x] Tests
- [x] Ready to be merged


src/document/interceptor.ts
```diff
@@ -72,8 +72,9 @@ export function prepareInterceptor<
       then,
     } = interceptorImpl.call(this, ...args)
 
-    const realFunc = ((!applyNative && objectDescriptor) ||
-      (prototypeDescriptor as PropertyDescriptor))[target] as (
+    const sourceDescriptor =
```
snowystinger, line 75: what was this change for?

snowystinger (commented):
Thanks for the PR.

Do we need to worry about all the other kinds of input? there are 22 of them https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/checkbox

You alluded to some of this in your comments, but providing links as well to the specs:

https://html.spec.whatwg.org/multipage/input.html#htmlinputelement defines
```
[[CEReactions](https://html.spec.whatwg.org/multipage/custom-elements.html#cereactions)] attribute [[LegacyNullToEmptyString](https://webidl.spec.whatwg.org/#LegacyNullToEmptyString)] DOMString [value](https://html.spec.whatwg.org/multipage/input.html#dom-input-value);
```
Where LegacyNullToEmptyString is defined https://webidl.spec.whatwg.org/#LegacyNullToEmptyString

additional information about santization of setting the value property https://html.spec.whatwg.org/multipage/input.html#value-sanitization-algorithm

I tested `$0.value = null` in https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/number and the text version across Safari, Chrome, Firefox, it appears to behave the same across all three.

cpruijsen:
Not part of the fix. The fix is only the two `null` guards: `isUIValue` accepting `null`, and `prepareValueInterceptor` coercing `null` to `''` before `sanitizeValue`, matching the `[LegacyNullToEmptyString]` IDL on `value`.

The `realFunc` rewrite is fallout from touching the file. The original reads:

```ts
const realFunc = ((!applyNative && objectDescriptor) ||
  (prototypeDescriptor as PropertyDescriptor))[target]
```

Editing `interceptor.ts` makes the linter strip that `as PropertyDescriptor` assertion, and typecheck then reports TS2532 on the indexed access, because the nested `intercept` call does not carry the narrowing through. The ternary plus optional access was the smallest thing that satisfied both.

I would rather this diff stay on the bug. Two ways to do that:

1. Restore the original expression with a one line lint disable for the assertion.
2. Keep the rewrite and move it to its own PR, leaving this one as just the `null` handling.

Either is fine by me. Tell me which and I will push it.

src/document/interceptor.ts
```diff
@@ -102,18 +103,20 @@ export function prepareValueInterceptor(
     'value',
     function interceptorImpl(
       this: HTMLInputElement | HTMLTextAreaElement,
-      v: UIValueString | string,
+      v: UIValueString | string | null,
     ) {
-      const isUI = isUIValue(v)
+      // `value` is a [LegacyNullToEmptyString] DOMString.
+      const value = v === null ? '' : v
+      const isUI = isUIValue(value)
```
snowystinger, line 110: `value` can never be null, so why did we add it to the signature for `isUIValue`?

snowystinger (commented):
don't worry about the logic around `realFunc`, just took a little bit of staring to make sure it was the same as before and this is easier to follow

PR #1332, opened 2026-09-12
fix: track value of inputs and textareas in shadow DOM

### LINKED ISSUE #1314: `user.type()` doesn't handle clearing of textareas in shadow DOM
### Reproduction example

https://codesandbox.io/p/devbox/still-leaf-dfv3dd?file=%2Fmwe.ts&workspaceId=ws_WjQcfkhNBektSLNjvEXy89

### Prerequisites

Do `user.type()` on e.g. a textarea _in shadow DOM / web component_, clear it with e.g. `.value = ""` or `form.reset()`, and `user.type()` some more.

### Expected behavior

The testing library should handle the clearing correctly. It works as expected in light DOM.

### Actual behavior

When you `user.type()` the second time, the content from the first `user.type()` get put into the textarea.

### User-event version

14.6.1

### Environment

`@testing-library/user-event`: `14.6.1`, Node `v24.15.0`, Linux

### Additional context

This MWE should make it explicit. It seems to be a problem with how the caching is handled? Note that the unexpected behavior is the most clear on the second to last line.

```ts
// @vitest-environment jsdom
/* package.json:
{
  "name": "user-event-shadow-dom-repro",
  "private": true,
  "type": "module",
  "scripts": {
    "test": "vitest run"
  },
  "devDependencies": {
    "@testing-library/user-event": "^14.6.1",
    "jsdom": "^30.0.1",
    "vitest": "^4.1.10"
  }
}
*/
import userEvent from "@testing-library/user-event";
import { expect, test } from "vitest";

function getCachedUIValue(ta: HTMLTextAreaElement) {
  for (const sym of Object.getOwnPropertySymbols(ta)) {
    if (sym.description === "Displayed value in UI") {
      return (ta as unknown as Record<symbol, unknown>)[sym] ?? "";
    }
  }
}

function lightTextarea(): HTMLTextAreaElement {
  const ta = document.createElement("textarea");
  document.body.append(ta);
  return ta;
}
test("light DOM works as expected", async () => {
  const ta = lightTextarea();

  // Type something into the <textarea>
  await user.type(ta, "hello");
  expect(ta.value).toBe("hello");
  expect(getCachedUIValue(ta)).toBe("hello");

  // Clear the textarea content. FWIW, form.reset() has the same effect.
  ta.value = "";
  expect(ta.value).toBe("");
  expect(getCachedUIValue(ta)).toBe("");

  // Type some more text. In principle, the <textarea> should now only contain "world"
  await user.type(ta, "world");
  expect(ta.value).toBe("world");
  expect(getCachedUIValue(ta)).toBe("world");
});

// MWE
function shadowTextarea(): { host: HTMLElement; ta: HTMLTextAreaElement } {
  const host = document.createElement("div");
  const shadow = host.attachShadow({ mode: "open" });
  const ta = document.createElement("textarea");
  shadow.append(ta);
  document.body.append(host);
  return { host, ta };
}
test.fails("but shadow DOM doesn't", async () => {
  const { ta } = shadowTextarea();

  // Type something into the <textarea>
  await user.type(ta, "hello");
  expect(ta.value).toBe("hello");
  expect(getCachedUIValue(ta)).toBe("hello");

  // Clear the textarea content. FWIW, form.reset() has the same effect.
  ta.value = "";
  expect(ta.value).toBe("");
  expect(getCachedUIValue(ta)).toBe("");

  // Type some more text. In principle, the <textarea> should now only contain "world"
  await user.type(ta, "world");
  expect(ta.value).toBe("world");
  expect(getCachedUIValue(ta)).toBe("world");
});

test("instead, shadow DOM has an outdated cache", async () => {
  const { ta } = shadowTextarea();

  // Type something into the <textarea>
  await user.type(ta, "hello");
  expect(ta.value).toBe("hello");
  expect(getCachedUIValue(ta)).toBe("hello");

  // Clear the textarea content. FWIW, form.reset() has the same effect.
  ta.value = "";
  expect(ta.value).toBe("");
  expect(getCachedUIValue(ta)).toBe("hello");

  // Type some more text. In principle, the <textarea> should now only contain "world"
  await user.type(ta, "world");

  // !!! This is the main MWE / unexpected behavior
  expect(ta.value).toBe("helloworld");
  // !!!

  expect(getCachedUIValue(ta)).toBe("helloworld");
});
```

TanbirRamim:
### What
Fixes #1314. For an `<input>` or `<textarea>` inside a shadow root, setting `.value` after `user.type()` was not picked up, so the next `user.type()` appended to the old text (`helloworld` instead of `world`). The `change` event on blur was also never dispatched for these elements.

### Why
The document-level `focus` and `blur` listeners in `prepareDocument` used `event.target`, which is retargeted to the shadow host, so the value interceptor was never installed on the actual element.

### How
Both listeners now use `event.composedPath()[0]` to get the original target. I added tests for value tracking and the blur `change` event in `tests/document/index.ts`, plus the reported `user.type()` case in `tests/utility/type.ts`; all three fail before the change and pass after, and lint and typecheck are clean. Closed shadow roots still resolve to the host, same as before.

### Checklist

- [x] Documentation N/A
- [x] Tests
- [x] Ready to be merged


src/document/prepareDocument.ts
```diff
@@ -23,7 +23,8 @@ export function prepareDocument(document: Document) {
   document.addEventListener(
     'focus',
     e => {
-      const el = e.target as Element
+      // Focus on elements in a shadow tree is retargeted to the shadow host.
+      const el = e.composedPath()[0] as Element
```
snowystinger, line 27: The first three steps of the composedPath algorithm suggest this could be an empty array
https://dom.spec.whatwg.org/#dom-event-composedpath

one place this can happen is if a synthetic focus or blur is called. Some libraries do this to trigger known side effects, so I think we need to check the length and fall back to target if it's empty

src/document/prepareDocument.ts
snowystinger, line 41: I think these lines need to use `getActiveElement` instead?

